PR #2, opened 2014-10-31
Update Elasticsearch Output Plugin to retry bulk

talevy:
Some actions may fail within the ES client bulk call.
Now, some messages (specifically errors 429 and 503s) will be
retried up to 3 times. If there are still actions that are
unsuccessfully indexed, Stud Buffer will continue its current
behavior (to retry indefinitely).

Stud Buffer will replay all events it first attempted to flush
to Elasticsearch. This means duplicate events may find themselves
in Elasticsearch.

(migrated from https://github.com/elasticsearch/logstash/pull/1997 with a few modifications to make 
tests run once plugin is installed)

### DIFF
--- a/lib/logstash/outputs/elasticsearch.rb
+++ b/lib/logstash/outputs/elasticsearch.rb
@@ -3,8 +3,10 @@ require "logstash/namespace"
 require "logstash/environment"
 require "logstash/outputs/base"
 require "logstash/json"
+require "concurrent_ruby"
 require "stud/buffer"
 require "socket" # for Socket.gethostname
+require "thread" # for safe queueing
 require "uri" # for escaping user input
 require 'logstash-output-elasticsearch_jars.rb'
 
@@ -204,17 +206,28 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
   # Asks host for the list of all cluster nodes and adds them to the hosts list
   config :sniffing, :validate => :boolean, :default => false
 
-  # helper function to replace placeholders
-  # in index names to wildcards
-  # example:
-  #    "logs-%{YYYY}" -> "logs-*"
-  def wildcard_substitute(name)
-    name.gsub(/%\{[^}]+\}/, "*")
-  end
+  # Set max retry for each event
+  config :max_retries, :validate => :number, :default => 3
+
+  # Set retry policy for events that failed to send
+  config :retry_max_items, :validate => :number, :default => 100
+
+  # Set max interval between bulk retries
+  config :retry_max_interval, :validate => :number, :default => 5
 
   public
   def register
+    @submit_mutex = Mutex.new
+    # retry-specific variables
+    @retry_flush_mutex = Mutex.new
+    @retry_teardown_requested = Concurrent::AtomicBoolean.new(false)
+    # needs flushing when interval
+    @retry_queue_needs_flushing = ConditionVariable.new
+    @retry_queue_not_full = ConditionVariable.new
+    @retry_queue = Queue.new
+
     client_settings = {}
+
 
     if @protocol.nil?
       @protocol = LogStash::Environment.jruby? ? "node" : "http"
@@ -288,33 +301,27 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
 
     @client = Array.new
 
-    if protocol == "node" or @host.nil? # if @protocol is "node" or @host is not set
-      options = {
-          :host => @host,
-          :port => @port,
-      }.merge(common_options)
-      @client << client_class.new(options)
+    if protocol == "node" || @host.nil? # if @protocol is "node" or @host is not set
+      options = { :host => @host, :port => @port }.merge(common_options)
+      @client = [client_class.new(options)]
     else # if @protocol in ["transport","http"]
-      @host.each do |host|
-          (_host,_port) = host.split ":"
-          options = {
-            :host => _host,
-            :port => _port || @port,
-          }.merge(common_options)
-          @logger.info "Create client to elasticsearch server on #{_host}:#{_port}"
-          @client << client_class.new(options)
-      end # @host.each
+      @client = @host.map do |host|
+        (_host,_port) = host.split ":"
+        options = { :host => _host, :port => _port || @port }.merge(common_options)
+        @logger.info "Create client to elasticsearch server on #{_host}:#{_port}"
+        client_class.new(options)
+      end # @host.map
     end
 
     if @manage_template
       for client in @client
-          begin
-            @logger.info("Automatic template management enabled", :manage_template => @manage_template.to_s)
-            client.template_install(@template_name, get_template, @template_overwrite)
-            break
-          rescue => e
-            @logger.error("Failed to install template: #{e.message}")
-          end
+        begin
+          @logger.info("Automatic template management enabled", :manage_template => @manage_template.to_s)
+          client.template_install(@template_name, get_template, @template_overwrite)
+          break
+        rescue => e
+          @logger.error("Failed to install template: #{e.message}")
+        end
       end # for @client loop
     end # if @manage_templates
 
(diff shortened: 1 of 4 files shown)


lib/logstash/outputs/elasticsearch.rb
```diff
@@ -472,4 +504,42 @@ def teardown
...
+      @retry_flush_mutex.unlock
+      @retry_queue_not_full.signal if @retry_queue.size < @retry_max_items
+    end
+  end
+
+  
+  def retry_push(action)
+    @retry_queue << action
+    @retry_queue_needs_flushing.signal if @retry_queue.size >= @retry_max_items
+  end
```
colinsurprenant: I suggest `retry_push` should also support a list of actions parameter to avoid calling `.signal` on all actions once `@retry_queue.size >= @retry_max_items`. 

Calling `retry_flush` could now be
```ruby
      unless failed_actions.empty?
        @logger.debug "#{failed_actions.size}/#{actions.size} events were unsuccessful in sending"
        retry_push(failed_actions)
      end
```

and `retry_flush` could handle single or multiple actions with something like:

```ruby
  def retry_push(actions)
    Array(actions).each{|action| @retry_queue << action}
    @retry_queue_needs_flushing.signal if @retry_queue.size >= @retry_max_items
  end
```

lib/logstash/outputs/elasticsearch.rb
```diff
@@ -425,43 +442,58 @@ def generate_jks cert_path
   def receive(event)
     return unless output?(event)
 
+    # block until we have not maxed out our 
+    # retry queue. This is applying back-pressure
+    # to slow down the receive-rate
+    @retry_queue_not_full.wait if @retry_queue.size > @retry_max_items
```
colinsurprenant: condition variables wait should always check for met condition upon returning from wait

```ruby
@retry_queue_not_full.wait while @retry_queue.size > @retry_max_items
```

lib/logstash/outputs/elasticsearch.rb
talevy: I now agree

lib/logstash/outputs/elasticsearch.rb
talevy: good catch

lib/logstash/outputs/elasticsearch.rb
```diff
@@ -472,4 +504,42 @@ def teardown
     require "logstash/outputs/elasticsearch/#{name}"
   end
 
+  def retry_flush(teardown=false)
```
colinsurprenant: cosmetic:
```ruby
def retry_flush(teardown = false)
```

lib/logstash/outputs/elasticsearch.rb
```diff
@@ -425,43 +442,58 @@ def generate_jks cert_path
...
-    # In the worst case, a failing flush (exception) will incur a retry from Stud::Buffer.
   end # def flush
 
   def teardown
     if @cacert # remove temporary jks store created from the cacert
       File.delete(@truststore)
     end
+
+    Thread.kill(@retry_timer_thread)
+    retry_flush(teardown=true)
```
colinsurprenant: cosmetic:
```ruby
retry_flush(teardown = true)
```

talevy:
rebased master, tended to the code changes, passed tests

suyograo:
@colinsurprenant are you good with these changes? I'd like us to merge this in today

colinsurprenant:
We're getting there :)
- The class is getting big and I think it would help if all `public`/`protected`/`private` methods would be grouped together in that order. we could write `protected` and `private` just once before each group.  I would also verify each method visibility to make sure the `protected` & `public` methods have the correct access or if we could make them `private`.
- The `register` method is too big.
- The `@@plugins` initialization is in the middle of the class between methods, I'd move that before the first method definition as this is class initialization code.
- I don't understand what the `@@plugins` code in the `register` method does and it looks fishy to me.
- Still need to resolve the potential concurrent client call between the flush/submit and retry_flush/submit threads.
- I am a bit puzzled over the lack of explicit JSon serializing for the bulk requests!? where is it done? ideally we'd want to control that by using `event.to_json` or `LogStash::Json.dump` on the `event.to_hash`. But by using` event.to_json`, this could potentially benefit eventual event json serialization caching

lib/logstash/outputs/elasticsearch.rb
```diff
@@ -477,4 +524,42 @@ def teardown
...
+    Array(actions).each{|action| @retry_queue << action}
+    @retry_queue_needs_flushing.signal if @retry_queue.size >= @retry_max_items
+  end
+
+  # helper function to replace placeholders
+  # in index names to wildcards
+  # example:
+  #    "logs-%{YYYY}" -> "logs-*"
+  private
+  def wildcard_substitute(name)
```
talevy: woops. didn't catch this during the rebase merge

talevy:
I am unclear on what `@@plugins` is all about as well, maybe @jsvd can chime in on that one.

I agree, there are lots of things that are due for some changing in the structure of this plugin, but I do not believe this PR is the appropriate place to address these other features.

lib/logstash/outputs/elasticsearch.rb
```diff
@@ -425,43 +442,58 @@ def generate_jks cert_path
...
-    #
-    # In the worst case, a failing flush (exception) will incur a retry from Stud::Buffer.
   end # def flush
 
   def teardown
     if @cacert # remove temporary jks store created from the cacert
       File.delete(@truststore)
     end
+
+    Thread.kill(@retry_timer_thread)
```
talevy: addressed

suyograo:
+1 on refactoring this class later

talevy:
I have added a mutex to guarantee thread-safety within `submit()`

I do not thing this plugin is the appropriate place to validate the safety of depended libraries.

lib/logstash/outputs/elasticsearch.rb
```diff
@@ -330,8 +337,141 @@ def register
...
+
+    index = event.sprintf(@index)
+
+    document_id = @document_id ? event.sprintf(@document_id) : nil
+    buffer_receive([event.sprintf(@action), { :_id => document_id, :_index => index, :_type => type }, event])
+  end # def receive
+
+  public
+  def submit(actions)
+    es_actions = actions.map { |a, doc, event| [a, doc, event.to_hash] }
```
colinsurprenant: suggested comment:

```ruby
# synchronize the @current_client.bulk call to avoid concurrency/thread safety issues with the
# client libraries which might not be thread safe. the submit method can be called from both the
# Stud::Buffer flush thread and from our own retry thread.
```

lib/logstash/outputs/elasticsearch.rb
```diff
@@ -426,55 +536,53 @@ def generate_jks cert_path
...
-      @logger.error "Got error to send bulk of actions to elasticsearch server at #{@host[@client_idx]} : #{e.message}"
-      raise e
+      @retry_queue_needs_flushing.wait(@retry_flush_mutex) unless @retry_teardown_requested.true?
     ensure
-      unless @protocol == "node"
-          @logger.debug? and @logger.debug "Shifting current elasticsearch client"
-          shift_client
-      end
+      @retry_flush_mutex.unlock
     end
```
colinsurprenant: maybe rewrite it like this?
```ruby
if @retry_teardown_requested.false?
    @retry_flush_mutex.lock
    begin
      @retry_queue_needs_flushing.wait(@retry_flush_mutex)
    ensure
      @retry_flush_mutex.unlock
    end
end
```

now, since this is not a tight loop, we could use `synchronize` instead of lock/unlock like this?

```ruby
@retry_flush_mutex.synchronize { @retry_queue_needs_flushing.wait(@retry_flush_mutex) } unless @retry_teardown_requested.true?
```

lib/logstash/outputs/elasticsearch.rb
```diff
@@ -330,8 +337,141 @@ def register
...
+  def submit(actions)
+    es_actions = actions.map { |a, doc, event| [a, doc, event.to_hash] }
+    @submit_mutex.lock
+    begin
+      bulk_response = @current_client.bulk(es_actions)
+    ensure
+      @submit_mutex.unlock
+    end
+    if bulk_response["errors"]
+        failed_actions = actions.select.with_index {|_,i| [429, 503].include?(bulk_response['statuses'][i]) }
```
colinsurprenant: cosmetic: indentation

lib/logstash/outputs/elasticsearch.rb
talevy: good call

lib/logstash/outputs/elasticsearch.rb
talevy: sure, I like the outer if. do not really have a preference with the synchronize, but less verbose. so it shall be done.

lib/logstash/outputs/elasticsearch.rb
talevy: noted.

lib/logstash/outputs/elasticsearch.rb
```diff
@@ -426,55 +536,53 @@ def generate_jks cert_path
...
+      buffer = @retry_queue.size.times.map do
+        next_action, next_doc, next_event = @retry_queue.pop
+        next_event['@metadata']['retry_count'] += 1
+
+        if next_event['@metadata']['retry_count'] > @max_retries
+          @logger.error "too many attempts at sending event. dropping: #{next_event}"
+          return nil
+        end
+
+        [next_action, next_doc, next_event]
```
colinsurprenant: `return nil` will actually return from the method and not produce `nil` for the map. this should be either
```ruby
if next_event['@metadata']['retry_count'] > @max_retries
  @logger.error "too many attempts at sending event. dropping: #{next_event}"
  nil
else
  [next_action, next_doc, next_event]
end
```
or
```ruby
if next_event['@metadata']['retry_count'] > @max_retries
  @logger.error "too many attempts at sending event. dropping: #{next_event}"
  next nil
end

[next_action, next_doc, next_event]
```

lib/logstash/outputs/elasticsearch.rb
colinsurprenant: obviously the synchronize construct is way more elegant and concise, but it involves creating and calling a closure which is less efficient so using lock/unlock will be better from a performance standpoint but only if on a perfromance critical path which I do not think it is here really...

lib/logstash/outputs/elasticsearch.rb
```diff
@@ -426,55 +539,51 @@ def generate_jks cert_path
...
+    unless @retry_queue.empty?
+      buffer = @retry_queue.size.times.map do
+        next_action, next_doc, next_event = @retry_queue.pop
+        next_event['@metadata']['retry_count'] += 1
 
-    document_id = @document_id ? event.sprintf(@document_id) : nil
-    buffer_receive([event.sprintf(@action), { :_id => document_id, :_index => index, :_type => type }, event.to_hash])
-  end # def receive
+        if next_event['@metadata']['retry_count'] > @max_retries
+          @logger.error "too many attempts at sending event. dropping: #{next_event}"
```
colinsurprenant: until we have the DLQ, wondering if we should log more upon dropping the event? 
```ruby
#{next_event.to_hash_with_metadata.inspect}
```
?? 

Any thoughts @jordansissel ?

lib/logstash/outputs/elasticsearch.rb
```diff
@@ -330,8 +337,144 @@ def register
...
+        sleep(@retry_max_interval)
+        @retry_queue_needs_flushing.signal
+      end
+    end
+
+    @retry_thread = Thread.new do
+      while @retry_teardown_requested.false?
+        retry_flush
+      end
+    end
```
colinsurprenant: I think we should move the signalling logic from the `retry_flush` method to here:

```ruby
    @retry_thread = Thread.new do
      while @retry_teardown_requested.false?
        @retry_flush_mutex.synchronize{@retry_queue_needs_flushing.wait(@retry_flush_mutex)}
        retry_flush
      end
    end
```

talevy: very true. didn't catch this during out refactoring

talevy:
@colinsurprenant updated again!

colinsurprenant:
Looks pretty good to me. We will need to followup with a code reorg/cleanup PR. 
I did not review the tests and I did not stress test or do integration test on this. I think we should. For http we could simply create a dummy http server to mock an ES instance.

talevy:
thanks @colin for the thorough review

jsvd:
WOOHOOO :fireworks:  so glad to see this merged! :tada:

ralphm:
I just ran into the issue of dropped events because I mistakenly tried to index a document with a type starting with an underscore. This yields a `400` error, but the drop happens silently and I only found out what happened by tcpdump. Although I see elasticsearch/logstash#1631 is closed because of this ticket, I couldn't find the follow-up PR that @colinsurprenant hinted at?

talevy:
@ralphm I will add more explicit logging to inform of this better. Currently we only retry `429` and `503` errors. Others will be logged and dropped until we have a persistent way of keeping track of these without halting the pipeline. Let me know if this would help, or maybe another behavior you would expect.

talevy:
@ralphm the logging changes have been merged into master. hope that makes things a little better until we add dead-letter queuing support!

PR #25, opened 2014-11-28
dynamic templates

Bradox:
Added the possiblity to use dinamyc templates. For example:
    
        elasticsearch {
                            cluster => "logstash"
                            index => "%{type}-logstash-%{+YYYY.MM.dd}"
                            template => "/etc/logstash/conf.d/%{type}-template.json"
                            template_name => "%{type}"
            }
    
It will also allow to parametrize the template. For example:
    
    {
        "template": "%{type}-",
        "settings": {
          "number_of_shards"            : 5,
          "index.cache.field.type"      : "soft",
          "index.refresh_interval"      : "2s",
          "index.store.compress.stored" : true,
          "index.query.default_field"   : "@message"
        },
        .......
    }

### DIFF
--- a/lib/logstash/outputs/elasticsearch.rb
+++ b/lib/logstash/outputs/elasticsearch.rb
@@ -360,7 +360,7 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
   def get_template(event)
     if @template.nil?
       @template = ::File.expand_path('elasticsearch/elasticsearch-template.json', ::File.dirname(__FILE__))
-      if !File.exists?(event.sprintf(@template))
+      if !File.exists?(@template)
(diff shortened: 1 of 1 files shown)


lib/logstash/outputs/elasticsearch.rb
```diff
@@ -425,6 +414,18 @@ def generate_jks cert_path
   def receive(event)
     return unless output?(event)
 
+    if @manage_template
```
andrewvc: Since there's been no activity here I'm going to close this issue. Thanks for he patch @Bradox , if you'd like to try to revive this idea I'd recommend opening an issue as a design discussion to suss out the pitfalls here!

PR #29, opened 2014-12-04
Add support for create actions in the node client

### LINKED ISSUE #28: Node Protocol Client does not support create requests
Similar to #16, the node client does not support `"create"` actions.

Create is the same as index operations except that it should require the ID (otherwise it's going to be implicitly create operation anyway!).

pickypg:
- Adds validation to require `_id` for create actions (including for the http protocol)

**Help**: I don't know how to properly test for the error condition. The `"create action fails without _id"` test fails because I'm not testing it right, but I don't know how to properly catch the error.

Note: #2 (retry on bulk failure) may need to special case failure to create (ignore it).

Closes #28

### DIFF
--- a/lib/logstash/outputs/elasticsearch/protocol.rb
+++ b/lib/logstash/outputs/elasticsearch/protocol.rb
@@ -45,11 +45,22 @@ module LogStash::Outputs::Elasticsearch
       #
       # If any action is missing a required arg, then a
       # Logstash::ConfigurationError is raised.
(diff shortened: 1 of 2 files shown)


suyograo:
@pickypg @talevy wanted to rebase this after #2 got merged in

suyograo:
Closed in lieu of https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/39

PR #40, opened 2015-01-21
add explicit description of protocol=node,host=nil scenario

talevy:
By default Elasticsearch has multicast discovery enabled.

This scenario will fail when the `discovery.zen.ping.multicast.enabled` option is set to `false`
```
elasticsearch.yml
-----------------
...
discovery.zen.ping.multicast.enabled: false
...
```

So be sure to have the proper settings when configuring this plugin to 
use the `node` protocol and not explicitly setting a `host`


suyograo:
@talevy can we also add this info here https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/master/lib/logstash/outputs/elasticsearch.rb#L88

this gets rendered to asciidoc

suyograo:
LGTM

PR #45, opened 2015-01-28
move plugin require code to bottom of class

jsvd:
since some plugins modify existing methods in host class,
this loop must be at the bottom, where all methods have been declared


purbon:
LGTM

talevy:
LGTM

PR #47, opened 2015-01-29
FTW dependency management

purbon:
This enables and simplify the ftw dependency gathering from ftw, polling the last minor update every time. This simplify later on our test workflow.


jsvd:
LGTM

PR #49, opened 2015-02-06
raise exception when bulk submission fails so stud:buffer retries

talevy:
in response to https://github.com/elasticsearch/logstash/issues/2516

had to revamp an existing test that was affected by the line change.


untergeek:
LGTM.  Glad it was something simple.

Issue #61, opened 2015-02-23
Add Better Logging for failed actions in bulk request


talevy:
update for better logging here: https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/master/lib/logstash/outputs/elasticsearch.rb#L415-L420

addresses comment provided here: https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/2#issuecomment-75611329

talevy:
WIP branch: https://github.com/talevy/logstash-output-elasticsearch/tree/fix/61-better-failure-logging

there is currently an issue when building and testing plugins against Logstash Master

talevy:
PR here: #62

PR #62, opened 2015-02-24
add better logging to failed actions

talevy:
addresses issue: https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/61

and comment: https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/2#issuecomment-75611329

No method signatures, or return values, are changed. That being said, there is some small non-trivial code cleanup to allow for log messages to include information about the exact error codes that are returned by Elasticsearch.

### DIFF
--- a/lib/logstash/outputs/elasticsearch.rb
+++ b/lib/logstash/outputs/elasticsearch.rb
@@ -35,6 +35,8 @@ require 'logstash-output-elasticsearch_jars.rb'
 # Elasticsearch to Logstash)
 class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
   include Stud::Buffer
+  RETRYABLE_CODES = [429, 503]
+  SUCCESS_CODES = [200, 201]
 
   config_name "elasticsearch"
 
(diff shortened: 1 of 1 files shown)


lib/logstash/outputs/elasticsearch.rb
```diff
@@ -413,11 +415,17 @@ def submit(actions)
       @submit_mutex.unlock
     end
     if bulk_response["errors"]
-      failed_actions = actions.select.with_index {|_,i| [429, 503].include?(bulk_response['statuses'][i]) }
-      unless failed_actions.empty?
-        @logger.debug "#{failed_actions.size}/#{actions.size} events were unsuccessful in sending"
-        retry_push(failed_actions)
+      actions_with_responses = actions.zip(bulk_response['statuses'])
+      actions_to_retry = []
+      actions_with_responses.each do |action, resp_code|
```
jordansissel: If I'm reading this correctly, this is to handle "partial retry" differently than "full retry" ? Where RETRYABLE_CODES indicates a partial retry and not being in SUCCESS_CODES means a full retry?

jordansissel: My confusion mostly is seeing the PR subject 'add better logging' but I see behavioral changes in this code, or at least, my quick glance seemed to indicate it :)

talevy: :disappointed: there was no access the the exact response code previously. so the error wouldn't be able to let people know what exactly was the reason something went wrong.

Full retries are still managed at a higher level by StudBuffer. mostly by client issues relating to connection errors. I just named them to be more explicit that 429 and 502 are a part of the class of errors we decided are retryable.

jordansissel: ok cool, makes sense. Code LGTM. I haven't tested yet.

lib/logstash/outputs/elasticsearch.rb
```diff
@@ -413,11 +415,17 @@ def submit(actions)
...
+        if RETRYABLE_CODES.include?(resp_code)
+          @logger.warn "retrying failed action with response code: #{resp_code}"
+          actions_to_retry << action
+        elsif not SUCCESS_CODES.include?(resp_code)
+          @logger.warn "failed action with response of #{resp_code}, retrying action: #{action}"
```
talevy: @jordansissel you are right... this should not say what it says! if errors are not a part of `RETRYABLE_CODES` or `SUCCESS_CODES`, they are dropped... not retried. mistake

jordansissel: ok I'll do some tests to verify things <3

PR #64, opened 2015-02-27
Support routing option for documents

masteinhauser:
The bulk API supports a specified `_routing` value for document
placement within a shard in an index. This should be exposed for custom
routing for aliases, filtering, etc.

http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/docs-bulk.html#bulk-routing

This also respects the move away from the routing parameter within the
event as defined in:
https://github.com/elasticsearch/elasticsearch/issues/6730

### DIFF
--- a/lib/logstash/outputs/elasticsearch.rb
+++ b/lib/logstash/outputs/elasticsearch.rb
@@ -49,6 +49,10 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
   # similar events to the same 'type'. String expansion `%{foo}` works here.
   config :index_type, :validate => :string
 
+  # A routing override to be applied to all processed events.
+  # This can be dynamic using the `%{foo}` syntax.
+  config :routing, :validate => :string
+
   # Starting in Logstash 1.3 (unless you set option `manage_template` to false)
   # a default mapping template for Elasticsearch will be applied, if you do not
   # already have one set to match the index pattern defined (default of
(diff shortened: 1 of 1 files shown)


lib/logstash/outputs/elasticsearch.rb
```diff
@@ -250,7 +254,7 @@ def register
       client_settings["network.host"] = @bind_host if @bind_host
       client_settings["transport.tcp.port"] = @bind_port if @bind_port
       client_settings["client.transport.sniff"] = @sniffing
- 
+
```
jordansissel: your editor appears to have made a bunch of whitespace changes.

masteinhauser:
Also, this was first requested back in [JIRA](https://logstash.jira.com/browse/LOGSTASH-1211) and is finally implemented. Though, not specifically for space optimization as originally mentioned.

lib/logstash/outputs/elasticsearch.rb
masteinhauser: Yeah, I noticed that too and was unhappy. Do I need to back them out? Should be easy to do.

jordansissel:
Can you fix the unrelated whitespace changes? It's hard to read the diff and beyond that it will make `git blame` point at you for things you shouldn't be blamed for :)

masteinhauser:
Fixed! A much easier diff to review.

masteinhauser:
Closing in support of #65 due to tests and more complete code.

suyograo:
@masteinhauser thanks for your contribution! we'll fold your PR into #65

PR #87, opened 2015-03-25
fix http protocol status response handling and only attempt to call retry_push if necessary

talevy:
- fix for how errors are parsed in the http protocol
- added more tests for error parsing from client responses in bulk api calls
- now retry_push is only called if there are events to be retried (it was being called regardless beforehand)

### DIFF
--- a/lib/logstash/outputs/elasticsearch/protocol.rb
+++ b/lib/logstash/outputs/elasticsearch/protocol.rb
@@ -78,6 +78,24 @@ module LogStash::Outputs::Elasticsearch
         Elasticsearch::Client.new client_options
       end
 
+      def self.normalize_bulk_response(bulk_response)
+        if bulk_response["errors"]
+          # The structure of the response from the REST Bulk API is follows:
+          # {"took"=>74, "errors"=>true, "items"=>[{"create"=>{"_index"=>"logstash-2014.11.17",
+          #                                                    "_type"=>"logs",
+          #                                                    "_id"=>"AUxTS2C55Jrgi-hC6rQF",
+          #                                                    "_version"=>1,
+          #                                                    "status"=>400,
+          #                                                    "error"=>"MapperParsingException[failed to parse]..."}}]}
+          # where each `item` is a hash of {OPTYPE => Hash[]}. calling first, will retrieve 
+          # this hash as a single array with two elements, where the value is the second element (i.first[1])
+          # then the status of that item is retrieved.
+          {"errors" => true, "statuses" => bulk_response["items"].map { |i| i.first[1]['status'] }}
+        else
+          {"errors" => false}
+        end
+      end
+
       def bulk(actions)
         bulk_response = @client.bulk(:body => actions.collect do |action, args, source|
           if source
@@ -86,11 +104,8 @@ module LogStash::Outputs::Elasticsearch
             next { action => args }
           end
         end.flatten)
-        if bulk_response["errors"]
-          return {"errors" => true, "statuses" => bulk_response["items"].map { |i| i["status"] }}
-        else
-          return {"errors" => false}
-        end
+
+        self.class.normalize_bulk_response(bulk_response)
       end # def bulk
 
       def template_exists?(name)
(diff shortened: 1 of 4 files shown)


lib/logstash/outputs/elasticsearch/protocol.rb
```diff
@@ -86,11 +94,8 @@ def bulk(actions)
             next { action => args }
           end
         end.flatten)
-        if bulk_response["errors"]
-          return {"errors" => true, "statuses" => bulk_response["items"].map { |i| i["status"] }}
-        else
-          return {"errors" => false}
-        end
+
+        self.class.simple_bulk_response(bulk_response)
```
suyograo: why use `self.class`? you can just do: `simple_bulk_response(bulk_response)`

lib/logstash/outputs/elasticsearch/protocol.rb
```diff
@@ -78,6 +78,14 @@ def build_client(options)
         Elasticsearch::Client.new client_options
       end
 
+      def self.simple_bulk_response(bulk_response)
+        if bulk_response["errors"]
+          {"errors" => true, "statuses" => bulk_response["items"].map { |i| i.first[1]['status'] }}
```
suyograo: comment please `i.first[1]['status']`

lib/logstash/outputs/elasticsearch/protocol.rb
```diff
@@ -187,6 +192,15 @@ def build_client(options)
         return nodebuilder.settings(@settings).node.client
       end # def build_client
 
+      def self.simple_bulk_response(bulk_response)
```
suyograo: same here ^^. no need for `self`

lib/logstash/outputs/elasticsearch/protocol.rb
```diff
@@ -195,13 +209,7 @@ def bulk(actions)
...
 
-        if response.has_failures()
-          return {"errors" => true,
-                  "statuses" => response.map { |i| (i.is_failed && i.get_failure.get_status.get_status) || 200 }}
-        else
-          return {"errors" => false}
-        end
-        # returns 200 for all successful actions, represents 201 & 200
+        self.class.simple_bulk_response(resp)
         # TODO(talevy): parse item response objects to retrieve correct 200 (OK) or 201(created) status codes
```
suyograo: move this TODO to the function

lib/logstash/outputs/elasticsearch/protocol.rb
suyograo: Also can we add some checks if the array `i.first[1]` is empty?

suyograo: never mind, this only enters if errors exist

spec/outputs/elasticsearch/protocol_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+require "logstash/outputs/elasticsearch/protocol"
+require "java"
+
+describe LogStash::Outputs::Elasticsearch::Protocols::NodeClient do
+  context "successful" do
+    it "should map correctly" do
+      index_response = org.elasticsearch.action.index.IndexResponse.new("my_index", "my_type", "my_id", 123, true)
+      delete_response = org.elasticsearch.action.delete.DeleteResponse.new("my_index", "my_type", "my_id", 123, true)
+      bulk_item_response_index = org.elasticsearch.action.bulk.BulkItemResponse.new(32, "index", index_response)
+      bulk_item_response_delete = org.elasticsearch.action.bulk.BulkItemResponse.new(32, "index", delete_response)
```
suyograo: `delete`?

lib/logstash/outputs/elasticsearch/protocol.rb
```diff
@@ -86,11 +104,8 @@ def bulk(actions)
             next { action => args }
           end
         end.flatten)
-        if bulk_response["errors"]
-          return {"errors" => true, "statuses" => bulk_response["items"].map { |i| i["status"] }}
-        else
-          return {"errors" => false}
-        end
+
+        simple_bulk_response(bulk_response)
```
suyograo: hmm can we call this `normalize_bulk_response` .. not sure `simple_bulk_response` is descriptive

suyograo:
LGTM, minor nitpick :)

lib/logstash/outputs/elasticsearch/protocol.rb
talevy: this gets complicated for testing the `HTTPClient` because it depends on `:client_settings` options being passed in, and that is only passed in during an integration with actual connections. I would rather prefer to leave this be a unit test

lib/logstash/outputs/elasticsearch/protocol.rb
talevy: same as above ^^^

lib/logstash/outputs/elasticsearch/protocol.rb
talevy: agreed. will do this.

PR #90, opened 2015-03-30
update tests to fix test suite to not be flaky

talevy:
Update tests to correctly unstub stubbing of ES clients
and other minor edits.

### DIFF
--- a/spec/outputs/elasticsearch_spec.rb
+++ b/spec/outputs/elasticsearch_spec.rb
@@ -486,8 +486,13 @@ describe "outputs/elasticsearch" do
         .any_instance.stub(:bulk).and_return(*resp)
       LogStash::Outputs::Elasticsearch::Protocols::NodeClient
         .any_instance.stub(:bulk).and_return(*resp)
-      LogStash::Outputs::Elasticsearch::Protocols::TransportClient
-        .any_instance.stub(:bulk).and_return(*resp)
+    end
+
+    def unmock_actions()
+      LogStash::Outputs::Elasticsearch::Protocols::HTTPClient
+        .any_instance.unstub(:bulk)
+      LogStash::Outputs::Elasticsearch::Protocols::NodeClient
+        .any_instance.unstub(:bulk)
     end
 
     ["node", "transport", "http"].each do |protocol|
@@ -526,6 +531,7 @@ describe "outputs/elasticsearch" do
           subject.receive(event2)
           subject.buffer_flush(:final => true)
           sleep(2)
+          unmock_actions
         end
 
         it "should raise exception and be retried by stud::buffer" do
@@ -557,6 +563,7 @@ describe "outputs/elasticsearch" do
           subject.receive(event2)
           subject.buffer_flush(:final => true)
           sleep(3)
+          unmock_actions
         end
         
         it "should retry actions with response status of 429" do
@@ -567,6 +574,7 @@ describe "outputs/elasticsearch" do
           subject.receive(event1)
           subject.buffer_flush(:final => true)
           sleep(3)
+          unmock_actions
         end
 
         it "should retry an event until max_retries reached" do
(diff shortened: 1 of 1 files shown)


purbon:
@talevy reviewing

purbon:
Most sure is not related with this, but wanted to know more:

If I run the test like ```bundle exec rspec -t elasticsearch``` having a local ES instance using the default cluster name (something to improve for sure in here as we recommend to change this name).

I get this error:

```
Failures:

  1) outputs/elasticsearch failures in bulk class expected behavior with protocol => http non-retryable errors like mapping errors (400) should be dropped and not be retried (unfortunetly)
     Failure/Error: insist { r["hits"]["total"] } == 0
     Insist::Failure:
       Expected 0, but got 2
     # ./spec/outputs/elasticsearch_spec.rb:610:in `(root)'
     # ./spec/outputs/elasticsearch_spec.rb:608:in `(root)'

Finished in 2 minutes 26.8 seconds
52 examples, 1 failure
```

can you educate me why do I get it?

- purbon

purbon:
One comment on using stub and any_instance, for now it works as we're still using a 2.x branch of rspec, however when we upgrade to 3.x.

According to https://relishapp.com/rspec/rspec-mocks/v/3-2/docs/old-syntax/any-instance: 

> any_instance is the old way to stub or mock any instance of a class but carries the baggage of a global > monkey patch on all classes. Note that we generally recommend against using this feature.

specifically, the ones that concern me the most about being not comfortable when seeing this are:

> Using this feature is often a design smell. It may be that your test is trying to do too much or that the 
> object under test is too complex. 
> It is the most complicated feature of rspec-mocks, and has historically received the most bug reports. > (None of the core team actively use it, which doesn't help.)

spec/outputs/elasticsearch_spec.rb
```diff
@@ -486,8 +486,13 @@ def mock_actions_with_response(*resp)
         .any_instance.stub(:bulk).and_return(*resp)
       LogStash::Outputs::Elasticsearch::Protocols::NodeClient
         .any_instance.stub(:bulk).and_return(*resp)
-      LogStash::Outputs::Elasticsearch::Protocols::TransportClient
-        .any_instance.stub(:bulk).and_return(*resp)
+    end
+
+    def unmock_actions()
```
purbon: This might very necessary in here, usually stub scope is:

```
Doubles, stubs, and message expectations are all cleaned out after each
example. This ensures that each example can be run in isolation, and in any
order.
```

but I actually always found tricky to discover the right scope for mock of this kind. My best guess in here is that their scope is the context, so that is why in the way you want to use them you need to unstub them, and there is actually no shame on doing that way.

However we might try to refactor this test, or also including the plugin code, to now need this kind of constructs anymore. I for example wonder why do we need to mock the TransportClient, NodeClient and the HTTPClient if we might be only using one of them, for example.

purbon: Until doing a more detailed knowledge finding, the best resource about scoping I could find is https://github.com/rspec/rspec-mocks/blob/master/features/basics/scope.feature

spec/outputs/elasticsearch_spec.rb
```diff
@@ -597,6 +612,12 @@ def mock_actions_with_response(*resp)
         end
 
         it "successful requests should not be appended to retry queue" do
+          # insist indices are empty
+          Stud::try(10.times) do
+            r = @es.search
+            insist { r["hits"]["total"] } == 0
```
purbon: Why not moving out of insist as we did in core? we can do most of this expectations by simply using the expect rspec method.

talevy: I agree with you, but, I would like to have it be consistent across the file for now. once we fix the flakiness. I will go ahead and make PR to `s/insist/expect`

spec/outputs/elasticsearch_spec.rb
talevy: I see. so then my issue is probably that I was stubbing out both the TransportClient and NodeClient. I believe something was conflicting. That is why the TransportClient stubbing is deleted

talevy:
Closing in favor of #97

PR #91, opened 2015-03-30
Disable norms for _all field in the template

rmuir:
For logs, length normalization for this field probably does more harm than good.
It could even cause confusion for scoring, because its a combination of so
many structured fields, different from traditional search use cases.

We can save a good deal of memory by disabling norms, like the other fields.

See https://github.com/elastic/elasticsearch/pull/4573 which added this option to ES.

### DIFF
--- a/lib/logstash/outputs/elasticsearch/elasticsearch-template.json
+++ b/lib/logstash/outputs/elasticsearch/elasticsearch-template.json
@@ -5,7 +5,7 @@
   },
   "mappings" : {
     "_default_" : {
-       "_all" : {"enabled" : true},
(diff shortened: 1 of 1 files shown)


kimchy:
++

talevy:
If any existing logstash users are interested in taking advantage of this new default mapping, you can simply restart your logstash instance with `elasticsearch` output to set the `template_overwrite` option to `true`. This will have logstash publish this new template to an existing cluster that already has the old template.
After doing this, future indices will pick up this template and save you some extra space!

If you are starting up a new cluster, you may ignore the above instructions and simply reap the benefits of this new template without doing a thing!

talevy:
LGTM

PR #96, opened 2015-04-01
Remove `"path": "full"` from geoip mapping

### LINKED ISSUE #94: Deprecated "path": "full" mapping in elasticsearch index template
The default index template bundled with the plugin specifies `"path": "full"` for the `geoip` field in `_default_`. This has been [deprecated as of 1.0.0](http://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-object-type.html#_path_4), but more importantly, it outright breaks new index creation on latest (master) builds of elasticsearch:

```
org.elasticsearch.index.mapper.MapperParsingException: Mapping definition for [geoip] has unsupported parameters:  [path : full]
```

The docs indicate `copy_to` is the successor to `path`. In any case, the current behavior of logstash + latest builds of elasticsearch is a broken `elasticsearch` output, which is an annoyance for now, but will end up being big-blow-up-no-fun-for-anyone when logstash gets paired with a 2.x elasticsearch instance down the road. :boom: :sweat_smile:

tylerjl:
This causes recent (see: master) builds of elasticsearch to fail new
index create when drawing from the default mapping. "path": "full" is no
longer supported, but the default behavior in elasticsearch is such that
operations that rely on the path notation should still function
normally.

Fixes #94

### DIFF
--- a/lib/logstash/outputs/elasticsearch/elasticsearch-template.json
+++ b/lib/logstash/outputs/elasticsearch/elasticsearch-template.json
@@ -31,7 +31,6 @@
          "geoip"  : {
            "type" : "object",
              "dynamic": true,
-             "path": "full",
              "properties" : {
                "location" : { "type" : "geo_point" }
              }


talevy:
LGTM

PR #100, opened 2015-04-07
add warning for embedded => true

talevy:


lib/logstash/outputs/elasticsearch.rb
```diff
@@ -299,6 +299,7 @@ def register
 
     if @embedded
       raise(LogStash::ConfigurationError, "The 'embedded => true' setting is only valid for the elasticsearch output under JRuby. You are running #{RUBY_DESCRIPTION}") unless LogStash::Environment.jruby?
+      @logger.warn("The 'embedded => true' setting is enabled")
```
suyograo: Lets be more explicit: 
`'embedded => true' setting is enabled. This is not recommended for production use!`

talevy: updated.

suyograo:
LGTM

PR #105, opened 2015-04-14
deprecate usage of :index_type => :document_type

### LINKED ISSUE #102: Parameter "index_type" should resemble Elasticsearch lingo (more closely)
Consider me a user. **)*
I had a hard (and embarrassing) time figuring out the parameter for setting the `doc_type` (used in python) or `_type` (native) for the Elasticsearch output. Here I'll take the embarrassment one level further and blame it on you ;)

It may seem obvious once learned, but: coming from Elasticsearch, I *never ever* encountered the term `index_type`. Please consider changing or aliasing the parameter name, or at least mention it in the docs for `index_type` with one of the more typical synonyms. Maybe use `document_type`, since you already use `document_id`.

Furthermore I wonder, if the docs are correct?
```
index_type
*    Value type is string
*    There is no default value for this setting. 
```
it may have happened somewhere else in the processing chain, but without setting the parameter, my events ended up with the mapping-type `logs`.

*) well, maybe a little absent-minded, but that's how some of us are...

talevy:
Some users feel this jargin does not match nicely
with Elasticsearch. So a renaming is warranted.

Closed #102.

### DIFF
--- a/lib/logstash/outputs/elasticsearch.rb
+++ b/lib/logstash/outputs/elasticsearch.rb
@@ -49,7 +49,13 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
 
   # The index type to write events to. Generally you should try to write only
   # similar events to the same 'type'. String expansion `%{foo}` works here.
-  config :index_type, :validate => :string
+  # 
+  # Deprecated in favor of `document_type` field.
+  config :index_type, :validate => :string, :deprecated => true
+
+  # The document type to write events to. Generally you should try to write only
+  # similar events to the same 'type'. String expansion `%{foo}` works here.
+  config :document_type, :validate => :string
 
   # Starting in Logstash 1.3 (unless you set option `manage_template` to false)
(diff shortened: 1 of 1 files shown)


lib/logstash/outputs/elasticsearch.rb
```diff
@@ -49,7 +49,13 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
 
   # The index type to write events to. Generally you should try to write only
   # similar events to the same 'type'. String expansion `%{foo}` works here.
-  config :index_type, :validate => :string
+  # 
+  # Depricated in favor of `document_type` field.
```
suyograo: typo.. Deprecated

suyograo:
Small typo, otherwise LGTM

lib/logstash/outputs/elasticsearch.rb
talevy: done.

suyograo:
:ship:

PR #109, opened 2015-04-16
remove ROOT_DIR constant from jars loading file

jsvd:
since this code is present on all plugins that contain jars, each
one will define ROOT_DIR, causing warnings


suyograo:
LGTM

PR #113, opened 2015-04-17
Feature: Add Shield authentication for node transport

alcinnz:
When specifying a username and password with the Elasticsearch node protocol this patch assumes your connecting to Elasticsearch using Shield. 

An example configuration would be:

```
output {
  elasticsearch {
    user     => "es_admin"
    password => "administrator"
    host     => "localhost"
    protocol => "node"
  }
}
```

### DIFF
--- a/lib/logstash/outputs/elasticsearch.rb
+++ b/lib/logstash/outputs/elasticsearch.rb
@@ -261,6 +261,11 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
       raise(LogStash::ConfigurationError, "action => 'create_unless_exists' is not supported under the HTTP protocol");
     end
 
+    # Node protocol supports auth with the shield.user option
+    if @protocol == "node" && @user && @password
+      client_settings["shield.user"] = @user + ":" + @password.value
+    end
+
     if ["node", "transport"].include?(@protocol)
       # Node or TransportClient; requires JRuby
       raise(LogStash::PluginLoadingError, "This configuration requires JRuby. If you are not using JRuby, you must set 'protocol' to 'http'. For example: output { elasticsearch { protocol => \"http\" } }") unless LogStash::Environment.jruby?
@@ -542,6 +547,7 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
   private
   def setup_basic_auth
     return {} unless @user && @password
+    return {} if @protocol == "node" # Authentication already setup for "node" protocol
 
     if @protocol =~ /http/
       {
--- a/spec/outputs/elasticsearch_spec.rb
+++ b/spec/outputs/elasticsearch_spec.rb
@@ -886,24 +886,40 @@ describe "ship lots of events w/ default index_type and dynamic routing key usin
   end
 
   describe "Authentication option" do
-    ["node", "transport"].each do |protocol|
-      context "with protocol => #{protocol}" do
-        subject do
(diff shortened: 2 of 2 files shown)


suyograo:
@alcinnz thanks for your contribution; we will not be able to accept this PR for these reasons:

1. Shield is a commercial plugin for Elasticsearch and we don't want to mix references to our commercial product from our open source codebase. For this reason, we currently have a separate *overlay* plugin (`logstash-output-elasticsearch-shield`) which can be installed on top of Logstash to enable Shield integration.

2. The overlay plugin only supports `transport` and `http` protocol integration. We deliberately decided to not support `node` protocol in this plugin. Node client joins the secure Elasticsearch cluster and becomes a trusted member of it -- this grants Logstash **full** access to rest of the ES cluster. In this scenario, we need to make sure requests coming in from Logstash node client are gated by our License infra and are properly authorized. For the user, this would mean adding an extra Elasticsearch License plugin on top of `logstash-output-elasticsearch-shield` plugin. This would also mean that Logstash would have to bundle the same authorization configuration files (`roles.yml`, `users.yml` etc) used by Shield.

Check the Shield [documentation](http://www.elastic.co/guide/en/shield/current/logstash.html#ls-transport) on how you can leverage the transport client to authenticate and encrypt messages when using Logstash

suyograo:
Let me know if you need more info, will be closing this PR

theflimflam:
That's a pitty. We're quite interested in having Shield on the cluster, and we trust our logstash nodes enough to give them licence and yaml configurations as we're pushing these out with puppet anyway. We feel that the node transport gives us speed.

What would it take to get ```logstash-output-elasticsearch-shield``` to support node transport?

I can't see any public places where this conversation is happening.

jordansissel:
> We feel that the node transport gives us speed.

There should be no difference in performance. If there is, it's a bug. If you've got any data showing performance differences, I'd love to use that to help us resolve any performance bugs :)

ceeeekay:
Hi guys,

Rather than speed being the issue, we would prefer to have logstash indexers join the cluster using node protocol for a couple of main reasons:

1. Using node protocol means we only have to specify the master nodes in the logstash output, and logstash discovers and connects to the rest of the cluster. This gives us a simple output config and highly-redundant connections as logstash creates a full mesh with the rest of the cluster.

2. Using transport protocol means we will need some kind of gateway nodes acting as load balancers for logstash indexers. We would need at least two of these for redundancy purposes, and it seems wasteful to have these present just to handle indexing, when node transport does a fine job on its own. We do not want to use the master nodes for this, for obvious reasons, and would prefer to keep indexing away from the existing nodes which are providing load-balancing for ES queries.

jordansissel:
> Using node protocol means we only have to specify the master nodes in the logstash output, and logstash discovers and connects to the rest of the cluster

We've added that to the `transport` client already, and will do soon for http: https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/master/lib/logstash/outputs/elasticsearch.rb#L231

> Using transport protocol means we will need some kind of gateway nodes acting as load balancers for logstash indexers

Nah, the transport client will balance on its own if given multiple hosts to target (or if it finds them using the `sniff` setting)

Hope this helps clarify! :)

ceeeekay:
So ... if we specify the masters in our output config, will transport then send directly to data nodes, or will it send in a 2-hop fashion, via the masters?

If it's the first way, then we're done here. 
If not, we really don't want to be sending anything via the master nodes.

PR #115, opened 2015-04-21
revert index template wildcard overwriting.

talevy:
This was originally done to close https://github.com/elastic/logstash/issues/1901 and pr for this was here: https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/4, but does not seem to be
clear enough and necessary. It leads to lazy behavior, and makes it
difficult to take control of the exact template pattern.


untergeek:
lgtm

PR #123, opened 2015-04-30
removed some debug log lines

### LINKED ISSUE #120: Logging throws exception when no @host is set
https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/master/lib/logstash/outputs/elasticsearch.rb#L461-L464

the above lines expect a host to exist and wish to log this information. It does not necessarily exist.

related: https://github.com/elastic/logstash/issues/2894#issuecomment-97177598
the above comment describes a scenario when there is an exception thrown during the sending or receiving of a bulk request, Information about this exception is not being relayed to the user because the log message itself is raising a `Nil` exception first.

talevy:
some log messages were referencing @host field, even when
the field was nil and not applicable (when protocol was node).

I do not see the benefit of these messages, as the exception is thrown in case of emergency anyways. and that exception will be logged.

Closes #120.

### DIFF
--- a/lib/logstash/outputs/elasticsearch.rb
+++ b/lib/logstash/outputs/elasticsearch.rb
@@ -458,10 +458,9 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
   public
   def flush(actions, teardown = false)
     begin
-      @logger.debug? and @logger.debug "Sending bulk of actions to client[#{@client_idx}]: #{@host[@client_idx]}"
       submit(actions)
     rescue => e
-      @logger.error "Got error to send bulk of actions to elasticsearch server at #{@host[@client_idx]} : #{e.message}"
+      @logger.error "Got error to send bulk of actions: #{e.message}"
       raise e
     ensure
       unless @protocol == "node"


lib/logstash/outputs/elasticsearch.rb
```diff
@@ -458,10 +458,8 @@ def submit(actions)
   public
   def flush(actions, teardown = false)
     begin
-      @logger.debug? and @logger.debug "Sending bulk of actions to client[#{@client_idx}]: #{@host[@client_idx]}"
       submit(actions)
     rescue => e
-      @logger.error "Got error to send bulk of actions to elasticsearch server at #{@host[@client_idx]} : #{e.message}"
```
suyograo: Well sometimes, the raise is not very helpful and cryptic to non ruby developers..so maybe we can keep this? and check for `host.nil?`

Agree with the first debug log line

talevy: all this is doing is printing "e.message", which is what the exception does as well, no?

suyograo: Yeah but we are missing the context around it which is in the message "Got error to send bulk of actions to elasticsearch server" .. better to see that than a stack trace with 
```
Failed to flush outgoing items {:outgoing_count=>26, :exception=>#<NoMethodError: undefined method[]' for nil:NilClass>, :backtrace=>["/opt/logstash/vendor/bundle/jruby/1.9/gems/logstash-output-elasticsearch-0.2.4-java/lib/logstash/outputs/elasticsearch.rb:464:in flush'", "/opt/logstash/vendor/bundle/jruby/1.9/gems/stud-
```

suyograo: We can remove the host parameter which does not add any more value for debugging

talevy: deal. I was flip-flopping on that point before I pushed the commit. I'll do that

suyograo: :8ball:

talevy: updated.

suyograo:
LGTM

PR #125, opened 2015-05-01
Issues/124

ppf2:
Clarified protocol, host, cluster, etc.. settings.

### DIFF
--- a/lib/logstash/outputs/elasticsearch.rb
+++ b/lib/logstash/outputs/elasticsearch.rb
@@ -98,30 +98,35 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
   # By default, it looks for a cluster named 'elasticsearch'.
   config :cluster, :validate => :string
 
-  # For the `node` protocol, if you do not specify `host` parameter, it will attempt to use
-  # multicast discovery to connect to Elasticsearch.  If multicast is disabled in Elasticsearch (http://www.elastic.co/guide/en/elasticsearch/guide/current/_important_configuration_changes.html#_prefer_unicast_over_multicast), 
+  # For the `node` protocol, if you do not specify `host`, it will attempt to use
+  # multicast discovery to connect to Elasticsearch.  If http://www.elastic.co/guide/en/elasticsearch/guide/current/_important_configuration_changes.html#_prefer_unicast_over_multicast[multicast is disabled] in Elasticsearch, 
   # you must include the hostname or IP address of the host(s) to use for Elasticsearch unicast discovery.
   # Remember the `node` protocol uses the http://www.elastic.co/guide/en/elasticsearch/reference/current/modules-transport.html#modules-transport[transport] address (eg. 9300, not 9200).
   #     `"127.0.0.1"`
   #     `["127.0.0.1:9300","127.0.0.2:9300"]`
   # When setting hosts for `node` protocol, it is important to confirm that at least one non-client
-  # node is listed in the `host` list.  Also keep in mind that the `host` parameter when used with the `node` protocol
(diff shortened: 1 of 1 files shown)


lib/logstash/outputs/elasticsearch.rb
```diff
@@ -94,23 +94,43 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
...
+  # the `node` protocol is for *discovery purposes only* (not for load balancing).  When multiple hosts 
+  # are specified, it will contact the first host to see if it can use it to discover the cluster.  If not, 
+  # then it will contact the second host in the list and so forth. With the `node` protocol, 
+  # Logstash will join the Elasticsearch cluster as a node client (which has a copy of the cluster
+  # state) and this node client is the one that will automatically handle the load balancing of requests 
+  # across data nodes in the cluster.  
+  # If you are looking for a high availability setup, our recommendation is to use the `transport` protocol (below), 
+  # set up multiple http://www.elastic.co/guide/en/elasticsearch/reference/current/modules-node.html[client nodes] and list the client nodes in the `host` parameter.
+  # 
+  # For the `transport` protocol, it will load balance requests across the hosts specified in the `host` parameter.
```
suyograo: @ppf2 LS does not load balance, we will only try a second host if it fails to establish connection with the first one. It keeps shifting to the next host in this list, until successful

lib/logstash/outputs/elasticsearch.rb
```diff
@@ -94,23 +94,43 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
...
+  # Remember the `transport` protocol uses the http://www.elastic.co/guide/en/elasticsearch/reference/current/modules-transport.html#modules-transport[transport] address (eg. 9300, not 9200).
   #     `"127.0.0.1"`
   #     `["127.0.0.1:9300","127.0.0.2:9300"]`
+  # There is also a `sniffing` option (see below) that can be used with the transport protocol to instruct it to use the host to sniff for
+  # "alive" nodes in the cluster and automatically use it as the hosts list (but will skip the dedicated master nodes).  
+  # If you do not use the sniffing option, it is important to exclude http://www.elastic.co/guide/en/elasticsearch/reference/current/modules-node.html[dedicated master nodes] from the `host` list
+  # to prevent Logstash from sending bulk requests to the master nodes. So this parameter should only reference either data or client nodes.
+  #
+  # For the `http` protocol, it will load balance requests across the hosts specified in the `host` parameter.
```
suyograo: No load balancing here too

lib/logstash/outputs/elasticsearch.rb
ppf2: It seems like this was added as part of https://github.com/elastic/logstash/pull/1791. 

```
{:timestamp=>"2015-05-06T00:09:59.411000-0700", :message=>"Switched current elasticsearch client to #0 at localhost:9300", :level=>:debug, :file=>"logstash/outputs/elasticsearch.rb", :line=>"494", :method=>"shift_client"}
{:timestamp=>"2015-05-06T00:10:00.323000-0700", :message=>"Switched current elasticsearch client to #1 at localhost:9301", :level=>:debug, :file=>"logstash/outputs/elasticsearch.rb", :line=>"494", :method=>"shift_client"}
```
And it looks like it is constantly shifting the client when using the transport protocol even though both nodes are available.

lib/logstash/outputs/elasticsearch.rb
talevy: as per our email thread.

`shift_client` is ensured after every flush with `http` and `transport` protocols.

PR #132, opened 2015-05-17
Add missing action values in documentation

BHSPitMonkey:
These seemed to be missing (despite being documented just below).


andrewvc:
Thanks for the patch @BHSPitMonkey , but this is now fixed in the latest docs :)

Issue #111, opened 2015-04-16
introduce Docker for integration testing


talevy:
this will include:

1. start-up elasticsearch container before test suite
2. allow plugin tests to reach ES url served in docker container for testing against
3. teardown container when done

suyograo:
+1000

untergeek:
:+1: times a billion

untergeek:
what makes this extra cool is that the ES docker image built here can be used by, say, the couchdb_changes integration tests too.

PR #151, opened 2015-05-19
[WIP] parameterize and split ES spec tests into integration and unit tests split up by specific feature tests

talevy:
Currently, it is rather difficult to maintain the ever-growing
elasticsearch spec file. This Commit is an inital refactor of
the spec and splits out some of the specs into their own files,
and separates integration tests from unit tests. Uses Docker to
manage external Elasticsearch instance and fetches host/port 
info from docker for the specific ES instance. Node Protocol is 
not currently supported.

Closes https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/111.


untergeek:
YAY!  I'm excited to get a version of that docker rspec helper material into its own space for _all_ spec tests.

spec/integration/docker_rspec_helper.rb
```diff
@@ -0,0 +1,62 @@
...
+    puts "good luck!"
+  end
+
+  def cleanup
+    # if @container
+    #   @container.stop
+    #   @container.delete
+    # end
+    if @image
+      @image.remove(:force => true)
```
electrical: You have to remove the container not the image.

suyograo:
@purbon review?

spec/es_spec_helper.rb
```diff
@@ -0,0 +1,27 @@
+require "logstash/devutils/rspec/spec_helper"
+require "ftw"
+require "logstash/plugin"
+require "logstash/json"
+require "stud/try"
+
```
purbon: Why is this file named ```es_spec_helper.rb``` ? I would rather use a more clear naming in here. for example ```docker_helper.rb``` or simply ```spec_helper.rb```, what do you think?

spec/es_spec_helper.rb
```diff
@@ -0,0 +1,27 @@
...
+
+RSpec.configure do |config|
+  config.before(:all) do
+    require "integration/docker_rspec_helper"
+    @docker = RspecDocker.new
+    @docker.provision("elasticsearch", "spec/integration/docker/Dockerfile")
+    @es_server = {
```
purbon: Why do you need this variable to be a class variable?

spec/es_spec_helper.rb
```diff
@@ -0,0 +1,27 @@
...
+      :host => @docker.host[:host_ip],
+      :port => {
+        'http' => @docker.host[:ports]['9200/tcp'][0]['HostPort'],
+        'transport' => @docker.host[:ports]['9300/tcp'][0]['HostPort'],
+        'node' => @docker.host[:ports]['9300/tcp'][0]['HostPort']
+      }
+    }
+    sleep 13
```
purbon: This 13 here for a reason? can we ping the host to be awake? more on the side of what @untergeek is doing for zabbix.

spec/integration/docker_rspec_helper.rb
```diff
@@ -0,0 +1,61 @@
...
+    dockerfile = IO.read(dockerfile_path)
+    @image = Docker::Image.build(dockerfile)
+    @container = Docker::Container.create({
+      'Image' => @image.id,
+      'Hostname' => "#{name}-#{image.id}",
+    })
+
+    @container.start({"PublishAllPorts" => true})
+    @host[:ports] = @container.json["NetworkSettings"]["Ports"]
+    puts "Provisioned"
```
purbon: This code here looks to me like debugging code leftover :-)

spec/integration/docker_rspec_helper.rb
```diff
@@ -0,0 +1,61 @@
+class RspecDocker
+  def initialize
```
purbon: A new line would benefit readability here.

On the other side, (disclaimer I have zero experience with docker, so please educate me if I ask something silly). I wonder if the task in this constructor can actually fail for some reason? if this an option? or will they always run smoothly.

I ask this because in my experience having constructors with behaviour tend to be error prone and a wrong design, it's a lot better to have this as another procedure in the class. Makes sense what I'm saying in this case? please educate me if not :-1:

spec/integration/docker_rspec_helper.rb
```diff
@@ -0,0 +1,61 @@
+class RspecDocker
```
purbon: I would rather name this as ```DockerHelper```, as this class is independent of the specs and is in reality helping setup and manage docker.

On the other side, I tend to see helpers living in their own directory, but this is mostly an organizational taste of mine, what do you think?

spec/integration/docker_rspec_helper.rb
```diff
@@ -0,0 +1,61 @@
...
+    @host = {
+      :host_ip => get_host_ip,
+      :ports => {}
+    }
+    @container = nil
+    @image = nil
+  end
+
+  def get_host_ip
+    # Let the crazy one-liner definition begin:
```
purbon: Is this long comment really meaningful, could probably be simplified ?

spec/integration/docker_rspec_helper.rb
```diff
@@ -0,0 +1,61 @@
...
+    if @container
+      @container.stop
+      @container.delete
+    end
+    if @image
+      @image.remove(:force => true)
+    end
+  end
+
+  attr_reader :host
```
purbon: why not having this at the beginning of the class? they usually go there.

spec/integration/outputs/elasticsearch/node_spec.rb
```diff
@@ -0,0 +1,35 @@
+require "spec/es_spec_helper"
```
purbon: you should call this using ```require_relative``` otherwise when running the test from logstash it will fail.

spec/integration/outputs/elasticsearch/node_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+  # Reference: http://www.elasticsearch.org/guide/reference/modules/discovery/zen/
+
+  it "should support hosts in both string and array" do
+    # Because we defined *hosts* method in NodeClient as private,
+    # we use *obj.send :method,[args...]* to call method *hosts*
+    client = LogStash::Outputs::Elasticsearch::Protocols::NodeClient.new(:host => "localhost")
+
+    # Node client should support host in string
+    # Case 1: default :host in string
+    insist { client.send :hosts, :host => "host",:port => 9300 } == "host:9300"
```
purbon: please avoid using insist, please :pray: :pray: :pray: :pray:, you can use so many and nice things coming from rspec3 matchers :-1: !!!

spec/integration/outputs/elasticsearch/node_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+require "logstash/outputs/elasticsearch/protocol"
+
+describe "elasticsearch node client" do
+  # Test ElasticSearch Node Client
+  # Reference: http://www.elasticsearch.org/guide/reference/modules/discovery/zen/
+
+  it "should support hosts in both string and array" do
+    # Because we defined *hosts* method in NodeClient as private,
+    # we use *obj.send :method,[args...]* to call method *hosts*
+    client = LogStash::Outputs::Elasticsearch::Protocols::NodeClient.new(:host => "localhost")
```
purbon: this would actually live more in a let variable, or even better as a subject of the test.

spec/integration/outputs/elasticsearch/node_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+
+    # Node client should support host in string
+    # Case 1: default :host in string
+    insist { client.send :hosts, :host => "host",:port => 9300 } == "host:9300"
+    # Case 2: :port =~ /^\d+_\d+$/
+    insist { client.send :hosts, :host => "host",:port => "9300-9302"} == "host:9300,host:9301,host:9302"
```
purbon: I would rather break all cases in a different test example as all test different things.

spec/integration/outputs/index_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+
+shared_examples "an indexer" do
+    let(:index) { 10.times.collect { rand(10).to_s }.join("") }
+    let(:type) { 10.times.collect { rand(10).to_s }.join("") }
+    let(:event_count) { 10000 + rand(500) }
+    let(:flush_size) { rand(200) + 1 }
+    let(:config) { "not implemented" }
+
+    it "ships events" do
+      insist { config } != "not implemented"
```
purbon: same thing here, please avoid using insist as you can do this by using except very easy. :pray: :pray: :pray:

spec/integration/outputs/index_spec.rb
```diff
@@ -0,0 +1,90 @@
+require "spec/es_spec_helper"
+
+shared_examples "an indexer" do
```
purbon: like the idea of using shared_examples when necessary, good job!

spec/integration/outputs/index_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+    let(:flush_size) { rand(200) + 1 }
+    let(:config) { "not implemented" }
+
+    it "ships events" do
+      insist { config } != "not implemented"
+
+      pipeline = LogStash::Pipeline.new(config)
+      pipeline.run
+
+      index_url = "http://#{@es_server[:host]}:#{@es_server[:port]['http']}/#{index}"
```
purbon: I would rather provide most of the data to build this from the helpers classes and then avoid using any kind of class variables in the test if not strictly necessary.

spec/integration/outputs/transport_create_spec.rb
```diff
@@ -0,0 +1,94 @@
+require "spec/es_spec_helper"
+
+describe "transport client create actions" do
+  require "logstash/outputs/elasticsearch"
+  require "elasticsearch"
+
+  def get_es_output(action, id = nil)
```
purbon: why not moving this to a lovely ```helper``` function?

purbon:
@talevy did a first iteration of comments/review, let me know your take on them, happy to go more in detail to some of them.

purbon:
on the other side, this PR would need a rabase most likely :+1:

spec/integration/outputs/elasticsearch/node_spec.rb
talevy: I can follow-up with a mass rspec3 migration and get rid of insist. until then I'd rather leave it consistent during this specific refactor.

spec/integration/outputs/transport_create_spec.rb
talevy: it is very specific to transport client test.

I will go ahead and add a similar function to the helper to create a new plugin instance

talevy: hopefully one that can be re-used by retry/templates/transport_create spec tests. all can benefit from such a method.

talevy:
I have made some changes. I have removed my RspecDocker stuff in favor of the Longshoreman gem that Aaron and I have merged our efforts into.

talevy:
Specifically, 

- I have moved unit tests into a `spec/unit` directory
- removed the `.rspec` filtering for default spec files
- re-added `:integration` tag to the integration tests even though they are in an integration directory, makes for more explicit and straightforward filtering in the `RSpec.Configure

I think this is a fairly reasonable structure going forward for all our projects. what do you think?

spec/integration/outputs/index_spec.rb
andrewvc: +1

spec/unit/outputs/elasticsearch_secure_spec.rb
```diff
@@ -0,0 +1,51 @@
+require "spec/es_spec_helper"
```
andrewvc: Might be nice to name this spec elasticsearch_ssl.rb , a little more direct than 'secure', but not a big deal.

andrewvc:
This all looks pretty good so far to me, in so far as its a marked improvement over what's currently there. It'd be great to merge this in as the basis for some of the future patches I'm working on. I'd rather put my tests in this structure than what's currently in master. 

I think this is a great point to merge, and tackle additional issues in subsequent PRs.

spec/integration/outputs/elasticsearch/node_spec.rb
talevy: this breaks for me...

spec/integration/outputs/index_spec.rb
talevy: thanks!

spec/integration/outputs/index_spec.rb
talevy: I'll go ahead and make another PR with updating these usages

spec/unit/outputs/elasticsearch_secure_spec.rb
talevy: mmm. I can get behind that. I'll rename

spec/integration/outputs/index_spec.rb
talevy: agreed. found a potential fix. testing now and will update

talevy:
thanks @andrewvc @purbon. updated.

I think things like moving away from `insist` can be tackled at a later PR.

another TODO is making the `:elasticsearch_secure` tests compatible with this new integration setup. currently those tests are left alone as `:elasticsearch_secure` specs

spec/es_spec_helper.rb
```diff
@@ -0,0 +1,42 @@
...
+require "longshoreman"
+
+module ESHelper
+  def get_host
+    Longshoreman.new.get_host_ip
+  end
+
+  def get_port(protocol)
+    container = Longshoreman::Container.new
+    container.get("elasticsearch")
```
untergeek: I highly recommend using some kind of pre-defined constant that has something random in it instead of a statically named "elasticsearch."

The possibility of containers colliding because of identical naming is real.  This is why I assign a constant with a random bit in the Zabbix RSpec tests, e.g. `NAME = "logstash-output-zabbix-#{rand(999).to_s}"`, and then using `container.get(NAME)` and `@docker = Longshoreman.new("elasticsearch:1.6", NAME)`

andrewvc: +1

spec/es_spec_helper.rb
```diff
@@ -0,0 +1,42 @@
...
+  def get_client
+    Elasticsearch::Client.new(:host => "#{get_host}:#{get_port('http')}")
+  end
+end
+
+RSpec.configure do |config|
+  config.include ESHelper
+
+  config.before(:suite, :integration => true) do
+    @docker = Longshoreman.new("elasticsearch:1.6", "elasticsearch")
```
untergeek: Having the image and tag named in a constant is my preferred way to do it:

```
    @docker = Longshoreman.new("#{IMAGE}:#{TAG}", NAME)
```

It makes it less likely you'll forget to bump a version, having it defined as a constant near the top.

spec/es_spec_helper.rb
```diff
@@ -0,0 +1,42 @@
...
+    # TODO(talevy): detect when ES is up and ready instead of sleeping
+    # an arbitrary amount
+    sleep 10
+  end
+
+  config.after(:suite, :integration => true) do
+    @docker.cleanup
```
untergeek: `.cleanup` is THE BOMB :+1:

andrewvc:
@talevy looks fine to me. I agree with @untergeek 's comments

untergeek:
@talevy Same as @andrewvc.  LGTM, with my minor comments.

talevy:
updated. also switched to `require_relative`

untergeek:
lgtm!

PR #152, opened 2015-05-21
Additional notes regarding keeping CouchDB _ids intact.

Grokling:
Took me quite a while to find the magic value, so I propose changing the docs so that the answer is readily found.

### DIFF
--- a/lib/logstash/outputs/elasticsearch.rb
+++ b/lib/logstash/outputs/elasticsearch.rb
@@ -87,6 +87,9 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
 
   # The document ID for the index. Useful for overwriting existing entries in
   # Elasticsearch with the same ID.
+  # Unless specified, Elasticsearch will generate it's own ID.
+  # When passing a document from couchdb to Elasticsearch (via logstash), 
+  # use: '''document_id => "%{[@metadata][_id]}"''' to preserve the couchDB provided _id.
   config :document_id, :validate => :string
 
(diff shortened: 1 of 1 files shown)


Grokling:
Can I suggest perhaps putting the information in both places? It's relevant to the logstash-input-couchdb_changes plugin, but more directly affects logstash-output-elasticsearch. (as in 'How do I get the original document _id to appear in elasticsearch results?) 

I couldn't find the information anywhere - if putting it in both places saves the next guy two hours of trial and error and a GH issue, then that's a win.

Also, potentially this affects other inputs sources coming via logstash which use an underscore prefix on their id?

untergeek:
We failed to thoroughly document the need to do `document_id => "%{[@metadata][_id]}"` for the couchdb_changes plugin.  However, this is specific to the couchdb_changes plugin, and perhaps one or two others (at least, for now).  

> Also, potentially this affects other inputs sources coming via logstash which use an underscore prefix on their id?

Sort of.  There are protected field names in Elastic, like `_id`, but when they appear, you'd still have to use `document_id => "%{_id}"` to make the Elasticsearch output assign the field properly.  This is because of where the index API expects a manually assigned `_id` to be.  This is why I approve of the first line being included, but not the second two.

Please do open a PR with documentation added to the couchdb_changes plugin indicating the need to configure the Elasticsearch output plugin accordingly.

andrewvc:
I agree with @untergeek that these comments belong in the couchdb_changes plugin. While I sympathize with the issue you've hit @Grokling we have to balance that with verbosity in the documentation here.

Issue #160, opened 2015-05-27
Add back option to disable_verify for ssl usecase


suyograo:
We used to have an option to disable verification of SSL certificate called `disable_verify` and it was removed because that defeats the purpose of having security enabled. I agree in general, but this may help quick POCs and development setups so I am wondering if we can discuss this option here. 

I know we should not be exposing a half solution for security, but maybe call it `ssl_verify_will_disable_security` or something explicit to warn users. Any feedback?

scholarj:
I'd like to add that all HTTPs clients I know of not only have an option to disable cert verification but apparently it's so commonly requested that it's often front and center in the documentation - not some obscure setting you have to hunt for. Ex: web browsers give a stern warning but most have a link right in the warning for continuing and adding an exception for the given host.

IMO the option should be presented to the user for a decision not made for them. I could see making the case that you cannot trust an average web browser user to make a good decision when faced with the overhead of adding a CA cert to their configuration. But certainly a logstash administrator is qualified to make such a decision for their deployment/testing.

kevinkluge:
I like the idea of being able to disable.  But the default should be enabled.  We should push people to a secure config.

Another personal preference -- make the config name include "enable".  Otherwise, with "...disable...", you get double negatives and people have to think too much.

jordansissel:
@kevinkluge +1 on avoiding double negatives. We can also try to standardize the setting's name across all plugins that use SSL and also support destroying trust of SSL.

jordansissel:
My rough proposal is:
* A simple-named setting `ssl_certificate_verification => false` or something
* A warning emitted to users that this setting disables trust and makes SSL insecure

suyograo:
@jordansissel +1 for `ssl_certificate_verification`

suyograo:
Fixed in #167

PR #167, opened 2015-06-04
add the option to disable certificate verification

jsvd:
solves #160

### DIFF
--- a/lib/logstash/outputs/elasticsearch.rb
+++ b/lib/logstash/outputs/elasticsearch.rb
@@ -216,6 +216,11 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
   # Enable SSL
   config :ssl, :validate => :boolean, :default => false
 
+  # Validate the server's certificate
+  # Disabling this severely compromises security
+  # For more information read https://www.cs.utexas.edu/~shmat/shmat_ccs12.pdf
+  config :ssl_certificate_verification, :validate => :boolean, :default => true
+
   # The .cer or .pem file to validate the server's certificate
   config :cacert, :validate => :path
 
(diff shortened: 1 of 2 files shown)


lib/logstash/outputs/elasticsearch.rb
```diff
@@ -216,6 +216,9 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
   # Enable SSL
   config :ssl, :validate => :boolean, :default => false
 
+  # Enable SSL
```
talevy: can we update message here to be more descriptive to what this config does?

suyograo: ++, this gets generated to the docs directly

spec/outputs/elasticsearch_spec.rb
```diff
@@ -900,6 +900,34 @@ def mock_actions_with_response(*resp)
...
       end
     end
+
+    context "when using http protocol" do
+      context "when using ssl without cert verification" do
+        protocol = "http"
+        subject do
+          require "logstash/outputs/elasticsearch"
+          settings = {
+            "protocol" => protocol,
```
talevy: why not hardcode `http` here if no other is tested, and it is even mentioned in the higher level context

spec/outputs/elasticsearch_spec.rb
```diff
@@ -900,6 +900,34 @@ def mock_actions_with_response(*resp)
         end
       end
     end
+
+    context "when using http protocol" do
+      context "when using ssl without cert verification" do
```
talevy: is having a separate context within a context necessary?

jsvd: not necessary but better categorizes the test. This is all in the hopes that more tests will come and the outer context will have other contexts/tests.

lib/logstash/outputs/elasticsearch.rb
jsvd: oops missed that, well spotted

talevy:
LGTM

talevy:
jenkins, please test this.

talevy:
please test this, jenkins.

talevy:
please jenkins, test this.

talevy:
please jenkins test this

PR #176, opened 2015-06-16
Http Path support

### LINKED ISSUE #168: feauture request. Configurable path for HTTP protocol
In my environment I'm required to have logstash connect to ES via a reverse HTTP proxy which is not mounted on the root URL.

I'm able to get what i want in a hacky way because of lib/logstash/outputs/elasticsearch/protocol.rb line 64 and the fact that 'port' is a string
```
uri = "#{options[:protocol]}://#{options[:host]}:#{options[:port]}"
````
In my config I can say:
```
output { elasticsearch { 
  host => "my_host" 
  protocol => 'http'
  port => "443/my-path/"
  ssl => true		  
}}
```
As far as I can see there is currently no other way to achieve this. It would be nice if this becomes an official supported feature. If you like I can provide the patch, but I'm afraid I don't have the time to do so anytime soon. For someone who is already familiar with the Logstash plugin architecture it should be a piece of cake.

### PR DESCRIPTION
**Pull request by andrewvc**
This fixes #168 , adding HTTP Path support. The unit tests here just check that the params are correct and are well formed(ish). I hand tested this against an Nginx proxy->elasticsearch, and it seems to work correctly. The manual test config I used was:

```
input {
  generator {
    lines => [
      "test 1",
      "test 2"
    ]
    count => 1
  }
}

output {
  elasticsearch {
          host => "127.0.0.1"
          port => 8080
          path => "elastic-proxy"
          protocol => "http"
        }

}

```

### CODE REVIEW

File: lib/logstash/outputs/elasticsearch.rb [hunk 1]
```diff
@@ -263,6 +265,10 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
   config :user, :validate => :string
   config :password, :validate => :password
 
+  # HTTP Path at which the Elasticsearch server lives. Use this if you must run ES behind a proxy that remaps
+  # the root path for the Elasticsearch HTTP API lives. This option is ignored for non-HTTP transports.
+  config :path, :validate => :string, :default => "/"
```
jordansissel: I think we should be explicit in saying this option is invalid, not just ignored, for non-HTTP Transports. Further, we should reject a configuration of transport or nodeclient with a path setting on startup.

lib/logstash/outputs/elasticsearch.rb
```diff
@@ -62,6 +62,8 @@
 #   the max interval time is reached, which is set in :retry_max_interval.
 # - Events which are not retryable or have reached their max retry count are logged to stderr.
 class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
+  attr_reader :client
```
jordansissel: I don't think we need to expose this, do we?

andrewvc: That was added in for testing. I could use an attr getter, but I think it's generally better to add an accessor.

lib/logstash/outputs/elasticsearch.rb
```diff
@@ -314,8 +320,12 @@ def register
...
     end
 
-    if @protocol == "http" && @action == "create_unless_exists"
-      raise(LogStash::ConfigurationError, "action => 'create_unless_exists' is not supported under the HTTP protocol");
+    if @protocol == "http"
+      if @action == "create_unless_exists"
+        raise(LogStash::ConfigurationError, "action => 'create_unless_exists' is not supported under the HTTP protocol");
+      end
+
+      client_settings[:path] = "/#{@path}/".gsub(/\/\//, "/") # Normalize slashes
```
jordansissel: let's log that this normalization happens, something like:

```
@logger.debug? && @logger.debug("Normalizing http path", :path => @path, :normalized => client_settings[:path])
```

lib/logstash/outputs/elasticsearch.rb
andrewvc: It might be inconvenient if someone's just trying out one transport vs. the other. Could be handy to leave it in? I can remove it if you think that doesn't make sense however.

spec/outputs/elasticsearch_spec.rb
```diff
@@ -306,6 +306,56 @@
...
+        "protocol" => "http",
+        "index" => "my-index",
+        "host" => "localhost",
+        "path" => "some-path"
+      }
+    }
+
+    let(:eso) {LogStash::Outputs::ElasticSearch.new(options)}
+
+    let(:manticore_host) {
```
jordansissel: Instead of exposing the client through eso, maybe we just have a reader method that computes the URL based on the configuration?

Then we can do this:
```ruby
eso = LogStash::Outputs::ElasticSearch.new(...)
...
expect(eso.http_url).to(be == "http://whatever/something")
```

This feels closer to a unit test. If we want to test through Manticore, I feel like we should setup a mock http server to point Manticore at and make assertions about its behavior.

Thoughts?

lib/logstash/outputs/elasticsearch.rb
andrewvc: Fixed!

spec/outputs/elasticsearch_spec.rb
andrewvc: Well, it's a little tricky given that the options are scoped entirely to the register function. To make this work with a zero-arity `http_url` method would require saving the options partway through in an instance var then computing the `http_url`. 

I think, given the code structure we have, the url should probably be computed during options parsing and saved to an `@options` instance var (currently options is method scoped). We could then expose that through an accessor to make all the processed options testable. 

Thoughts?

It might also make sense to do a small refactor and split up `#register` a bit to make this understandable. One to idempotently process the config values and return the options hash, another to build the client. It might be nice for register to look like:

```ruby
attr_reader :options
def register
  #... mutex and other instance var setup
  @options = process_options
  build_client
  start_embedded_elasticsearch if @embedded
end
```

andrewvc:
@jordansissel let me know if these changes make any sense :)

lib/logstash/outputs/elasticsearch.rb
jordansissel: I agree that it may be inconvenient for folks just trying different transports. It's a bit of a pickle because we have many settings that don't work on differet protocols (bind_host/port isn't used in transport or http, for example).

It can stay how it is, and we can address it further with warnings or errors if users get confused by it. +1 leaving as is.

spec/outputs/elasticsearch_spec.rb
jordansissel: Agreed that it's tricky; leave it for now, we can improve this later once we figure out some better testing strategies (scheduled for the next few months on the early road to 1.6/2.0)

spec/outputs/elasticsearch_spec.rb
```diff
@@ -306,6 +306,56 @@
...
+      eso.client.first.send(:client).transport.options[:host].first
+    }
+
+    around(:each) do |block|
+      thread = eso.register
```
jordansissel: register shouldn't return anything, if it does, it's not expected.

The register method for this plugin is already pretty wild, I'll make a note that it needs some love to make it easier to manage and test.

spec/outputs/elasticsearch_spec.rb
```diff
@@ -306,6 +306,56 @@
...
+
+
+      it "should properly set the path on the HTTP client adding slashes" do
+        expect(manticore_host).to include("/" + options["path"] + "/")
+      end
+
+      context "with extra slashes" do
+        let(:path) { "/slashed-path/ "}
+        let(:eso) {
+          LogStash::Outputs::ElasticSearch.new(options.merge("path" => "/some-path/"))
```
jordansissel: Did you intend to set this `options.merge("path" => path)` ? (otherwise, I don't see the `path` let being used.

spec/outputs/elasticsearch_spec.rb
andrewvc: K! I just noticed that it happened to return a thread. It made sense to me to just go ahead and kill it. I'm not sure if that's necessary though. Do you have any ideas there? I assumed it was intended to either be join()ed or kill()ed. I'll look into why this is returning a thread.

spec/outputs/elasticsearch_spec.rb
andrewvc: I did! good catch.

spec/outputs/elasticsearch_spec.rb
jordansissel: It's returning a thread because of ruby's implicit return _and_ it was written long before @colinsurprenant convinced me that implicit return was good :P  -- my old assertion was that "methods that should return things should use the `return` statement" which makes any return value from the register method, under this idea, undefined/unsupported, even if accidental.

Best guess, for now, remove the `around(...) ...` block and the threads will die when rspec exits. We can refactor later for easier thread management once we know what/how things are interruptable.

andrewvc: @jordansissel Great. Thanks for all the insight. Will do :)

jordansissel:
Other than my last comment (https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/176#discussion_r32785644), LGTM. Looks like you may need to rebase since we just merged @talevy's epic test restructuring patch.

andrewvc:
Thanks @jordansissel will rebase then merge

andrewvc:
Jarvis didn't like my merge commits, moved here https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/178

PR #177, opened 2015-06-18
Refactor in support of id reuse on error

### LINKED ISSUE #171: When Retrying events, use supplied autogenerated ID from ES (when available)
When ES bulk response returns with errors, an id is associated with the failed event, we should assign this to the logstash event so it will re-use the id upon subsequent submits.

```
PUT _bulk
{ "index": {"_index":"index","_type":"type"}}
{"f":"1"}
```

```
{
  "took": 1,
  "errors": true,
  "items": [
      {
        "create": {
            "_index": "index",
            "_type": "type",
            "_id": "AU3j8SFVDtO3gVUsP8bb",
            "status": 500,
            "error": "ElasticsearchException[bla]"
        }
      }
  ]
}
```

in this example, we should reuse the id: `AU3j8SFVDtO3gVUsP8bb` and ignore any 'already_exists` errors from ES if ES did end up successfully indexing the event in question.

andrewvc:
This is an exploratory patch meant for discussion. Breaking up the methods along these lines should make writing tests much easier. @talevy what are your thoughts?

Fixes #171

### DIFF
--- a/lib/logstash/outputs/elasticsearch.rb
+++ b/lib/logstash/outputs/elasticsearch.rb
@@ -485,6 +485,9 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
 
   public
   def submit(actions)
+    require 'pry'
+    binding.pry
+    raise "whut"
     resp = submit_bulk_request(actions)
     handle_bulk_errors(resp)
     resp
@@ -493,7 +496,7 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
   private
   # Return the type of retry this status requires
   def bin_for_status(status)
-    if SUCCESS_OCDES.include?(status)
+    if SUCCESS_CODES.include?(status)
       :success
     elsif RETRYABLE_CODES.include?(status)
       :retryable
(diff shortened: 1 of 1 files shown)


lib/logstash/outputs/elasticsearch.rb
```diff
@@ -484,30 +484,69 @@ def receive(event)
...
+  def submit(actions)
+    resp = submit_bulk_request(actions)
+    handle_bulk_errors(resp)
+    resp
+  end
+
+  private
+  # Return the type of retry this status requires
+  def bin_for_status(status)
+    if SUCCESS_OCDES.include?(status)
```
talevy: typo

lib/logstash/outputs/elasticsearch.rb
```diff
@@ -484,30 +484,69 @@ def receive(event)
...
-      retry_push(actions_to_retry) unless actions_to_retry.empty?
-    end
+    bulk_response
+  end
+
+  private
+  def handle_bulk_errors(bulk_response)
+    return unless errors
+
+    retryable, fatal = group_statuses(actions,bulk_response['statuses']).values_at(:retryable,:fatal)
```
talevy: never seen this `values_at`. sneaky

andrewvc: @talevy I've been forever corrupted by clojure. Must.... destructure... all the things....

lib/logstash/outputs/elasticsearch.rb
andrewvc: Ah, good one

talevy:
I like this, `#submit` has grown difficult to decipher as it attempted to do more.

I may have missed something, but how does this fix #171?

andrewvc:
@talevy Does not fix it (yet)! Just wanted to make sure this approach made sense. I'll start adding in tests and then the new stuff around fixing #171. I want to have tests against these broken out methods first before I add code + tests around #171 . Thanks!

andrewvc:
This is just hopelessly out of date, I'm going to close this.

PR #178, opened 2015-06-19
HTTP Path support

andrewvc:
This is the exact same code as https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/176 
Jarvis didn't like my merge commits, so I squashed everything myself.


andrewvc:
Fixes #168

PR #180, opened 2015-06-19
fix before/after integration hooks

talevy:
:suite hook was not filtering based on :integration tag


spec/es_spec_helper.rb
```diff
@@ -34,14 +34,26 @@ def get_client
 RSpec.configure do |config|
   config.include ESHelper
 
-  config.before(:suite, :integration => true) do
-    @docker = Longshoreman.new("#{CONTAINER_IMAGE}:#{CONTAINER_TAG}", CONTAINER_NAME)
-    # TODO(talevy): detect when ES is up and ready instead of sleeping
-    # an arbitrary amount
-    sleep 10
+  config.before(:all, :integration => true) do
```
suyograo: Can we add a comment here why we use `:all` vs `:suite`? For the sake of the next person looking at it?

talevy: this `:all` hook gets run before every `describe` block that is tagged with `:integration => true`.

spec/es_spec_helper.rb
```diff
@@ -34,14 +34,26 @@ def get_client
...
+    # check if container exists already
+    begin
+      ls = Longshoreman::new
+      ls.container.get(CONTAINER_NAME)
+    rescue Docker::Error::NotFoundError
+      Longshoreman.new("#{CONTAINER_IMAGE}:#{CONTAINER_TAG}", CONTAINER_NAME)
```
suyograo: why are we creating a new container here and not using it?

spec/es_spec_helper.rb
```diff
@@ -34,14 +34,26 @@ def get_client
...
-    @docker = Longshoreman.new("#{CONTAINER_IMAGE}:#{CONTAINER_TAG}", CONTAINER_NAME)
-    # TODO(talevy): detect when ES is up and ready instead of sleeping
-    # an arbitrary amount
-    sleep 10
+  config.before(:all, :integration => true) do
+    # check if container exists already
+    begin
+      ls = Longshoreman::new
+      ls.container.get(CONTAINER_NAME)
+    rescue Docker::Error::NotFoundError
```
talevy: we want to make sure that the container is running before the `describe` block, but we also don't want to create a new one if one exists from the previous hook run

spec/es_spec_helper.rb
suyograo: ah never mind

spec/es_spec_helper.rb
```diff
@@ -34,14 +34,26 @@ def get_client
...
+      Longshoreman.new("#{CONTAINER_IMAGE}:#{CONTAINER_TAG}", CONTAINER_NAME)
+      # TODO(talevy): verify ES is running instead of static timeout
+      sleep 10
+    end
   end
 
-  config.after(:suite, :integration => true) do
-    @docker.cleanup
+  config.after(:suite) do
```
talevy: we want to do a final cleanup after all `:integration` runs... but we don't want to clean up before the last block... so we must cleanup after the suite has run. so this is a final blind check to see if the ES docker container is running and needs to be cleaned up. If no container can be found and/or docker is not running on the system, we do nothing

spec/es_spec_helper.rb
talevy: to reduce instance variables / preserving state between spec runs, all docker state is saved in the docker environment externally. So we create the container, give it a name, and always search for it  from the docker-api before doing something with it

spec/es_spec_helper.rb
suyograo: ok, can you add this as comments in the code? ^^

suyograo:
LGTM

PR #184, opened 2015-06-23
NodeClient does not get cleaned up during tests

talevy:
during the `node_spec.rb` tests, NodeClient does not get
cleaned-up correctly. Since we are just testing a utility
function within the class, I have changed the `#hosts` method
to be a class variable (still private). This should now let
`bundle exec rspec --tag integration` terminate.

### DIFF
--- a/lib/logstash/outputs/elasticsearch/protocol.rb
+++ b/lib/logstash/outputs/elasticsearch/protocol.rb
@@ -148,7 +148,7 @@ module LogStash::Outputs::Elasticsearch
         @settings = org.elasticsearch.common.settings.ImmutableSettings.settingsBuilder
         if options[:host]
           @settings.put("discovery.zen.ping.multicast.enabled", false)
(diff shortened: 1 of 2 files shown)


suyograo:
LGTM, did not run the tests

ph:
@talevy Not sure I understand the issue here, you were create a new instance of the `NodeClient` within the subject so It be correctly reset on every example?

talevy:
@ph not sure. but something about them was not being cleaned up. and the `hosts` is just a utility method which does not require the NodeClient to start-up, so technically now I can make this a unit test now that I think about it

ph:
@talevy The problem with class method is they are harder to refactor.. this is why I personally don't really like them. 

Can you add the rspec error you had to this issue?

talevy:
there was no rspec error.

all tests pass. but bundler wouldn't exit

ph:
LGTM.

Issue #170, opened 2015-06-09
Add support for client certificate based authentication


suyograo:
Today this output supports basic auth when used with `protocol=>http`. This feature request is to add client certificate based authentication for http. Should be straightforward to add using Manticore's support for passing ssl parameters. 

```
:ssl => {:keystore => "/path/to/keystore", :keystore_password => "password"}
```

suyograo:
Closed by https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/196

PR #196, opened 2015-07-10
allow the specification of client certificates

jsvd:
Certificates for client authentication by the server can be specified with the keystore argument, which can  be either a jks or p12 keystore.

Solves #170


andrewvc:
LGTM. Looking through the code however, I noticed that there's code in there to generate a .jks when cacert is specified. Manticore does this natively now, so this may not be required. Is there a reason for it here, or should I open a ticket to pull that out.

jsvd:
Please open a ticket to pull that out :)

suyograo:
Done https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/200

Issue #199, opened 2015-07-10
Add support for using forwarding proxy for http protocol


suyograo:
This plugin should support sending http requests through a forwarding proxy. With manticore client, it should be straightforward to pass proxy URL configs. See: https://github.com/cheald/manticore/blob/master/lib/manticore/client.rb#L11

We should also investigate if we can populate these settings using the `HTTP_PROXY` environment variable or through `JAVA_OPTS`

See discussion here: http://stackoverflow.com/questions/31283677/trying-to-configure-logstash-to-use-a-proxy-with-ls-java-opts-but-it-doesnt-see

and also:
https://discuss.elastic.co/t/shipping-logs-via-proxy-with-logstash/24723

PR #201, opened 2015-07-14
Forward proxy support for ES servers

andrewvc:
This adds support for http forward proxies. Tested with the following config:

Logstash:
```
input {
  stdin {}
}

output {
  elasticsearch {
    protocol => http
    proxy => "http://52.2.157.171:80"
    host => "127.0.0.1"`
  }
}
```

nginx
```
    server {
        listen       80;

        location / {
            resolver 8.8.8.8;
            proxy_pass http://$http_host$uri$is_args$args;
        }

        error_page   500 502 503 504  /50x.html;
        location = /50x.html {
            root   html;
  }
```

### DIFF
--- a/lib/logstash/outputs/elasticsearch.rb
+++ b/lib/logstash/outputs/elasticsearch.rb
@@ -309,6 +309,12 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
   # Set max interval between bulk retries
   config :retry_max_interval, :validate => :number, :default => 5
 
+  # Set the address of a forward HTTP proxy. Must be used with the 'http' protocol
+  # Can be either a string, such as 'http://localhost:123' or a hash in the form
+  # {host: 'proxy.org' port: 80 scheme: 'http'}
+  # Note, this is NOT a SOCKS proxy, but a plain HTTP proxy
+  config :proxy
+
   public
   def register
     @submit_mutex = Mutex.new
@@ -372,6 +378,7 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
     end
 
     client_settings.merge! setup_ssl()
+    client_settings.merge! setup_proxy()
 
     common_options = {
       :protocol => @protocol,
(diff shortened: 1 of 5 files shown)


lib/logstash/outputs/elasticsearch.rb
```diff
@@ -309,6 +309,12 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
   # Set max interval between bulk retries
   config :retry_max_interval, :validate => :number, :default => 5
 
+  # Set the address of a forward HTTP proxy. Must be used with the 'http' protocol
+  # Can be either a string, such as 'http://localhost:123' or a hash in the form
+  # {host: 'proxy.org' port: 80 scheme: 'http'}
+  # Note, this is NOT a SOCKS proxy, but a plain HTTP proxy
+  config :proxy
```
talevy: I know it doesn't do much, but might as well have it :)

```
conig :proxy, :validate => :string
```

suyograo:
@andrewvc any thoughts on populating this config from `HTTP_PROXY`env variable .. this may warrant a generic way to populate config from environment variables (outside of this PR) ... We could in theory use [logstash-filter-environment](https://www.elastic.co/guide/en/logstash/current/plugins-filters-environment.html)

andrewvc:
@suyograo my worry there is that there may be multiple forward proxies. IMHO that behavior might be surprising. Let's say, for instance, that you were using the http_poller + the elasticsearch output. I would likely guess you were using a forward proxy for one side of that pipeline but not the other (for instance, you probably would need the forward proxy to break out of the LAN for the polling requests, but not for the intranet accessible ES instance).

In that scenario the ES output would break without a strong apparent cause. I think that while it may cause the users to repeat themselves a bit more it follows the principle of least astonishment.

andrewvc:
Closes #199

suyograo:
@andrewvc makes sense. Lets leave it as is for now

suyograo:
LGTM

PR #212, opened 2015-07-22
Sync docs with logstash-docs repo

suyograo:
https://github.com/elastic/logstash-docs/blob/master/docs/plugins/outputs/elasticsearch.asciidoc


suyograo:
self merging because these changes are already rendered in https://www.elastic.co/guide/en/logstash/current/plugins-outputs-elasticsearch.html

Issue #172, opened 2015-06-11
Configuration option "host" is inconsistent with elasticsearch input "hosts" option.


joshuar:
It looks like the elasticsearch output plugin uses [`host`](https://www.elastic.co/guide/en/logstash/current/plugins-outputs-elasticsearch.html#plugins-outputs-elasticsearch-host) while the elasticsearch input plugin uses [`hosts`](https://www.elastic.co/guide/en/logstash/current/plugins-inputs-elasticsearch.html#plugins-inputs-elasticsearch-hosts).  Both options essentially do the same thing in the respective plugins.  It would be nice if the name was consistent in both plugins.

andrewvc:
This has been fixed in master, and will be released with LS2.0 :) Thanks for filing the issue @joshuar !

PR #216, opened 2015-07-30
Http only WIP

### LINKED ISSUE #150: RFC: Change the default protocol to http
The current protocol is `node`. This means Logstash joins the cluster. The cons of this `node` default are:

* bidirectional communication must be possible between Logstash and the _entire_ cluster.
* Logstash's NodeClient can serve as a client node and answer search queries. Probably not expected.
* Shield users cannot use NodeClient, anyway, without granting Logstash full access to the cluster as a member of the cluster (full privileges)
* The default NodeClient uses multicast for discovery, and some operating systems (like CentOS, I think?) and networks (like AWS EC2's network) block this by default.
* Exceptions in Elasticsearch are serialized over the network using the standard Java Serializable. In our experience, Java's Serializable isn't stable across different JVM versions (exceptions will fail to deserialize).

Changing to HTTP as the default would make it:
* easier to integrate with Shield
* easier to debug (http is simpler than the transport protocol, probably)
* independent of Elasticsearch ABI (transport serialization)
* independent of JVM version
* Logstash release could be shipped _without_ the Elasticsearch libraries (saves 10-20mb)

Downsides:
* We'd want to have a default `host` setting, `localhost` is probably OK.
* We need discovery and sniffing enabled in the http protocol also.

andrewvc:
This is a WIP for efforts to remove support for non-HTTP protocols 

Fixes #150


spec/es_spec_helper.rb
```diff
@@ -36,6 +31,7 @@ def get_client
 
   # this :all hook gets run before every describe block that is tagged with :integration => true.
   config.before(:all, :integration => true) do
+    require 'pry'; binding.pry
```
talevy: WIP //here just as a reminder

talevy:
we can take a look at simplifying the retry logic, since we do not need to normalize the response across protocols.

andrewvc:
@talevy what specific ideas did you have wrt cleaning up the retry code? Was this the Stud::Buffer cleanup you mentioned earlier we decided to forego?

BTW, I think this is getting close to good for an LGTM. I don't plan on merging this till we have logstash-output-elasticsearch_native ready

talevy:
@andrewvc 

Nothing for this PR, but I will make a follow-up PR to make the information nicer. I have the "normalize" step for the error message returned from a bulk request. this normalization was done mainly to help wrangle between the response from the java client and the http client. Now that we only have one, this step is not necessary.

andrewvc:
Mirrored the present content of this repo into https://github.com/logstash-plugins/logstash-output-elasticsearch_native to begin refactor there to remove HTTP.

andrewvc:
@talevy @jordansissel what are your thoughts on the current state of this plugin? What do we need to get to LGTM?

lib/logstash/outputs/elasticsearch/protocol.rb
```diff
@@ -1,307 +1,93 @@
...
-
-      def template_exists?(name)
-        @client.indices.get_template(:name => name)
-        return true
-      rescue Elasticsearch::Transport::Transport::Errors::NotFound
-        return false
-      end # def template_exists?
+    def bulk(actions)
+      bulk_response = @client.bulk(:body => actions.collect do |action, args, source|
+                                     if source
```
talevy: large indentation intended?

talevy:
@andrewvc running this plugin against elasticsearch-shield/secure tests should be a blocker. @jsvd, how can we run this against your elasticsearch-secure specs?

lib/logstash/outputs/elasticsearch/protocol.rb
andrewvc: ahhhh good catch.

lib/logstash/outputs/elasticsearch.rb
```diff
@@ -1,5 +1,5 @@
 # encoding: utf-8
-require "logstash/namespace"
+require "logstash/pipeline"
```
talevy: why this change?

lib/logstash/outputs/elasticsearch.rb
```diff
@@ -8,32 +8,20 @@
...
-# as configuration options, there are two methods:
-#
-# * Create an `elasticsearch.yml` file in the $PWD of the Logstash process
-# * Pass in es.* java properties (`java -Des.node.foo=` or `ruby -J-Des.node.foo=`)
-#
-# With the default `protocol` setting ("node"), this plugin will join your
-# Elasticsearch cluster as a client node, so it will show up in Elasticsearch's
-# cluster status.
+# We strongly encourage the use of HTTP over the node protocol. It is just as
+# fast and far easier to administer. For those wishing to use the java protocol please see the 'elasticsearch_java' gem.
```
talevy: we should probably provide a direct link here in the future.

lib/logstash/outputs/elasticsearch.rb
andrewvc: TBH I don't recall . I'll switch it back

lib/logstash/outputs/elasticsearch.rb
andrewvc: Yep, once we have a docs page up.

lib/logstash/outputs/elasticsearch.rb
```diff
@@ -165,44 +122,9 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
...
   config :host, :validate => :array
 
-  # The port for Elasticsearch transport to use.
-  #
-  # If you do not set this, the following defaults are used:
-  # * `protocol => http` - port 9200
-  # * `protocol => transport` - port 9300-9305
-  # * `protocol => node` - port 9300-9305
+  # You can set the remote port as part of the host, or explicitly here as well
   config :port, :validate => :string
```
talevy: we can default to `9200` now, no?

lib/logstash/outputs/elasticsearch.rb
```diff
@@ -257,6 +153,7 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
   # - index: indexes a document (an event from Logstash).
   # - delete: deletes a document by id
   # - create: indexes a document, fails if a document by that id already exists in the index.
+  # - update: updates a document by id
   # following action is not supported by HTTP protocol
   # - create_unless_exists: creates a document, fails if no id is provided
```
talevy: we can just remove this option from the the http-only repo entirely

lib/logstash/outputs/elasticsearch.rb
```diff
@@ -329,101 +237,41 @@ def register
...
-        client_settings["node.name"] = @node_name
-      else
-        client_settings["node.name"] = "logstash-#{Socket.gethostname}-#{$$}-#{object_id}"
-      end
-
-      @@plugins.each do |plugin|
-        name = plugin.name.split('-')[-1]
-        client_settings.merge!(LogStash::Outputs::ElasticSearch.const_get(name.capitalize).create_client_config(self))
-      end
+    if @action == "create_unless_exists"
```
talevy: see note above, but I think we can go ahead and remove this since it does not apply to http.

lib/logstash/outputs/elasticsearch.rb
```diff
@@ -329,101 +237,41 @@ def register
...
+    client_settings[:path] = "/#{@path}/".gsub(/\/+/, "/") # Normalize slashes
+    @logger.debug? && @logger.debug("Normalizing http path", :path => @path, :normalized => client_settings[:path])
 
-    if @port.nil?
-      @port = case @protocol
-        when "http"; "9200"
-        when "transport", "node"; "9300-9305"
-      end
-    end
+    @port ||= "9200"
```
talevy: I'd prefer having this type of logic via the `default` setting in the config declaration above

lib/logstash/outputs/elasticsearch.rb
```diff
@@ -329,101 +237,41 @@ def register
...
 
-    if @host.nil? && @protocol != "node" # node can use zen discovery
+    if @host.nil?
       @logger.info("No 'host' set in elasticsearch output. Defaulting to localhost")
       @host = ["localhost"]
```
talevy: does it even make sense to ever have host nil in the http-only world? we can just default this to localhost above and get rid of this logic. what do you think?

spec/es_spec_helper.rb
```diff
@@ -9,57 +9,61 @@
 CONTAINER_IMAGE = "elasticsearch"
 CONTAINER_TAG = "1.6"
 
-module ESHelper
+DOCKER_INTEGRATION = ENV["DOCKER_INTEGRATION"]
```
talevy: should we also allow environment variables to explicitly target an ES instance with custom host/port?

this can wait for another follow-up PR if we think its a useful option

spec/es_spec_helper.rb
```diff
@@ -9,57 +9,61 @@
...
+    config.before(:all, :integration => true) do
+
+
+      # check if container exists already before creating new one.
+      begin
+        ls = Longshoreman::new
+        ls.container.get(CONTAINER_NAME)
+      rescue Docker::Error::NotFoundError
+        Longshoreman.new("#{CONTAINER_IMAGE}:#{CONTAINER_TAG}", CONTAINER_NAME)
+        # TODO(talevy): verify ES is running instead of static timeout
```
talevy: making a note here for a possible follow-up PR.

maybe we can add an endpoint in the http_client wrapper we have going to check if ES is up. then we can loop-tick the check here until it says yes here as a replacement of the long-optimistic 10-second wait here

spec/es_spec_helper.rb
andrewvc: I think it's not a great benefit at the moment. YAGNI IMHO

spec/unit/outputs/elasticsearch_spec.rb
```diff
@@ -2,48 +2,18 @@
 
 describe "outputs/elasticsearch" do
   context "registration" do
-    it "should register" do
-      output = LogStash::Plugin.lookup("output", "elasticsearch").new("embedded" => "false", "protocol" => "transport", "manage_template" => "false")
-      # register will try to load jars and raise if it cannot find jars
-      expect {output.register}.to_not raise_error
-    end
-
     it "should fail to register when protocol => http, action => create_unless_exists" do
```
talevy: if we decide to remove this option from the vocabulary of this plugin, we can remove this test as well. woohoo!

spec/unit/outputs/elasticsearch_ssl_spec.rb
```diff
@@ -2,81 +2,56 @@
...
-          "ssl" => true,
-          "keystore" => keystore_path,
-          "keystore_password" => "test"
-        }
-        next LogStash::Outputs::ElasticSearch.new(settings)
-      end
+    subject do
+      require "logstash/outputs/elasticsearch"
+      settings = {
+        "host" => "node01",
```
talevy: maybe now is a good time (another possible follow-up PR!) to work with @jsvd to finish up codifying this setup from within the plugin tests so that we can run these locally as well.

andrewvc: I'm missing where/how this host is defined. Any context @talevy ?

spec/es_spec_helper.rb
talevy: sounds good, I don't feel strongly about it

spec/unit/outputs/elasticsearch_ssl_spec.rb
talevy: thats just it, the context is not localized to this repo. it is a custom setup that is done for internal testing

talevy:
LGTM

lib/logstash/outputs/elasticsearch.rb
```diff
@@ -8,32 +8,20 @@
...
-require 'logstash-output-elasticsearch_jars.rb'
+require "logstash/outputs/elasticsearch/http_client"
 
 # This output lets you store logs in Elasticsearch and is the most recommended
 # output for Logstash. If you plan on using the Kibana web interface, you'll
-# need to use this output.
+# want to use this output.
 #
-# *VERSION NOTE*: Your Elasticsearch cluster must be running Elasticsearch 1.0.0 or later.
+# This output only speaks the HTTP, which is the preferred protocol for interacting with elasticsearch. By default
```
suyograo: Can we capitalize elasticsearch?

andrewvc: For sure! (/me remembers when in all official copy Elasticsearch was always lowercased)

suyograo: sorry, nit pick :)

lib/logstash/outputs/elasticsearch.rb
```diff
@@ -165,43 +122,8 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
 
   config :host, :validate => :array
```
suyograo: @andrewvc what are your thoughts on deprecating `host` and introducing `hosts`. See https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/172

andrewvc: We should definitely put this in before 2.0

lib/logstash/outputs/elasticsearch.rb
```diff
@@ -165,43 +122,8 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
...
-  # This setting no longer does anything. It exists to keep config validation
-  # from failing. It will be removed in future versions.
-  config :max_inflight_requests, :validate => :number, :default => 50, :deprecated => true
-
-  # The node name Elasticsearch will use when joining a cluster.
-  #
-  # By default, this is generated internally by the ES client.
-  config :node_name, :validate => :string
+  # You can set the remote port as part of the host, or explicitly here as well
+  config :port, :validate => :string, :default => 9200
```
suyograo: do we need to deprecate or remove this if we support `host:port`. Just an opportunity to clean up here..

spec/es_spec_helper.rb
suyograo: @talevy I would create an issue for this. This will get lost otherwise

andrewvc:
This has been merged in #240

PR #226, opened 2015-08-18
update jars and tests to work for es2.0

talevy:
WIP

not sure what to do with `path.home` setting that now (in 2.0) throws exception if not set.

lots of jar dependency changes as ES is trying to reduce the number of shadowed jars


suyograo:
Tested ES 2.0 beta with existing LS version (1.5.3). Transport client breaks with this error:

```
exception caught on t
ransport layer [[id: 0x27c935df, /127.0.0.1:49223 => /127.0.0.1:9300]], closing connection
java.lang.IllegalStateException: Message not fully read (request) for requestId [9], action [clust
er/nodes/info], readerIndex [39] vs expected [57];
```

suyograo:
| What        | ES cluster          | Transport and node protocol  |
| ------------- |:-------------:| -----|
| LS 1.5 with ES 1.7 jars       | 2.0 | Does not work |
| LS 1.5 with ES 2.0 jars       |  2.0     |   Works |
| LS 1.5 with ES 2.0 jars  | 1.7     |    Does not work |

suyograo:
This has been ported to elasticsearch-java plugin

Issue #219, opened 2015-08-06
Better User Experience When Elasticsearch is Down


andrewvc:
Currently if elasticsearch dies weird errors can occur. We should perform some sort of sniffing to check if ES is up, and report it ASAP to the user in a very noticeable way. This was moved from this thread: https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/141#issuecomment-127418528

talevy:
++

andrewvc:
One of the current issues is that you just get generic errors around sending events to ES not working. It'd be better to give users a reason why these errors are happening. I think if we say "We can't talk to Elasticsearch at this URL" that will be much more helpful.

Additionally, on register we could check if say `/_cluster/health` can be reached to give an early warning, before any events are sent, that perhaps we may not be able to hit our endpoint.

So, I propose two changes:

1. Change "got error to send bulk" error messaging to say something like "Could not write event '{}' through bulk API to http://server:port/_bulk" which will help admins diagnose the issue better.
2. Add a ping healthcheck on register that will detect if ES is reachable or not with strong messaging around the error. Something like "Elasticsearch output healthcheck failed. Elasticsearch is not reachable at http://server:port/_cluster/health. Are you sure you've configured your Elasticsearch settings correctly?". We could even set a flag if the healthcheck passes and make a note of this on "got error to send to bulk" errors to say "a healthcheck failed on plugin startup as well".

jordansissel:
Agreed on helping provide stronger diagnostics from ES on errors (specific events, or network faults, or cluster problems)

andrewvc:
@jordansissel do you think the two pronged approach I mentioned is the right path here?

jordansissel:
@andrewvc for item 1; I think showing which event(s) failed and each specific failure for that event (I think the bulk response has the per-doc failure exceptions we could show to users). For item 2, having a background health check going on to check cluster health could be useful indeed to help better-inform any reports to users. +1 on both changes. I suspect we'll add more in the future, but this is a great start :)

andrewvc:
@jordansissel thanks! I'll get started then :)

hkarpf:
+1 on this, as I'm having read timeout issues, but my cluster is up and running.  Does it continue to try and output to elasticsearch, e.g. a retry?

andrewvc:
This has been a part of the ES output since LS5.x. Closing.

PR #227, opened 2015-08-20
Better error reporting

andrewvc:
This addresses #219 by improving error messages in the event that ES is down. 

It also cleans up the HTTP support by moving protocol.rb to http_client.rb

### DIFF
--- a/lib/logstash/outputs/elasticsearch.rb
+++ b/lib/logstash/outputs/elasticsearch.rb
@@ -8,6 +8,8 @@ require "stud/buffer"
 require "socket" # for Socket.gethostname
 require "thread" # for safe queueing
 require "uri" # for escaping user input
+require "logstash/outputs/elasticsearch/http_client"
+
 
 # This output lets you store logs in Elasticsearch and is the most recommended
 # output for Logstash. If you plan on using the Kibana web interface, you'll
@@ -249,8 +251,8 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
       (_host,_port) = host.split ":"
       options = { :host => _host, :port => _port || @port }.merge(common_options)
 
-      @logger.info "Create client to elasticsearch server on #{_host}:#{_port}"
-      LogStash::Outputs::Elasticsearch::HTTPClient.new(options)
+      @logger.info "Create client for elasticsearch server on #{_host}:#{_port}"
+      LogStash::Outputs::Elasticsearch::HttpClient.new(options)
     end
 
     if @manage_template
@@ -368,9 +370,31 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
   def flush(actions, teardown = false)
     begin
       submit(actions)
+    rescue Manticore::SocketException => e
+      # If we can't even connect to the server let's just print out the URL (:host is actually a URL)
+      # and let the user sort it out from there
+      @logger.error(
+        "Attempted to send a bulk request to Elasticsearch configured at '#{@current_client.client_options[:host]}',"+
+          " but Elasticsearch appears to be unreachable or down!",
+        :client_config => @current_client.client_options,
+        :error_message => e.message,
+        :error_class => e.class.name,
+        :backtrace => e.backtrace
+      )
+      @logger.debug("Failed actions for last bad bulk request!", actions: actions)
     rescue => e
-      @logger.error "Got error to send bulk of actions: #{e.message}"
-      raise e
+      # For all other errors print out full connection issues
+      @logger.error(
+        "Attempted to send a bulk request to Elasticsearch configured at '#{@current_client.client_options[:host]}'," +
+            " but an error occurred and it failed! Are you sure you can reach elasticsearch from this machine using " +
+          "the configuration provided?",
+        :client_config => @current_client.client_options,
+        :error_message => e.message,
+        :error_class => e.class.name,
+        :backtrace => e.backtrace
+      )
+
+      @logger.debug("Failed actions for last bad bulk request!", actions: actions)
     ensure
       @logger.debug? and @logger.debug "Shifting current elasticsearch client"
       shift_client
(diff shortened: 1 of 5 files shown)


talevy:
code looks good, but I don't think tests will pass given the change `HTTPClient` -> `HttpClient`

one example spot: https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/master/spec/integration/outputs/retry_spec.rb#L14-L15

andrewvc:
@talevy good catch. I changed that at the last minute before push (my bad). Fixing.....

andrewvc:
@talevy fixed. If this gets an LGTM I'll merge it directly into my http-only branch

lib/logstash/outputs/elasticsearch.rb
```diff
@@ -368,9 +370,31 @@ def submit(actions)
...
+    rescue Manticore::SocketException => e
+      # If we can't even connect to the server let's just print out the URL (:host is actually a URL)
+      # and let the user sort it out from there
+      @logger.error(
+        "Attempted to send a bulk request to Elasticsearch configured at '#{@current_client.client_options[:host]}',"+
+          " but Elasticsearch appears to be unreachable or down!",
+        :client_config => @current_client.client_options,
+        :error_message => e.message,
+        :error_class => e.class.name,
+        :backtrace => e.backtrace
```
jordansissel: I'm hesitant to show the backtrace by default, especially in an error message. My tendency is to hide backtraces because they confuse users and make helping them harder (anecdotally).

I'm open to including it here, but tell me a good story about it? :)

lib/logstash/outputs/elasticsearch.rb
```diff
@@ -368,9 +370,31 @@ def submit(actions)
...
+      )
+      @logger.debug("Failed actions for last bad bulk request!", actions: actions)
     rescue => e
-      @logger.error "Got error to send bulk of actions: #{e.message}"
-      raise e
+      # For all other errors print out full connection issues
+      @logger.error(
+        "Attempted to send a bulk request to Elasticsearch configured at '#{@current_client.client_options[:host]}'," +
+            " but an error occurred and it failed! Are you sure you can reach elasticsearch from this machine using " +
```
jordansissel: indentation changes across these 3 lines of string. Expected?

Also, for long messages like this, we should consider putting them in `locales/en.yml` and looking them up with I18n.t(...) - thoughts?

lib/logstash/outputs/elasticsearch.rb
```diff
@@ -368,9 +370,31 @@ def submit(actions)
...
+        :error_class => e.class.name,
+        :backtrace => e.backtrace
+      )
+      @logger.debug("Failed actions for last bad bulk request!", actions: actions)
     rescue => e
-      @logger.error "Got error to send bulk of actions: #{e.message}"
-      raise e
+      # For all other errors print out full connection issues
+      @logger.error(
+        "Attempted to send a bulk request to Elasticsearch configured at '#{@current_client.client_options[:host]}'," +
```
jordansissel: I trend towards having fixed strings + context, so in this case having a string interpolation inject the host  would be deviating from that. Any thoughts on moving the "current client host" to the context hash? :current_host => .... ?

lib/logstash/outputs/elasticsearch.rb
```diff
@@ -368,9 +370,31 @@ def submit(actions)
...
+            " but an error occurred and it failed! Are you sure you can reach elasticsearch from this machine using " +
+          "the configuration provided?",
+        :client_config => @current_client.client_options,
+        :error_message => e.message,
+        :error_class => e.class.name,
+        :backtrace => e.backtrace
+      )
+
+      @logger.debug("Failed actions for last bad bulk request!", actions: actions)
```
jordansissel: use hash rocket `:actions  => actions`

spec/integration/outputs/retry_spec.rb
```diff
@@ -11,7 +11,7 @@
   let(:max_retries) { 3 }
 
   def mock_actions_with_response(*resp)
-    LogStash::Outputs::Elasticsearch::HTTPClient
+    LogStash::Outputs::Elasticsearch::HttpClient
```
jordansissel: I'm more in favor of HTTPClient for camel casing because HTTP is an abbreviation that is naturally capitalized. If it weren't an abbreviation, it'd be HyperTextTransportProtocolClient, right? :)

LET US DO BATTLE>

No but seriously, I'm willing to accept this change and see how I feel about it over the long term, but my past/present preference is all-caps abbreviations for camel casing. Practically speaking, it has little-to-no effect on the code or its readability, and we should figure that out in our style discussion that other style ticket on logstash ;P

andrewvc: @jordansissel 

I have no strong feelings about this

lib/logstash/outputs/elasticsearch.rb
andrewvc: My rationale here is that this is a very common error that shows up in your support tickets often. I think its hard to evaluate it if there's any ambiguity in the messages, so including more info helps us troubleshoot issues for our users more quickly.

Additionally, people tend to blame logstash, but having the backtrace makes it more apparent more is at work here.

lib/logstash/outputs/elasticsearch.rb
andrewvc: Go go rubymine!

But errr yeah, I'll move it to the locales file.

lib/logstash/outputs/elasticsearch.rb
andrewvc: I think that we get a lot of mistaken error reports about this stuff so I wanted it to be less ambiguous. Putting it there says 'look at this variable first, is this the right address for ES? Is ES up?'.

lib/logstash/outputs/elasticsearch.rb
andrewvc: old habits die hard, will fix

spec/integration/outputs/retry_spec.rb
andrewvc: Fixed.

lib/logstash/outputs/elasticsearch.rb
jordansissel: Sweet, thanks for the perspective. I'm in favor of backtrace for this.

high five

PR #229, opened 2015-08-21
Add missing possible actions in the list

bboutier:
Just a little update of documentation.


andrewvc:
Thanks for the patch @bboutier , we wound up merging a different patch for this a while ago however!

PR #231, opened 2015-08-22
Update elasticsearch.rb

YiuTerran:
Just add a blank line...


andrewvc:
Thanks for the patch @YiuTerran , but we wound up rewriting that section of the docs in a separate patch anyway!

PR #236, opened 2015-08-27
Use the client for multi-host logic

andrewvc:
elasticsearch-ruby/Manticore is perfectly capable of handling multiple connections. This patch removes the extra logic we had in place to do this juggling ourselves. This should also handle conn failures better.


lib/logstash/outputs/elasticsearch.rb
```diff
@@ -258,31 +258,29 @@ def register
...
-      options = { :host => _host, :port => _port || @port }.merge(common_options)
-
-      @logger.info "Create client for elasticsearch server on #{_host}:#{_port}"
-      LogStash::Outputs::Elasticsearch::HttpClient.new(options)
+    option_hosts = @hosts.map do |host|
+      host_name, port = host.split(":")
+      { :host => host_name, :port => (port || @port).to_i }
     end
 
+    puts "WILL REG #{option_hosts}"
```
talevy: yup!

lib/logstash/outputs/elasticsearch/http_client.rb
```diff
@@ -76,7 +76,8 @@ def build_client(options)
         @client_options[:headers] = { "Authorization" => "Basic #{token}" }
       end
 
-      c = Elasticsearch::Client.new client_options
+      puts "COPT #{client_options}"
```
talevy: yup!

lib/logstash/outputs/elasticsearch.rb
suyograo: why the `puts` here? is it a place holder?

lib/logstash/outputs/elasticsearch.rb
```diff
@@ -113,14 +113,14 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
...
-  # Sets the host of the remote instance. If given an array it will load balance requests across the hosts specified in the `host` parameter.
+  # Sets the host(s) of the remote instance. If given an array it will load balance requests across the hosts specified in the `host` parameter.
   # Remember the `http` protocol uses the http://www.elastic.co/guide/en/elasticsearch/reference/current/modules-http.html#modules-http[http] address (eg. 9200, not 9300).
   #     `"127.0.0.1"`
   #     `["127.0.0.1:9200","127.0.0.2:9200"]`
   # It is important to exclude http://www.elastic.co/guide/en/elasticsearch/reference/current/modules-node.html[dedicated master nodes] from the `host` list
   # to prevent LS from sending bulk requests to the master nodes.  So this parameter should only reference either data or client nodes.
 
-  config :host, :validate => :array
+  config :hosts, :validate => :array
```
suyograo: IMO, we should still leave host as-is, but deprecate it. So new users will hopefully use `hosts` and existing users will migrate to the new setting

andrewvc: Isn't this the sort of thing that's perfect to change in 2.0?

suyograo: It is, but I worry that we didn't provide enough warning to deprecate :( I think we have a chance to deprecate it now, and foster the use of `hosts` moving forward. We can remove it completely in another version. What say?

andrewvc:
This now depends on this PR being accepted into elasticsearch ruby ( https://github.com/elastic/elasticsearch-ruby/pull/208 )

andrewvc:
Actually, thinking about the Logstash threading model we don't need a threadsafe client since we instantiate a new manticore client for each worker.

lib/logstash/outputs/elasticsearch.rb
```diff
@@ -354,17 +356,13 @@ def receive(event)
...
   def submit(actions)
     es_actions = actions.map { |a, doc, event| [a, doc, event.to_hash] }
-    @submit_mutex.lock
-    begin
-      bulk_response = @current_client.bulk(es_actions)
-    ensure
-      @submit_mutex.unlock
-    end
+
+    bulk_response = @client.bulk(es_actions)
```
talevy: thread safe I see?

talevy: this is only called per "worker". but `submit` is called by two threads. Stud Buffer and the retry thread

andrewvc: In that case it IS a good idea for us to not do a final release till those thread safety patches make it into the ruby client. Good catch @talevy

lib/logstash/outputs/elasticsearch.rb
```diff
@@ -130,7 +130,7 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
   # events before flushing that out to Elasticsearch. This setting
   # controls how many events will be buffered before sending a batch
   # of events.
-  config :flush_size, :validate => :number, :default => 5000
+  config :flush_size, :validate => :number, :default => 500
```
talevy: noice

suyograo: :8ball:

talevy:
LGTM

PR #238, opened 2015-08-28
Bring back the jar-dependencies 

purbon:
This PR brings jar-dependencies back, for now what this changes are doing is introducing a way to fetch jars at build time or during test execution, like this jars don't have to be living in the repo and things like upgrading versions gets simplified, plus no need to say reducing the size of the repo.

As jars are embedded also in the gem (at gem build time) this would mean offline plugin installation is still safe.

Looking forward to know your thoughts on it,

Relates to the changes/discussions proposed in https://github.com/elastic/logstash/issues/3735


purbon:
@talevy / @andrewvc  would like to know your thoughts on this, comments are super welcome!

andrewvc:
@purbon this is awesome! Can you re-submit this as a PR against https://github.com/andrewvc/logstash-output-elasticsearch_java/tree/native_only ? We're moving the jars out of this repo.

PR #252, opened 2015-09-18
Fix misleading documentation for HTTP(S) auth

mnylen:
Previously the site (https://www.elastic.co/guide/en/logstash/current/plugins-outputs-elasticsearch.html#plugins-outputs-elasticsearch-user) had the following documentation for user field:

    Username and password (only valid when protocol is HTTP; this setting works with HTTP or HTTPS auth)

And the password setting didn't have any documentation at all.

At least to me this was quite misleading, because I thought it means I can provide something like user => 'username:password'.

Fixed by separating documentation for user & password fields, and referencing them from each other.


mnylen:
I've signed the CLA at https://www.elastic.co/contributor-agreement/, why isn't it showing here? It worked with the main elasticsearch repository.

ph:
@mnylen CLA isn't updated automatically, I have double checked with our database and the CLA is fine. Can you rebase your changes with the master branch?

mnylen:
It seems the current master has already fixed this. Closing this one.

ph:
@mnylen the new version of the documentation should be online soon.

PR #255, opened 2015-09-23
Remove dependency on generator plugin for specs

andrewvc:
This un-breaks the current builds by removing the dependency on the `generator` input which does not work because there is not yet a released version that works with LS2.0


talevy:
LGTM

PR #276, opened 2015-10-21
Update template to use doc_values

### LINKED ISSUE #275: Improve default template
I suggest the following:
* Use "index": "not_analyzed" on all string fields (except message?)
* Use "doc_values": true on all string fields

untergeek:
Also, include mapping examples for other numeric types.

Types other than `long` and `double` will not be used unless manually
mapped by the user.

fixes #275

### DIFF
--- a/lib/logstash/outputs/elasticsearch/elasticsearch-template.json
+++ b/lib/logstash/outputs/elasticsearch/elasticsearch-template.json
@@ -5,37 +5,89 @@
   },
   "mappings" : {
     "_default_" : {
-       "_all" : {"enabled" : true, "omit_norms" : true},
-       "dynamic_templates" : [ {
-         "message_field" : {
-           "match" : "message",
-           "match_mapping_type" : "string",
-           "mapping" : {
-             "type" : "string", "index" : "analyzed", "omit_norms" : true
-           }
-         }
-       }, {
-         "string_fields" : {
-           "match" : "*",
-           "match_mapping_type" : "string",
-           "mapping" : {
-             "type" : "string", "index" : "analyzed", "omit_norms" : true,
-               "fields" : {
-                 "raw" : {"type": "string", "index" : "not_analyzed", "ignore_above" : 256}
-               }
-           }
-         }
-       } ],
-       "properties" : {
-         "@version": { "type": "string", "index": "not_analyzed" },
-         "geoip"  : {
-           "type" : "object",
-             "dynamic": true,
-             "properties" : {
-               "location" : { "type" : "geo_point" }
-             }
-         }
-       }
+      "_all" : {"enabled" : true, "omit_norms" : true},
+      "dynamic_templates" : [ {
+        "message_field" : {
+          "match" : "message",
+          "match_mapping_type" : "string",
+          "mapping" : {
+            "type" : "string", "index" : "analyzed", "omit_norms" : true
+          }
+        }
+      }, {
+        "string_fields" : {
+          "match" : "*",
+          "match_mapping_type" : "string",
+          "mapping" : {
+            "type" : "string", "index" : "analyzed", "omit_norms" : true,
+            "fields" : {
+              "raw" : {"type": "string", "index" : "not_analyzed", "doc_values" : true, "ignore_above" : 256}
+            }
+          }
+        }
+      }, {
+        "float_fields" : {
+          "match" : "*",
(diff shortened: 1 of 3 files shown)


lib/logstash/outputs/elasticsearch/elasticsearch-template.json
```diff
@@ -5,37 +5,89 @@
...
-         "geoip"  : {
-           "type" : "object",
-             "dynamic": true,
-             "properties" : {
-               "location" : { "type" : "geo_point" }
-             }
-         }
-       }
+      "_all" : {"enabled" : true, "omit_norms" : true},
+      "dynamic_templates" : [ {
```
andrewvc: Ah, I see this also uses a sane 2, rather than 3 space indentation!

andrewvc:
LGTM!

jordansissel:
Holding on this while we get some Elasticsearch expertise to evaluate this change with respect to older-but-still-supported Elasticsearch versions (1.4, 1.3, 1.2, etc)

andrewvc:
@jordansissel if this winds up being LS2.0 only is that an issue? We could announce as supporting 1.5.x+ (or whichever version supports doc values)

lib/logstash/outputs/elasticsearch/elasticsearch-template.json
```diff
@@ -5,37 +5,89 @@
...
+          "mapping" : { "type" : "byte", "doc_values" : true }
+        }
+      }, {
+        "short_fields" : {
+          "match" : "*",
+          "match_mapping_type" : "short",
+          "mapping" : { "type" : "short", "doc_values" : true }
+        }
+      }, {
+        "integer_fields" : {
```
pickypg: Integer fields smaller than `long` (`byte`, `short`, and `integer`) are good examples for templates, but the _type_ will never match because dynamic reflection of integer fields always parse as `long`s. This makes sense because it protects you from sending in `123` and having that get matched as a `byte`, only to then send in `123456` as the next value.

lib/logstash/outputs/elasticsearch/elasticsearch-template.json
```diff
@@ -5,37 +5,89 @@
...
+          "match_mapping_type" : "string",
+          "mapping" : {
+            "type" : "string", "index" : "analyzed", "omit_norms" : true,
+            "fields" : {
+              "raw" : {"type": "string", "index" : "not_analyzed", "doc_values" : true, "ignore_above" : 256}
+            }
+          }
+        }
+      }, {
+        "float_fields" : {
```
pickypg: The same applies to `float` relative to `double` as noted for integer fields below.

pickypg:
Technically this is missing boolean type, but since ES 1.x does not support doc values for it and it's the default in ES 2.x, I just wanted to note that it doesn't matter.

lib/logstash/outputs/elasticsearch/elasticsearch-template.json
untergeek: That's addressed in my first comment:

> Types other than `long` and `double` will not be used unless manually mapped by the user.

lib/logstash/outputs/elasticsearch/elasticsearch-template.json
untergeek: This, too, is addressed by the same comment:

> Types other than `long` and `double` will not be used unless manually mapped by the user.

They really are there just as good examples.

lib/logstash/outputs/elasticsearch/elasticsearch-template.json
pickypg: I must admit, I didn't even read the summary -- only the mapping. :100:

pickypg:
LGTM

GlenRSmith:
Can you elaborate on "unless manually mapped"? As far as I can tell, the only way to map another numeric type is to add the field to properties. Well, you can also pull some shenanigans by something like:

```
        {
          "integer_fields": {
            "match": "int-*",
            "match_mapping_type": "long",
            "mapping": {
              "type": "integer",
              "doc_values": true
            }
          }
        },
```

untergeek:
@GlenRSmith the "manual" part means that you need to map the field by field name or wild-card (like you suggested).  These are only in as "examples" that won't really do anything, since you can't put comments in JSON.

PR #277, opened 2015-10-22
Make host obsolete

suyograo:
`host` was removed in 2.0 and replaced by `hosts` setting. Making this `obsolete` so users can get feedback when they start LS with their current config


ph:
lgtm

PR #280, opened 2015-10-26
Require 2.0.0.rc1 min ls-core, fix specs to match

andrewvc:
Since the `port` option is now obsolete, this upgrades the specs to use the port as part of the hostname with `host:port`


andrewvc:
jenkins, test it

spec/es_spec_helper.rb
```diff
@@ -26,7 +27,7 @@ def get_port
   end
 
   def get_client
-    Elasticsearch::Client.new(:host => "#{get_host}:#{get_port}")
+    Elasticsearch::Client.new(:host => "#{get_host_port}:#{get_port}")
```
untergeek: You don't need the `:#{get_port}` part here, do you?  I see that `get_host_port` does that already.

spec/integration/outputs/retry_spec.rb
```diff
@@ -30,8 +30,7 @@ def mock_actions_with_response(*resp)
       "manage_template" => true,
       "index" => "logstash-2014.11.17",
       "template_overwrite" => true,
-      "hosts" => get_host(),
-      "port" => get_port(),
+      "hosts" => get_host_port(),
```
untergeek: why do some calls have `()` and some not?  Should this be made more consistent (and I can see that you're following the per-file patterns).

andrewvc: Good call! Interesting that the specs still pass...

andrewvc: Oops, responded to the wrong thing. Hmmm, that is just historical. Does the style guide say anything here? (I personally prefer ())

spec/es_spec_helper.rb
andrewvc: Ah thanks!

spec/integration/outputs/retry_spec.rb
untergeek: I'm not sure on the style guide.  @colinsurprenant @suyograo @jordansissel ?

untergeek:
LGTM™

talevy:
LGTM

PR #282, opened 2015-10-26
The default template to use doc_values wherever possible

### LINKED ISSUE #275: Improve default template

untergeek:
Includes mapping examples for other numeric types.

Types other than long and double will not be used unless manually
mapped by the user.

fixes #275

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,12 @@
+## 2.1.3
+ - Improved the default template to use doc_values wherever possible.
+ - Template contains example mappings for every numeric type. You must map your
+   own fields to make use of anything other than long and double.
+
+## 2.1.2
+ - Fixed dependencies (#280)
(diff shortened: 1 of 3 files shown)


jordansissel:
"resubmit ..." is a confusing commit message, can you reword it to describe the change being made? (for posterity)

jordansissel:
The changelog entry "the default template to use doc_values wherever possible" seems like a good candidate for the git commit subject line :)

untergeek:
Roger that. :+1:

andrewvc:
Tests pass. Aside from Jordan's comment, LGTM!

untergeek:
I'll call it good, then as the major review for this was in #276

Issue #130, opened 2015-05-06
[1.5.0 rc2 & 1.5.0 rc3 & 1.5.0 rc4] Basic auth is failing 


deuvarneys:
I have elasticsearch protected behind an nginx server with basic auth. In 1.4.2, I was able to push logs from logstash using the elasticsearch_http output plugin

output {
	elasticsearch_http{ 
		user => "user"
		password => "password"
		host => "host"
		port => "80"
	}
}

However, since 1.5 (rc2 & rc3), I keep getting "401 Authorization Required" from nginx when using both the method described above and 

output {
	elasticsearch{ 
		protocol => "http"
		user => "user"
		password => "password"
		host => "host"
		port => "80"
        }
}

jsvd:
@deuvarneys I've just tried this with

* elasticsearch 1.5.2
* logstash 1.5.2
* nginx 1.9.1

#### nginx conf

```
worker_processes  1;
daemon off;
events { worker_connections  1024; }

http {
    include       mime.types;
    default_type  application/octet-stream;
    sendfile        on;
    keepalive_timeout  65;
    server {
       listen 8080;
       location / {
          proxy_pass http://localhost:9200;
          auth_basic "Restricted";
          auth_basic_user_file htpasswd;
       }
    }
}
```

#### htpasswd file

`htpasswd -c conf/htpasswd test`

and entered `pwd` as password

##### logstash config

```
input { generator { count => 1 } }
output {
  elasticsearch {
    protocol => http
    user => test
    password => pwd
    host => localhost
    port => 8080
  }
}
```

#### elasticsearch

just unpacked a release and did `bin/elasticsearch`

_______

after starting elasticsearch, nginx and logstash I saw 1 event indexed in elasticsearch without issues.

deuvarneys:
Okay, I believe I found the issue. I was able to replicate the settings our production server and found the difference between it and the settings you provided. In our production settings, we are using the "@" special character within the password field. It appears that they special character is supported in Logstash 1.4 but not in Logstash 1.5 which is causing the unauthorized message from nginx. Will you no longer be supporting special characters within the password field for future releases?

jordansissel:
@deuvarneys nice work tracking that down! This is a bug.

untergeek:
Considering this fixed by #287

PR #287, opened 2015-10-27
Fix handling of escaped chars in user/password for basic auth

andrewvc:
This used to need to be escaped manually because we stuck it in the URL.
Now that it's just in the header we don't need to!

Fixes https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/130#issuecomment-111880658

### DIFF
--- a/spec/unit/outputs/elasticsearch_spec.rb
+++ b/spec/unit/outputs/elasticsearch_spec.rb
@@ -26,6 +26,21 @@ describe "outputs/elasticsearch" do
       thread.kill()
     end
 
+    describe "auth setup with url encodable passwords" do
+      let(:user) { "foo@bar"}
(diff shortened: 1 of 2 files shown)


spec/unit/outputs/elasticsearch_spec.rb
```diff
@@ -26,6 +26,21 @@
       thread.kill()
     end
 
+    describe "auth setup with url encodable passwords" do
+      let(:user) { "foo@bar"}
+      let(:password) {"baz@blah" }
+      let (:options) { super.merge("user" => user, "password" => password) }
```
jsvd: minor whitespace slip up

jsvd:
minor comment, otherwise LGTM, tests pass :+1:

PR #292, opened 2015-10-29
Refactor most code into separate modules

### LINKED ISSUE #291: Refactor this plugin to be more modular in order to support the java version
This plugin needs a major refactor, and will be undergoing one over the next week or so. Keep this in mind if you're submitting a PR! The underlying code will likely all move!

The code in this repository is currently not written in a very modular way. This has created a maintenance nightmare when it comes to sharing code with this plugin and https://github.com/logstash-plugins/logstash-output-elasticsearch_java .  Currently fixing a bug means significant back/cross-porting. This process is error-prone and time consuming.

To address this @talevy and I developed the following plan of attack:

1. Move as much code as possible out of elasticsearch.rb and into separate classes/modules
2. After 1. Elasticsearch.rb should be a thin wrapper calling out to those classes
3. Refactor the _java version of this plugin to have this repository as a dependency. Leverage this shared code as much as possible.

andrewvc:
This is a WIP tracking work to fix #291

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,3 +1,4 @@
 source 'https://rubygems.org'
 
 gemspec
+gem "stud", :path => "../ruby-stud"
--- a/Gemfile
+++ b/Gemfile
@@ -1,4 +1,3 @@
 source 'https://rubygems.org'
 
 gemspec
-gem "stud", :path => "../ruby-stud"
--- a/lib/logstash/outputs/elasticsearch.rb
+++ b/lib/logstash/outputs/elasticsearch.rb
@@ -49,113 +49,18 @@ require "uri" # for escaping user input
 # Keep in mind that a connection with keepalive enabled will
 # not reevaluate its DNS value while the keepalive is in effect.
 class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
-  attr_reader :client, :hosts
+  require "logstash/outputs/elasticsearch/http_client"
+  require "logstash/outputs/elasticsearch/http_client_builder"
+  require "logstash/outputs/elasticsearch/common_configs"
+  require "logstash/outputs/elasticsearch/common"
 
-  RETRYABLE_CODES = [409, 429, 503]
-  SUCCESS_CODES = [200, 201]
+  # Most config options are defined here!
+  include(LogStash::Outputs::ElasticSearch::CommonConfigs)
+
+  # Common impl. defined here
+  include(LogStash::Outputs::ElasticSearch::Common)
 
   config_name "elasticsearch"
-
-  # The index to write events to. This can be dynamic using the `%{foo}` syntax.
-  # The default value will partition your indices by day so you can more easily
-  # delete old data or only search specific date ranges.
-  # Indexes may not contain uppercase characters.
-  # For weekly indexes ISO 8601 format is recommended, eg. logstash-%{+xxxx.ww}
-  config :index, :validate => :string, :default => "logstash-%{+YYYY.MM.dd}"
-
-  # The index type to write events to. Generally you should try to write only
-  # similar events to the same 'type'. String expansion `%{foo}` works here.
-  #
-  # Deprecated in favor of `document_type` field.
-  config :index_type, :validate => :string, :obsolete => "Please use the 'document_type' setting instead. It has the same effect, but is more appropriately named."
-
-  # The document type to write events to. Generally you should try to write only
-  # similar events to the same 'type'. String expansion `%{foo}` works here.
-  # Unless you set 'document_type', the event 'type' will be used if it exists
-  # otherwise the document type will be assigned the value of 'logs'
-  config :document_type, :validate => :string
-
-  # Starting in Logstash 1.3 (unless you set option `manage_template` to false)
-  # a default mapping template for Elasticsearch will be applied, if you do not
-  # already have one set to match the index pattern defined (default of
-  # `logstash-%{+YYYY.MM.dd}`), minus any variables.  For example, in this case
-  # the template will be applied to all indices starting with `logstash-*`
-  #
-  # If you have dynamic templating (e.g. creating indices based on field names)
-  # then you should set `manage_template` to false and use the REST API to upload
-  # your templates manually.
-  config :manage_template, :validate => :boolean, :default => true
-
-  # This configuration option defines how the template is named inside Elasticsearch.
-  # Note that if you have used the template management features and subsequently
-  # change this, you will need to prune the old template manually, e.g.
-  #
-  # `curl -XDELETE <http://localhost:9200/_template/OldTemplateName?pretty>`
-  #
-  # where `OldTemplateName` is whatever the former setting was.
-  config :template_name, :validate => :string, :default => "logstash"
-
-  # You can set the path to your own template here, if you so desire.
-  # If not set, the included template will be used.
-  config :template, :validate => :path
-
-  # Overwrite the current template with whatever is configured
-  # in the `template` and `template_name` directives.
-  config :template_overwrite, :validate => :boolean, :default => false
-
-  # The document ID for the index. Useful for overwriting existing entries in
-  # Elasticsearch with the same ID.
-  config :document_id, :validate => :string
-
-  # A routing override to be applied to all processed events.
-  # This can be dynamic using the `%{foo}` syntax.
-  config :routing, :validate => :string
-
-  # Sets the host(s) of the remote instance. If given an array it will load balance requests across the hosts specified in the `hosts` parameter.
-  # Remember the `http` protocol uses the http://www.elastic.co/guide/en/elasticsearch/reference/current/modules-http.html#modules-http[http] address (eg. 9200, not 9300).
-  #     `"127.0.0.1"`
-  #     `["127.0.0.1:9200","127.0.0.2:9200"]`
-  # It is important to exclude http://www.elastic.co/guide/en/elasticsearch/reference/current/modules-node.html[dedicated master nodes] from the `hosts` list
-  # to prevent LS from sending bulk requests to the master nodes.  So this parameter should only reference either data or client nodes in Elasticsearch.
-
-  config :hosts, :validate => :array
-
-  config :host, :obsolete => "Please use the 'hosts' setting instead. You can specify multiple entries separated by comma in 'host:port' format."
-
-  # The port setting is obsolete.  Please use the 'hosts' setting instead.
-  # Hosts entries can be in "host:port" format.
-  config :port, :obsolete => "Please use the 'hosts' setting instead. Hosts entries can be in 'host:port' format."
-
-  # This plugin uses the bulk index API for improved indexing performance.
-  # To make efficient bulk API calls, we will buffer a certain number of
-  # events before flushing that out to Elasticsearch. This setting
-  # controls how many events will be buffered before sending a batch
-  # of events. Increasing the `flush_size` has an effect on Logstash's heap size.
-  # Remember to also increase the heap size using `LS_HEAP_SIZE` if you are sending big documents
-  # or have increased the `flush_size` to a higher value.
-  config :flush_size, :validate => :number, :default => 500
-
-  # The amount of time since last flush before a flush is forced.
-  #
-  # This setting helps ensure slow event rates don't get stuck in Logstash.
-  # For example, if your `flush_size` is 100, and you have received 10 events,
-  # and it has been more than `idle_flush_time` seconds since the last flush,
-  # Logstash will flush those 10 events automatically.
-  #
-  # This helps keep both fast and slow log streams moving along in
-  # near-real-time.
-  config :idle_flush_time, :validate => :number, :default => 1
-
-  # The Elasticsearch action to perform. Valid actions are:
-  #
-  # - index: indexes a document (an event from Logstash).
-  # - delete: deletes a document by id (An id is required for this action)
-  # - create: indexes a document, fails if a document by that id already exists in the index.
-  # - update: updates a document by id. Update has a special case where you can upsert -- update a
-  #   document if not already present. See the `upsert` option
-  #
-  # For more details on actions, check out the http://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html[Elasticsearch bulk API documentation]
-  config :action, :validate => %w(index delete create update), :default => "index"
 
   # Username to authenticate to a secure Elasticsearch cluster
   config :user, :validate => :string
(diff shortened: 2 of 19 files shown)


Gemfile
```diff
@@ -1,3 +1,4 @@
 source 'https://rubygems.org'
 
 gemspec
+gem "stud", :path => "../ruby-stud"
```
suyograo: todo: remove?

andrewvc: yes!

lib/logstash/outputs/elasticsearch.rb
```diff
@@ -232,6 +231,9 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
 
   public
   def register
+    require "logstash/outputs/elasticsearch/template_manager"
```
talevy: can we move these to the top of the file?

lib/logstash/outputs/elasticsearch.rb
```diff
@@ -203,14 +202,16 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
...
-  # Set retry policy for events that failed to send
-  config :retry_max_items, :validate => :number, :default => 5000
+  # Set max interval between bulk retries.
+  config :retry_max_interval, :validate => :number, :default => 2
 
-  # Set max interval between bulk retries
-  config :retry_max_interval, :validate => :number, :default => 5
+  # DEPRECATED This setting no longer does anything. If you need to change the number of retries in flight
+  # try increasing the total number of workers to better handle this.
+  config :retry_max_items, :validate => :number, :default => 5000, :deprecated => true
```
suyograo: Make this 500, even though its deprecated? We missed this in 2.0 and is out of sync with `flush_size`

lib/logstash/outputs/elasticsearch.rb
```diff
@@ -273,68 +270,86 @@ def register
...
+    # Initially we submit the full list of actions
+    submit_actions = actions
+
+    while submit_actions && submit_actions.length > 0 && retries_left > 0
+      return if !submit_actions # If everything's a success we move along
+      # We retry with whatever is didn't succeed
+      begin
+        submit_actions = submit(submit_actions)
+      rescue => e
+        @logger.warn("Encountered an unexpected error submitting a bulk request!",
```
suyograo: Do we still log the individual requests that failed and its corresponding ES errors?

suyograo: nm, i see it below

lib/logstash/outputs/elasticsearch.rb
andrewvc: Can do :)

lib/logstash/outputs/elasticsearch.rb
```diff
@@ -273,68 +270,86 @@ def register
...
-  def get_template
-    if @template.nil?
-      @template = ::File.expand_path('elasticsearch/elasticsearch-template.json', ::File.dirname(__FILE__))
-      if !File.exists?(@template)
-        raise "You must specify 'template => ...' in your elasticsearch output (I looked for '#{@template}')"
+  def submit(actions)
+    es_actions = actions.map { |a, doc, event| [a, doc, event.to_hash] }
+
+    bulk_response = safe_bulk(es_actions,actions)
+    return if @stopping.true? # bulk_response is nil when @stopping == true
```
suyograo: Shouldn't this be done on the top level loop? Much cleaner IMO

lib/logstash/outputs/elasticsearch.rb
```diff
@@ -273,68 +270,86 @@ def register
...
+      # We retry with whatever is didn't succeed
+      begin
+        submit_actions = submit(submit_actions)
+      rescue => e
+        @logger.warn("Encountered an unexpected error submitting a bulk request!",
+        :message => e.message,
+        :class => e.class.name,
+        :backtrace => e.backtrace)
       end
+      retries_left -= 1
```
suyograo: Wait for `retry_max_interval` before looping again?

lib/logstash/outputs/elasticsearch.rb
andrewvc: Yes. I'd ask to actually refrain from commenting on the PR just this minute. I'm in the midst of cleanup at the minute, so it'll change a lot soon :)

lib/logstash/outputs/elasticsearch.rb
suyograo: Can we also log how many total events we are retrying?

lib/logstash/outputs/elasticsearch.rb
```diff
@@ -273,68 +270,86 @@ def register
...
 
-    if @manage_template
-      begin
-        @logger.info("Automatic template management enabled", :manage_template => @manage_template.to_s)
-        @client.template_install(@template_name, get_template, @template_overwrite)
-      rescue => e
-        @logger.error("Failed to install template: #{e.message}")
-      end
-    end
+    TemplateManager.install_template(self)
```
suyograo: nice!

andrewvc:
jenkins, test it

andrewvc:
@suyograo I hope the doc generator can work with these changes!

andrewvc:
@suyograo @talevy I would recommend just looking at the new source, I doubt the changelog will be possible to follow.

lib/logstash/outputs/elasticsearch/buffer.rb
```diff
@@ -0,0 +1,130 @@
...
+      end
+    end
+
+    def flush_unsafe
+      if @buffer.size > 0
+        @submit_proc.call(@buffer)
+        @buffer.clear
+      end
+
+      @last_flush = Time.now # This must always be set to ensure correct timer behavior
```
talevy: should we set this to be UTC always, so that local changes to time do not affect this. Since, we only care about diffs.

andrewvc: Time.now has knowledge of UTC offsets, so it should be accurate. For instance:

```
irb(main):001:0> Time.now
=> 2015-11-04 16:44:33 -0600
```

lib/logstash/outputs/elasticsearch/buffer.rb
```diff
@@ -0,0 +1,130 @@
...
+    private
+
+    def push_unsafe(item)
+      @buffer << item
+      if @buffer.size >= @max_size
+        flush_unsafe
+      end
+    end
+
+    def spawn_interval_flusher
```
talevy: does it make sense to use something like Buffer.Interval here? or are we trying to get rid of even that as a dependency

lib/logstash/outputs/elasticsearch/common.rb
```diff
@@ -0,0 +1,171 @@
...
+        "Attempted to send a bulk request to Elasticsearch configured at '#{@client.client_options[:hosts]}',"+
+          " but Elasticsearch appears to be unreachable or down!",
+        :client_config => @client.client_options,
+        :error_message => e.message,
+        :class => e.class.name
+      )
+      @logger.debug("Failed actions for last bad bulk request!", :actions => actions)
+
+      # We retry until there are no errors! Errors should all go to the retry queue
+      sleep 1
```
talevy: let me know if this crosses the line in the form of a scope-creep.... adding exponential backoff to the retry intervals.

lib/logstash/outputs/elasticsearch/buffer.rb
andrewvc: I didn't see the benefit here in terms of complexity, and I thought reducing the deps would be nice.

lib/logstash/outputs/elasticsearch/http_client.rb
```diff
@@ -150,4 +151,4 @@ def template_put(name, template)
       @client.indices.put_template(:name => name, :body => template)
     end
   end
-end
+end end end
```
talevy: end!

lib/logstash/outputs/elasticsearch/common.rb
andrewvc: I like that idea, but I think that can go in a future PR. Maybe open a ticket?

lib/logstash/outputs/elasticsearch/http_client.rb
andrewvc: So many ends! I hope @guyboertje is proud

spec/integration/outputs/retry_spec.rb
```diff
@@ -22,7 +22,7 @@ def mock_actions_with_response(*resp)
       }
     end
 
-    allow_any_instance_of(LogStash::Outputs::Elasticsearch::HttpClient).to receive(:bulk).and_return(*expanded_responses)
+    allow_any_instance_of(LogStash::Outputs::ElasticSearch::HttpClient).to receive(:bulk).and_return(*expanded_responses)
```
talevy: haha, retribution

spec/unit/outputs/elasticsearch_spec.rb
```diff
@@ -118,12 +114,13 @@
...
       # Expect a timeout to be logged.
-      expect(eso.logger).to receive(:warn).with("Failed to flush outgoing items",
-                                                hash_including(:exception => "Manticore::SocketTimeout"))
+      expect(eso.logger).to receive(:error).with(/Attempted to send a bulk request/, anything)
     end
 
     after do
       listener[0].close
+      # Stop the receive buffer, but don't flush because that would hang forever in this case since ES never returns a result
+      eso.instance_variable_get(:@buffer).stop(false,false)
```
talevy: space after comma?

lib/logstash/outputs/elasticsearch.rb
```diff
@@ -50,114 +49,18 @@
 # Keep in mind that a connection with keepalive enabled will
 # not reevaluate its DNS value while the keepalive is in effect.
 class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
-  attr_reader :client
+  require "logstash/outputs/elasticsearch/http_client"
```
suyograo: Can we not do lazy load please..

spec/integration/outputs/retry_spec.rb
andrewvc: YES!

lib/logstash/outputs/elasticsearch.rb
andrewvc: I too am against lazy loads.

This isn't actually a lazy load, its more of an ordered load. This will be loaded when the file is loaded. The reason its here, rather than top of file, is that doing this at the top of the file breaks unless all modules look like `module LogStash; module Outputs; module ElasticSearch < LogStash::Outputs::Base`.

The subclassing belongs here, not there, so this was the best workaround I was able to find.

andrewvc: For clarity, since this is in the class def it isn't really lazy. You needn't call a method to do the load.

lib/logstash/outputs/elasticsearch/common.rb
suyograo: Shouldn't we be sleeping for `retry_max_interval`?

lib/logstash/outputs/elasticsearch.rb
suyograo: cool, thanks for explaining :) TIL

jordansissel: I solve this annoying problem by having a 'namespace.rb' that all things require that describes the namespace of modules that will be used by anything. It's not a great solution, so the solution you have here is just fine :)

lib/logstash/outputs/elasticsearch.rb
```diff
@@ -50,114 +49,18 @@
...
+  require "logstash/outputs/elasticsearch/http_client_builder"
+  require "logstash/outputs/elasticsearch/common_configs"
+  require "logstash/outputs/elasticsearch/common"
 
-  include Stud::Buffer
-  RETRYABLE_CODES = [409, 429, 503]
-  SUCCESS_CODES = [200, 201]
+  # Most config options are defined here!
+  include(LogStash::Outputs::ElasticSearch::CommonConfigs)
```
jordansissel: Tell me about what differentiates config settings that go into CommonConfigs vs ones that are declared in this file?

lib/logstash/outputs/elasticsearch.rb
```diff
@@ -222,330 +127,12 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
...
-    }
-  end
-
-  private
-  def retry_push(actions)
-    Array(actions).each{|action| @retry_queue << action}
-    @retry_flush_mutex.synchronize {
-      @retry_queue_needs_flushing.signal if @retry_queue.size >= @retry_max_items
-    }
+  def build_client
```
jordansissel: Feels weird that the build_client method specific to HTTP remains in the top level here. Thoughts?

andrewvc: Yes, this is http specific. The Java plugin won't want this mixed in when
it uses common.rb

lib/logstash/outputs/elasticsearch.rb
andrewvc: Stuff in common will be used by the Java plugin, the stuff in elastic
search.rb is for http

lib/logstash/outputs/elasticsearch/buffer.rb
```diff
@@ -0,0 +1,130 @@
...
+      Thread.new do
+        loop do
+          sleep 0.2
```
jordansissel: Having the loop block here moved to a separate method would aid readability, I think; at this point in the code we're pretty deep indentation and it only goes deeper a few lines from here ;P

Something like:

```
Thread.new do
  loop { myfunc }
end
```

jordansissel: AlternateLY:

Thread.new do
  loop do
    sleep 0.2
    break if stopping?
    synchronize { myfunc }
  end
end
```

File: lib/logstash/outputs/elasticsearch/buffer.rb [hunk 18]
```diff
@@ -0,0 +1,130 @@
+require 'concurrent'
+java_import java.util.concurrent.locks.ReentrantLock
+
+module LogStash; module Outputs; class ElasticSearch
+  class Buffer
+    def initialize(logger, max_size, flush_interval, &block)
```
jordansissel: Conceptually, this class replaces Stud::Buffer?

andrewvc: Yes, the idea was to get the retry logic out of the buffer impl and
encapsulate the whole thing in a class rather than a mixin. My goal was to
move us incrementally forward for the next gen pipeline, which will be more
similar to this class.

jordansissel: +1 thanks for the background! :)

lib/logstash/outputs/elasticsearch/buffer.rb
andrewvc: I'll adopt your second approach. Good call @jordansissel :)

andrewvc:
Thanks for the reviews @jordansissel @suyograo @talevy  . I believe I've addressed all the raised concerns.

How's it looking insofar as getting to LGTM?

suyograo:
I am good with these changes, would wait for a final LGTM from @talevy since he knows this code best

talevy:
LGTM

PR #307, opened 2015-11-16
Add support for parent relationship

### LINKED ISSUE #129: Elasticsearch output does not support parent-child indexing
Hi!

I might be wrong, but seems like current implementation of elasticsearch output does not support [parent-child indexing](http://www.elastic.co/guide/en/elasticsearch/guide/master/parent-child.html). This way relevant documents cannot be posted into elasticsearch via logstash.

It looks like required change in plugin is very small and can be achieved by additional parameter + slight code modification (according to [relevant documentation](http://www.elastic.co/guide/en/elasticsearch/guide/master/indexing-parent-child.html)). Not a ruby expert, but the change may look like following:

elasticsearch-output

This is definitely enhancement, but kind of an easy win :)

Thank you.

jimmyjones2:
Fixes #129

### DIFF
--- a/lib/logstash/outputs/elasticsearch/common.rb
+++ b/lib/logstash/outputs/elasticsearch/common.rb
@@ -112,10 +112,10 @@ module LogStash; module Outputs; class ElasticSearch;
         :_id => @document_id ? event.sprintf(@document_id) : nil,
         :_index => event.sprintf(@index),
         :_type => type,
-        :_routing => @routing ? event.sprintf(@routing) : nil,
-        :parent => @parent ? event.sprintf(@parent) : nil
+        :_routing => @routing ? event.sprintf(@routing) : nil
       }
 
+      params[:parent] = event.sprintf(@parent) if @parent
       params[:_upsert] = LogStash::Json.load(event.sprintf(@upsert)) if @action == 'update' && @upsert != ""
       params
     end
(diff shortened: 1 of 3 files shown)


jimmyjones2:
@untergeek Would you mind reviewing please?

andrewvc:
@jimmyjones2 sorry for the backlog! @untergeek , myself, or someone else will try to look at this PR as soon as possible. Thank you so much for the contribution, this is valuable new functionality!

lib/logstash/outputs/elasticsearch/common.rb
```diff
@@ -112,7 +112,8 @@ def event_action_params(event)
         :_id => @document_id ? event.sprintf(@document_id) : nil,
         :_index => event.sprintf(@index),
         :_type => type,
-        :_routing => @routing ? event.sprintf(@routing) : nil
+        :_routing => @routing ? event.sprintf(@routing) : nil,
+        :parent => @parent ? event.sprintf(@parent) : nil
```
andrewvc: This is breaking some old tests. Rather than setting this to nil we should simply not set the key if it isn't present.

We should probably go back and do the same for routing as well, but that's beyond the scope of this PR.

spec/integration/outputs/parent_spec.rb
```diff
@@ -0,0 +1,73 @@
+require_relative "../../../spec/es_spec_helper"
+
+shared_examples "a parent indexer" do
```
andrewvc: Thanks for writing so many tests!

jimmyjones2:
hey @andrewvc how does this look?

andrewvc:
LGTM! Will merge this. Thanks for the patch @jimmyjones2

PR #309, opened 2015-11-18
Disable fielddata use for analyzed strings

### LINKED ISSUE #308: Disable fielddata for all analyzed string fields in LS template
Have come across this multiple times in the past with various users in the field.  ES does not support doc values for analyzed string fields today (even in ES 2.0).  Kibana warns about, but does not adequately prevent users from hurting themselves by running aggs/sorting/scripting against an analyzed string field (https://github.com/elastic/kibana/issues/5189).  

We recently added doc_values: true to the latest LS template in master which is a great addition (esp. for users on 1.x).  I will suggest taking this further by disabling fielddata for all analyzed string fields by default (adding the following to all analyzed string fields):

```
"fielddata": {
                "format": "disabled"
              }
```

For those who need fielddata for analyzed string and don't care about issues that stem from loading analyzed string field values into the heap, they can change the template themselves.

From what I have seen, 90%+ of the use cases out there do not require agg/sorting/scripting against analyzed string fields and these are operations are done against not_analyzed string fields.

pickypg:
By default, users should not be sorting, aggregating, or scripting against the analyzed string variant of fields. Instead, users should do those things against the `.raw` multifield variant.

This _will_ prevent the use of `significant_terms` aggregation against the associated field, which will be undesirable in rare cases. Given the rarity of that need, I would much rather have those users manually enable fielddata explicitly for the field (it is dynamically updatable!).

Closes #308

### DIFF
--- a/lib/logstash/outputs/elasticsearch/elasticsearch-template.json
+++ b/lib/logstash/outputs/elasticsearch/elasticsearch-template.json
@@ -11,7 +11,8 @@
           "match" : "message",
           "match_mapping_type" : "string",
           "mapping" : {
-            "type" : "string", "index" : "analyzed", "omit_norms" : true
+            "type" : "string", "index" : "analyzed", "omit_norms" : true,
+            "fielddata" : { "format" : "disabled" }
           }
         }
       }, {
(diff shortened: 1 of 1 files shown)


untergeek:
LGTM™

@pickypg thanks for adding this!  This is a great addition, which I presume comes from your experience in supporting Elasticsearch. :+1:

pickypg:
> which I presume comes from your experience in supporting Elasticsearch.

Indeed! :)

PR #313, opened 2015-11-23
Sprintf arg checking

### LINKED ISSUE #312: Enum-style validation of "action" option blocks variable interpolation
As evidenced by https://discuss.elastic.co/t/couchdb-plugin-set-dynamic-type-for-elasticsearch/2293/18,

```
Invalid setting for elasticsearch output plugin:

output {
elasticsearch {
# This setting must be a ["index", "delete", "create", "update"]
# Expected one of ["index", "delete", "create", "update"], got ["%{[@metadata][action]}"]
action => "%{[@metadata][action]}"
...
}
```

the static validation of the `action` option doesn't work with the dynamic variable interpolation that eventually will take place, but that code won't be reached because of the validation. I seems either the validation should be relaxed or calls to event.sprintf() should be dropped for `@action` to not give the impression that you actually can include %{name} variable references there.

andrewvc:
This fixes #312 . I had to fix a broken integration spec as well to get tests to pass. To that end I also threw in a .travis.yml file.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,9 @@
+before_install:
+  - curl -s https://download.elasticsearch.org/elasticsearch/release/org/elasticsearch/distribution/deb/elasticsearch/2.0.0/elasticsearch-2.0.0.deb > elasticsearch.deb
+  - sudo dpkg -i --force-confnew elasticsearch.deb
+  - sudo service elasticsearch start
+  - sleep 10
+language: ruby
+rvm:
+  - jruby-19mode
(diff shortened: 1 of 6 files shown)


jordansissel:
Code change LGTM. I like the way you fixed this by still having some kind of validation if the string is not formatted. <3

jordansissel:
Travis is happy. Woot. LGTM

andrewvc:
Thanks @jordansissel ! I also just pushed a new commit to support travis docker containers + bundle gem caching for an even faster test cycle!

PR #315, opened 2015-11-24
Fix sniffing spec to sleep for the proper amount of time.

andrewvc:
It sniffs once per second, so sleeping for 2s put it right on the boundary.
This should make our travis tests pass consistently


untergeek:
LGTM™

andrewvc:
@untergeek ty!

PR #322, opened 2015-12-04
serialize access to the connection pool in es-ruby client

jsvd:
a sniffing action triggers the reload of connections which can
close a socket being used by a running request
this patch introduces a mutex so that bulk operations, template
installation and sniffing don't execute at the same time

### DIFF
--- a/lib/logstash/outputs/elasticsearch/http_client.rb
+++ b/lib/logstash/outputs/elasticsearch/http_client.rb
@@ -18,18 +18,27 @@ module LogStash; module Outputs; class ElasticSearch;
       # @options = DEFAULT_OPTIONS.merge(options)
       @options = options
       @client = build_client(@options)
+      # mutex to prevent requests and sniffing to access the
+      # connection pool at the same time
+      @request_mutex = Mutex.new
       start_sniffing!
     end
 
     def template_install(name, template, force=false)
-      if template_exists?(name) && !force
-        @logger.debug("Found existing Elasticsearch template. Skipping template management", :name => name)
-        return
+      @request_mutex.synchronize do
+        if template_exists?(name) && !force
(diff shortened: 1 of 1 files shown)


jsvd:
@andrewvc can you review? :)

lib/logstash/outputs/elasticsearch/http_client.rb
```diff
@@ -18,18 +18,27 @@ def initialize(options={})
...
+        template_put(name, template)
       end
-      template_put(name, template)
     end
 
     def bulk(actions)
+      @sniff_request_mutex.synchronize { synchronized_bulk(actions) }
+    end
+
+    def synchronized_bulk(actions)
```
andrewvc: Errr, this is a weird method name because it is in fact not synchronized. Maybe `unsafe_bulk` or `unsynchronized_bulk` would be more clear?

andrewvc:
Aside from my comment on the method naming LGTM. I do feel strongly that that should be changed however.

lib/logstash/outputs/elasticsearch/http_client.rb
jsvd: makes perfect sense, will change :+1:

jsvd:
@andrewvc done

lib/logstash/outputs/elasticsearch/http_client.rb
```diff
@@ -18,18 +18,27 @@ def initialize(options={})
       # @options = DEFAULT_OPTIONS.merge(options)
       @options = options
       @client = build_client(@options)
+      # mutex to prevent requests and sniffing to access the
+      # connection pool at the same time
+      @sniff_request_mutex = Mutex.new
```
suyograo: @jsvd minor nitpick, but why not just call it `request_mutex`? I was actually confused when I read the code below as to why you were using a sniff_mutex to lock template update request

PR #327, opened 2015-12-11
Sane retry logic!

### LINKED ISSUE #321: Retry logic does the wrong thing
Currently our retry logic is the following:
1. Retry [409, 429, 503] up `max_attempts` times per batch.
2. Retry network errors infinitely.

As @gingerwizard pointed out this all makes no sense. We would expect 429/503 to retry infinitely. These both mean that Elasticsearch is just too busy to do anything useful.

For the 409 use case there's no reason to retry on the ES output. We should let Elasticsearch handle that with `retry_on_conflict`.

The great thing about this is we'll need _less_ code in this plugin for better performance / functionality!

andrewvc:
Fixes #321

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,4 +8,4 @@ language: ruby
 cache: bundler
 rvm:
   - jruby-19mode
-script: bundle exec rspec spec && bundle exec rspec spec
+script: bundle exec rspec spec && bundle exec rspec spec --tag integration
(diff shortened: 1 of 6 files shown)


suyograo:
@andrewvc can you update the docs too?

andrewvc:
@suyograo aye captain, will do monday!

untergeek:
Sane... I like it!  LGTM™ after the docs have been improved.

andrewvc:
@untergeek @suyograo documentation has been updated. How's it look now?

suyograo:
LGTM

Issue #326, opened 2015-12-11
elasticsearch output plugin does not respect max_retry_interval during bulk send when elasticsearch responds with HTTP 503


CoaxVex:
The [documentation on the elasticsearch output plugin](https://www.elastic.co/guide/en/logstash/current/plugins-outputs-elasticsearch.html#_retry_policy) describes that when an elasticsearch node responds with  a HTTP 429, 409 or 503 code, logstash should retry the send according to the :max_retries, :retry_max_items and :retry_max_interval settings.

The code in [logstash-output-elasticsearch/lib/logstash/outputs/elasticsearch/common.rb](https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/aa898fcaedd15a2102d132695d815abb1df125a1/lib/logstash/outputs/elasticsearch/common.rb#L159), however, only triggers the retry logic upon a socket exception or a socket timeout: 
Line 159:

    rescue Manticore::SocketException, Manticore::SocketTimeout => e

This means, that other errors will not trigger the retry function, and will fail immediately, dropping the message that is to be stored in elasticsearch.

The expected behavior would be that logstash retries the bulk transfer with :retry_max_interval seconds in between, until :max_retries is reached when elasticsearch responds with a HTTP error code that indicates that a write is currently impossible.

guyboertje:
There are two types of retry in play here.  One where the ES client does get a response and its code is one of 429 or 503.  The other when the client cannot connect to ES as seen by the rescued Manticore exception.  Other exceptions are logged but not retried.

@CoaxVex - what error is being logged?

CoaxVex:
Attempted to send a bulk request to Elasticsearch configured at '[\"http://192.168.50.7:9200/\"]', but an error occurred and it failed!

Log file extract attached:
[logstash.txt](https://github.com/logstash-plugins/logstash-output-elasticsearch/files/75478/logstash.txt)

Output configuration:

    output {
        stdout { codec => rubydebug }

        elasticsearch {
            hosts => ["192.168.50.7"]
            max_retries => 10
            retry_max_interval => 5
            retry_max_items => 5000
        }
    }


It looks like it actually **does** retry, but just doesn't respect the retry_max_interval parameter.

Screenshot of tcpdump/wireshark capture:
![elasticsearch](https://cloud.githubusercontent.com/assets/10154961/12063447/ffabad70-afae-11e5-9467-361a79d996f0.png)

[...]

PR #337, opened 2016-01-05
Fixes a bug whereby http errors would not sleep for max_retry_interval

andrewvc:
This fixes https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/326
by correctly sleeping between retries for HTTP error codes, not just connection errors.

It also fixes a defensive array length check

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,5 +7,5 @@ before_install:
 language: ruby
 cache: bundler
 rvm:
-  - jruby-19mode
+  - jruby-1.7.23
 script: bundle exec rspec spec && bundle exec rspec spec --tag integration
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -7,5 +7,5 @@ before_install:
 language: ruby
 cache: bundler
 rvm:
-  - jruby-19mode
+  - jruby-1.7.22
```
untergeek: shouldn't this be 1.7.23?

untergeek:
Other than the niggling 1.7.22 jruby version, LGTM™

andrewvc:
Thanks @untergeek, just bumped the version, will merge after the tests pass.

PR #345, opened 2016-01-20
Clean logstash elasticsearch template

rhoml:
Since doc_values are already defined it is not required to defined again at the field type declaration.


untergeek:
While it is true that it is not necessary to do so, it hurts nothing to be explicit and redundant here.  I am of a mind to keep it this way, if only to be visible as a concise example of a statically mapped field.

rhoml:
I feel it is quite misleading though

andrewvc:
Thank you for the feedback @rhomi, but I'm inclined to agree with @untergeek here. I'm going to close this issue since there isn't any more progress to be made here. If you have any future concerns don't hesitate to open an issue!

PR #350, opened 2016-01-28
Use debug instead of info level for log message

### LINKED ISSUE #336: Move "flushing buffer at interval messages" to debug level?
It looks like we are now writing out the full contents of the buffer at the info/verbose level which can drown other more useful info messages.

```
Flushing buffer at interval {:instance=>"#<LogStash::Outputs::ElasticSearch::Buffer:
```

https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/06ea707f1891201020e67ebe680838be2036adee/lib/logstash/outputs/elasticsearch/buffer.rb#L90

Have we considered writing this only to the debug level?

suyograo:
Writing out the full contents of the buffer at the info/verbose level
is a lot of noise. Downgrade to debug

Fixes #336


talevy:
LGTM

PR #351, opened 2016-01-29
Update docs for template management

### LINKED ISSUE #347: Clarify template_overwrite option
For template_overwrite, the [doc](https://www.elastic.co/guide/en/logstash/current/plugins-outputs-elasticsearch.html#plugins-outputs-elasticsearch-template_overwrite) says:

>Overwrite the current template with whatever is configured in the template and template_name directives.

It does not match up with the description of the parameter [here](https://www.elastic.co/blog/new-in-logstash-1-3-elasticsearch-index-template-management) - which suggests that if you don't have your own template that it is actually the parameter to use if you want to overwrite the template with the one that is provided by the plugin that is installed (always keep up to date with the currently installed plugin) - to keep up to date as you upgrade LS.

The part about using this parameter to keep up to date with the out of the box LS template is something that we will nice to add to the guide.  And maybe also add the disclaimer that if they are using a customized version of the Logstash template (logstash), this is not a good option to use for it will tell Logstash to overwrite the "logstash" template (i.e. remove all customized settings).  So if custom settings are used, they want to make sure to use template+template_name, etc..

suyograo:
Fixes #347

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 2.4.1
+ - Used debug level instead of info when emitting flush log message
+ - Updated docs about template
+
 ## 2.4.0
  - Scripted update support courtesy of @Da-Wei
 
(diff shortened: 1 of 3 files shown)


ppf2:
Forgot to put here earlier, LGTM! thx!

PR #353, opened 2016-02-02
Fix the flush_size option to work with NG Pipeline

andrewvc:
@suyograo would you be able to have a look at this at some point? Not sure if this is in time to make it with 2.2, if not, no biggie. The 500 flush size default means that unless you set `-b` above 500 this will have no effect anyway.

### DIFF
--- a/lib/logstash/outputs/elasticsearch/common.rb
+++ b/lib/logstash/outputs/elasticsearch/common.rb
@@ -25,7 +25,9 @@ module LogStash; module Outputs; class ElasticSearch;
 
     # Receive an array of events and immediately attempt to index them (no buffering)
     def multi_receive(events)
-      retrying_submit(events.map {|e| event_action_tuple(e) })
+      events.each_slice(@flush_size) do |slice|
+        retrying_submit(slice.map {|e| event_action_tuple(e) })
+      end
     end
 
     # Convert the event into a 3-tuple of action, params, and event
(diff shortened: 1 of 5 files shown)


lib/logstash/outputs/elasticsearch/common.rb
```diff
@@ -25,7 +25,9 @@ def receive(event)
 
     # Receive an array of events and immediately attempt to index them (no buffering)
     def multi_receive(events)
-      retrying_submit(events.map {|e| event_action_tuple(e) })
+      events.each_slice(@flush_size) do |slice|
```
suyograo: Shouldn't there be a test if `batch_size > @flush_size` and then slice it? Or does `each_slice` takes care of that internally using the `size` of `events` ?

andrewvc: `each_slice` does the right thing!

```
irb(main):002:0> [1,2,3,4].each_slice(5).to_a
=> [[1, 2, 3, 4]]
```

suyograo: nice!

suyograo:
@andrewvc LGTM, but this is late for 2.2 packaging, i've cut the release already. We can go ahead and publish this plugin though

lib/logstash/outputs/elasticsearch/common.rb
suyograo: Can you add a comment here, saying why we are doing this in relationship to `batch_size`?

Issue #356, opened 2016-02-03
Update action wrong.


kiryam:
In
lib/logstash/outputs/elasticsearch/http_client.rb

Line
`source = { 'doc' => source }`

It is wrong.
In documenation
https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html

Example:
```
{ "index" : { "_index" : "test", "_type" : "type1", "_id" : "1" } }
{ "field1" : "value1" }
{ "delete" : { "_index" : "test", "_type" : "type1", "_id" : "2" } }
{ "create" : { "_index" : "test", "_type" : "type1", "_id" : "3" } }
{ "field1" : "value3" }
{ "update" : {"_id" : "1", "_type" : "type1", "_index" : "test"} }
{ "doc" : {"field2" : "value2"}}

```

If I push code bellow, I will have in ES like this:
```
curl -XPOST http://localhost:9200/test/_search?pretty
{
  "took" : 1,
  "timed_out" : false,
  "_shards" : {
    "total" : 5,
    "successful" : 5,
    "failed" : 0
  },
  "hits" : {
    "total" : 2,
    "max_score" : 1.0,
    "hits" : [ {
      "_index" : "test",
      "_type" : "type1",
      "_id" : "1",
      "_score" : 1.0,
      "_source" : {
        "field1" : "value1",
        "@version" : "1",
        "@timestamp" : "2016-02-03T11:46:12.337Z",
        "doc" : {
          "field2" : "value2"
        }
      }
    }, {
      "_index" : "test",
      "_type" : "type1",
      "_id" : "3",
      "_score" : 1.0,
      "_source" : {
        "field1" : "value3",
        "@version" : "1",
        "@timestamp" : "2016-02-03T11:46:12.337Z"
      }
    } ]
  }
}
```

kiryam:
#357

Da-Wei:
Hi @kiryam 

When I use your example : 
```bash
curl http://localhost:9200/_bulk -d '
{ "index" : { "_index" : "test", "_type" : "type1", "_id" : "1" } }
{ "field1" : "value1" }
{ "delete" : { "_index" : "test", "_type" : "type1", "_id" : "2" } }
{ "create" : { "_index" : "test", "_type" : "type1", "_id" : "3" } }
{ "field1" : "value3" }
{ "update" : {"_id" : "1", "_type" : "type1", "_index" : "test"} }
[...]

PR #357, opened 2016-02-03
source = { 'doc' => source }

kiryam:
It is hack.
https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html

From docs:
{ "update" : {"_id" : "1", "_type" : "type1", "_index" : "index1"} }
{ "doc" : {"field2" : "value2"} }

If use explicit add "doc" it will be like this:
{ "update" : {"_id" : "1", "_type" : "type1", "_index" : "index1"} }
{{"doc":  "doc" : {"field2" : "value2"} }}

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 2.4.3
+ - Fix bug with update document. (#356)
+
 ## 2.4.2
  - Make flush_size actually cap the batch size in LS 2.2+
 
--- a/lib/logstash/outputs/elasticsearch.rb
+++ b/lib/logstash/outputs/elasticsearch.rb
@@ -16,7 +16,7 @@ require "uri" # for escaping user input
 # We strongly encourage the use of HTTP over the node protocol for a number of reasons. HTTP is only marginally slower,
 # yet far easier to administer and work with. When using the HTTP protocol one may upgrade Elasticsearch versions without having
 # to upgrade Logstash in lock-step. For those still wishing to use the node or transport protocols please see
-# the https://www.elastic.co/guide/en/logstash/2.0/plugins-outputs-elasticsearch_java.html[logstash-output-elasticsearch_java] plugin.
+# the <<plugins-outputs-elasticsearch_java,elasticsearch_java output plugin>>.
 #
 # You can learn more about Elasticsearch at <https://www.elastic.co/products/elasticsearch>
 #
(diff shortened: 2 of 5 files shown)


kiryam:
#357

kiryam:
#356

untergeek:
Code LGTM™ and tests pass.  This bug has been there for a while, and I thank you for reporting it and patching it.

Could you please bump the version in the gemspec (you may need to rebase first, as master is currently at 2.4.2) and add a line in the Changelog about this change?  We will publish the updated version as soon as we merge this.

kiryam:
@untergeek I do all correctly? Can it be merged?

untergeek:
Let me see if I can get one more LGTM from someone else on the team, then yes.

jordansissel:
I'm not really sure what this fixes. From what I've read this seems to be an issue with how you are inputting data into RabbitMQ (input), not how we output it? I think I'm missing some context on what this fixes.

I've read #356, but I"m still not sure what's going on.

The way I read the code change is that if the event contains a `doc` field at the top level, we will use that whole event as the thing to pass to the 2nd line of the bulk request (where the 1st line is the `{ "update" : { ... }`).

I believe this implies an assumption that the event is some kind of partial-bulk entity (it contains the `{ "doc": ... }` clause, but is missing the `{ "update": ... }` clause.

This means if anyone has a `"doc"` field in the top level, we now enforce that it _is_ the `doc` clause for the update, which may not be correct, and this causes at least two problems:

1) Users with a field called `doc` who are updating with today's behavior will see corruption (incorrect updates) or errors with no configuration change if this patch is merged as-is.
2) Users with a top-level `doc` field in an event could be a valid Event but not valid for use with the elasticsearch bulk `update`.


Let's discuss more in #356 about your use case. I believe this proposed change will break existing behavior and further that I don't quite understand (but I wish to understand!) the problem you're trying to solve.

untergeek:
https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-update.html#_literal_doc_as_upsert_literal

I think the issue needs to be specific to `doc_as_upsert`, rather than generic updates.  Without `doc_as_upsert`, a `doc` field should be treated just like any other field.  In order to fully respect the Elasticsearch API, it would be appropriate to get the Elasticsearch output plugin to also catch and respect this behavior.

salimane:
i wanted to check if this is related to #364 
Thanks

kiryam:
@untergeek Any progress with this issue?

kiryam:
@jordansissel @salimane has error because plugin doesn't use strict api.

salimane:
@kiryam @untergeek what do you guys think of #367 ?

kiryam:
I make new pull request https://github.com/logstash-plugins/logstash-codec-es_bulk/pull/6 into es_bulk codec

PR #363, opened 2016-02-10
Improve URL parsing in 'hosts' to take a full URL

andrewvc:
This will also warn on more error cases, like `localhost:9200/` /cc @pmusa


lib/logstash/outputs/elasticsearch.rb
```diff
@@ -81,8 +81,10 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
   # the root path for the Elasticsearch HTTP API lives.
   config :path, :validate => :string, :default => "/"
 
-  # Enable SSL/TLS secured communication to Elasticsearch cluster
-  config :ssl, :validate => :boolean, :default => false
+  # Enable SSL/TLS secured communication to Elasticsearch cluster. Leaving this unspecified will use whatever scheme
+  # is specified in the URLs listed in 'hosts'. If no explicit protocol is specified plain HTTP will be used.
+  # If SSL is explicitly disabled here the plugin will refuse to start if an HTTPS URL is given in 'hosts'
+  config :ssl, :validate => :boolean
```
suyograo: If `ssl` is `true` and url does not have protocol, do we default to `https`?

> If SSL is explicitly disabled here the plugin will refuse to start if an HTTPS URL is given in 'hosts'
good

suyograo: Ok, checked the code. Does the right thing with `ssl=>true`

spec/unit/outputs/elasticsearch_spec.rb
```diff
@@ -24,6 +24,10 @@
       eso.close
     end
 
+    describe "setting the buffer timeout" do
```
suyograo: remove this?

spec/unit/outputs/elasticsearch/http_client_spec.rb
```diff
@@ -0,0 +1,106 @@
...
+      expect(subject.send(:host_to_url, hostname_port)).to eql(http_hostname_port)
+    end
+
+    it "should raise an error when the URL is an invalid format" do
+      expect {
+        subject.send(:host_to_url, "#{hostname_port}/")
+      }.to raise_error(LogStash::ConfigurationError)
+    end
+
+    it "should parse full URLs correctly" do
```
suyograo: Can we add one test for trailing `/` ? `http://#{hostname_port}/`

andrewvc: For sure

spec/unit/outputs/elasticsearch_spec.rb
andrewvc: Oops! Good catch.

CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## 2.5.0
+ - Host settings now are more robust to bad input and can take (almost) full URLs
```
jordansissel: almost?

Also, I would mention the URL part first since that's the major change. Possibly make this two entries; one about URLs acceptance, and one about the input validation

spec/unit/outputs/elasticsearch/http_client_spec.rb
```diff
@@ -0,0 +1,106 @@
...
+require "logstash/outputs/elasticsearch/http_client"
+require "java"
+
+describe LogStash::Outputs::ElasticSearch::HttpClient do
+  let(:base_options) { {:hosts => ["127.0.0.1"], :logger => Cabin::Channel.get }}
+
+  describe "Host/URL Parsing" do
+    subject { described_class.new(base_options) }
+
+    let(:hostname) { "my.hostname" }
```
suyograo: Also, one more test for hostname with `-`. `frontend-logs.abc.com`

lib/logstash/outputs/elasticsearch/common_configs.rb
```diff
@@ -74,6 +74,8 @@ def self.included(mod)
       # Remember the `http` protocol uses the http://www.elastic.co/guide/en/elasticsearch/reference/current/modules-http.html#modules-http[http] address (eg. 9200, not 9300).
       #     `"127.0.0.1"`
       #     `["127.0.0.1:9200","127.0.0.2:9200"]`
+      #     `["https://127.0.0.1:9200"]`
```
jordansissel: Let's have an example w/ a url without a port.

lib/logstash/outputs/elasticsearch/http_client.rb
```diff
@@ -136,6 +128,54 @@ def build_client(options)
       Elasticsearch::Client.new(client_options)
     end
 
+    HOSTNAME_PORT_REGEX=/\A(?<hostname>([A-Za-z0-9\.\-]+))(:(?<port>\d+))?\Z/
+    # Parse a configuration host to a normalized URL
+    def host_to_url(host, ssl=nil, path=nil)
```
jordansissel: I don't see any instance variables in use here. Is that right? If so, this can probably go in its own module (outside a class instance)

lib/logstash/outputs/elasticsearch/http_client.rb
```diff
@@ -136,6 +128,54 @@ def build_client(options)
       Elasticsearch::Client.new(client_options)
     end
 
+    HOSTNAME_PORT_REGEX=/\A(?<hostname>([A-Za-z0-9\.\-]+))(:(?<port>\d+))?\Z/
```
jordansissel: I'm not sure it's important, but this will fail to match IPv6 address:port combos which usually show up as `[ipv6]:port` (with the brackets, like localhost:9200 ipv6 would be `[::1]:9200`

We can just make a note about this and move on. I'm not sure many folks will need that anyway when DNS is usually simpler to work with ;P

lib/logstash/outputs/elasticsearch/http_client.rb
```diff
@@ -136,6 +128,54 @@ def build_client(options)
...
+    # Parse a configuration host to a normalized URL
+    def host_to_url(host, ssl=nil, path=nil)
+      explicit_scheme = if ssl.nil?
```
jordansissel: feels like this would be better as a case statement instead of a if/else+ternary

lib/logstash/outputs/elasticsearch/http_client.rb
```diff
@@ -136,6 +128,54 @@ def build_client(options)
...
+                          ssl ? "https"  : "http"
+                        end
+
+      # Ensure path starts with a /
+      if path && path[0] != '/'
+        path = "/#{path}"
+      end
+
+      url = nil
+      if host =~ /\A#{URI::regexp(['http', 'https'])}\z/
```
jordansissel: Can you make this regex a constant?

lib/logstash/outputs/elasticsearch/http_client.rb
```diff
@@ -136,6 +128,54 @@ def build_client(options)
...
+        url = URI.parse(host)
+        if url.scheme == "http" && ssl == true
```
jordansissel: style: instead of `ssl == true` and later `ssl == false` -- thoughts on just using `ssl` as the value?

like: `url.scheme == "http" && ssl`

lib/logstash/outputs/elasticsearch/http_client.rb
andrewvc: will fix

lib/logstash/outputs/elasticsearch/http_client.rb
andrewvc: Is this a general principle? In most of the LS codebase I see people weigh single responsibility principle against too many classes.

lib/logstash/outputs/elasticsearch/http_client.rb
andrewvc: That is a good point! I think I should patch this to work.

lib/logstash/outputs/elasticsearch/http_client.rb
jordansissel: For this specific case, I don't have strong feelings, so as a way to not add noise to a review, I retract my comment :)

jordansissel:
Read over the specs and they look good to me. Conditional LGTM if travis passes and the minor comments from myself and @suyograo are addressed.

jordansissel:
Travis failed:

```
....F.................[2016-02-10 22:06:42,053][INFO ][rest.suppressed          ] /some-path//_template/logstash Params: {id=logstash, index=some-path, type=_template}
[some-path] IndexNotFoundException[no such index]

```

https://travis-ci.org/logstash-plugins/logstash-output-elasticsearch/builds/108396566#L257

lib/logstash/outputs/elasticsearch/http_client.rb
andrewvc: What do you think of this regex?
`\A(?<hostname>([A-Za-z0-9\.\-]+)|\[[0-9\:]+\])(:(?<port>\d+))?\Z`

andrewvc: I should say
`\A(?<hostname>([A-Za-z0-9\.\-]+)|\[[0-9A-Fa-f\:]+\])(:(?<port>\d+))?\Z`

lib/logstash/outputs/elasticsearch/http_client.rb
andrewvc: @jordansissel neutral

lib/logstash/outputs/elasticsearch/http_client.rb
andrewvc: Ah, but we're handling the `nil` case differently!

andrewvc: I should probably leave a comment about that.

spec/unit/outputs/elasticsearch/http_client_spec.rb
andrewvc: If you don't mind I'll just put the dash in the hostname for completeness.

CHANGELOG.md
andrewvc: can do!

lib/logstash/outputs/elasticsearch/common_configs.rb
andrewvc: Can do!

andrewvc:
Just addressed the minor comments with a force push + fixed the failing test (the expectation was bad).

suyograo:
LGTM

Issue #377, opened 2016-02-29
Connections not shared across worker instances


andrewvc:
We need to refactor our use of the ruby client and make this worker a threadsafe shared object to make more efficient use of connections.

andrewvc:
This issue will be closed when https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/381 is closed

suyograo:
Fixed in #390

PR #381, opened 2016-03-02
Make this plugin a 'threadsafe' output (requires LS >= 2.2.0)

andrewvc:
Fixes https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/377
Before acceptance this will need https://github.com/elastic/elasticsearch-ruby/pull/281
Merged in and used for the ruby client


andrewvc:
Since threadsafe plugins only support 2.2 I removed `#receive` which meant letting us remove a big chunk of old code.

andrewvc:
I'm realizing now we need to now add connection pool options (max_connections, max_per_host) to limit connections with the shared pool.

lib/logstash/outputs/elasticsearch/buffer.rb
```diff
@@ -1,124 +0,0 @@
-require 'concurrent'
```
suyograo: Was this file/class unused?

suyograo: Ah, I follow - we are using the batches from pipeline so there's no need for `Stud::Buffer`

andrewvc: This was the internal buffer used for `#receive`. In logstash 2.2+ we only call `#multi_receive` so no internal buffering is necessary!

lib/logstash/outputs/elasticsearch/common_configs.rb
```diff
@@ -88,21 +88,11 @@ def self.included(mod)
...
-      # In Logstashes <= 2.1 this plugin uses its own internal buffer of events.
-      # This config option sets that size. In these older logstashes this size may
-      # have a significant impact on heap usage, whereas in 2.2+ it will never increase it.
-      # To make efficient bulk API calls, we will buffer a certain number of
-      # events before flushing that out to Elasticsearch. This setting
-      # controls how many events will be buffered before sending a batch
-      # of events. Increasing the `flush_size` has an effect on Logstash's heap size.
-      # Remember to also increase the heap size using `LS_HEAP_SIZE` if you are sending big documents
-      # or have increased the `flush_size` to a higher value.
       mod.config :flush_size, :validate => :number, :default => 500
```
suyograo: Let's also mark it `deprecated`. We should make it `obsolete` in 5.0

andrewvc: I'm not sure if we want to, @ppf2 has a use case where they have multiple ES outputs only one of which needs to have a flush_size of 1. Since the cost to us is so low (it's like 3 lines of code) I'd rather not deprecate it . I know @jordansissel feels strongly here. We should discuss this further!

jordansissel: What's the use case of keeping flush_size? Why would someone set it to 1?

andrewvc: @ppf2 has a case where they receive occasional 120MB+ events and they don't want to have 2 of them in the same bulk request. Most requests go through the other output with a normal flush_size

andrewvc: @jordansissel to clarify they check to see if they have jumbo messages and shunt those to the `flush_size = 1` ES output.

jordansissel: Gotcha, so the problem is large (byte size) messages. We should move towards an intelligent bulk sizing by bytesize, not by event size, and this decision should be made by the plugin automatically - I know this isn't the case right now, but making this improvement would resolve the need to hardcode "how many documents per bulk" in anyone's config.

ppf2: +1 on having flush size by byte size (eg. https://github.com/elastic/logstash/issues/4000), currently, it's quite cumbersome for the end user having to write their own ruby filter to check for byte size, and define multiple ES output plugins with different flush_size settings to accommodate messages of various sizes (and most of the time, the admins setting up the Elastic Stack has no control over the content of these messages for they are fed in by different application groups).

suyograo: I think its ok to mark it `deprecated` if we have to keep this setting around for just one use case (the 120MB event). IMO this is an edge case. We should discourage folks from using `flush_size` because you have the `pipeline-batch-size`. Having to tweak event batches in 2 places is just confusing. 

Open to discussing this via zoom with a larger group

suyograo: Also, we don't have to do this now - i.e. this PR does not need to be blocked by this decision.

andrewvc: @suyograo I don't think we should mark things as deprecated when there isn't a replacement yet. If we say 'deprecated, use flush_byte_size instead that makes sense, but until then, not a great idea IMHO. It's frustrating to tell a user 'don"t use this, and there's no replacement''

jordansissel: My hope is that we don't have a `flush_byte_size` setting at all.

We don't need to deprecate flush_size in this PR, but before our next minor release, I think we should. We should discourage users from using it now and in the future (which has always been my internal meaning of 'deprecated' -- aka: "do not use").  Let's open a separate ticket to discuss when+how to remove flush_size and continue this specific thread there so we don't cloud up this PR, sound ok?

andrewvc: :+1:

suyograo: Agree with @jordansissel on this. I strongly recommend we get away from having to tune batches in multiple places. But, I completely agree this is out of line for this PR, so lets take this discussion to a new issue :)

suyograo:
Changes looks good to me, pending the upstream elasticsearch-ruby patch. Would also like another set of eyes to take a look at this

PR #387, opened 2016-03-07
Improved .travis.yml to support multiple ES versions + use better formatting

andrewvc:
The .travis.yml file should support multiple ES versions, this fixes this. It also cleans up the output to be more legible and no longer interleaves the background ES process logs with the rspec logs. It will barf out the background logs in the event of an error.


jordansissel:
Briefly reviewed the [travis results](https://travis-ci.org/logstash-plugins/logstash-output-elasticsearch/builds/114301692) and this LGTM.

Nice work! :)

andrewvc:
Thanks @jordansissel !

PR #389, opened 2016-03-10
Set ssl options regardless of :ssl true or false

### LINKED ISSUE #388: https URLs should set ssl options
Using an HTTPS URL will ignore custom keystore settings, this needs to be fixed.

jsvd:
instead of only formatting ssl options if :ssl => true, always do it.

This allows ssl settings to be used if :ssl is set to false but the URL passed in the hosts array uses the https schema (i.e. `https://..`)

fixes #388

### DIFF
--- a/lib/logstash/outputs/elasticsearch/http_client_builder.rb
+++ b/lib/logstash/outputs/elasticsearch/http_client_builder.rb
@@ -59,8 +59,14 @@ module LogStash; module Outputs; class ElasticSearch;
     end
 
     def self.setup_ssl(logger, params)
-      return {} if params["ssl"].nil?
-      return {:ssl => {:enabled => false}} if params["ssl"] == false
+
+      if params["ssl"].nil?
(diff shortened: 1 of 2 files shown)


andrewvc:
LGTM

andrewvc:
@jsvd do you mind squashing and rebasing? jarvis can't merge this.

andrewvc:
@jsvd what is the purpose of this PR? What feature does it enable?

jsvd:
@andrewvc I edited the PR description to describe its purpose

andrewvc:
@jsvd as I wrote in https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/388 I don't think this should work that way. Maybe we should continue the discussion there?

jsvd:
this has been fixed in https://github.com/logstash-plugins/logstash-output-elasticsearch/commit/e7259de52f88643cc3311ce73d4dfa4663041a87

PR #390, opened 2016-03-10
Threadsafety + Exponential Backoff

andrewvc:
Fixes 
https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/377
Requires elasticsearch-ruby from this branch: https://github.com/elastic/elasticsearch-ruby/pull/296

This moves the sniffer to the new one in the elasticsearch_ruby client and adds two new options for connection pooling and one for health checking. The new concept of health checks may break existing clusters where '/' is blocked for security. I'm not sure what the best approach here is since we do want a way to check if a cluster is up without interrupting regular execution flow.

The new `pool_max` and `max_per_route` options set max # of concurrent connections + the maximum concurrent connections to a given backend. The default values of 50 and 30 should be fine given that the new connection selector strategy tries to evenly spread connections across backends.

This also adds exponential backoff to connection retries with a ceiling of `retry_max_interval` which is the most time to wait between retries, and `retry_initial_interval` which is the initial amount of time to wait. `retry_initial_interval` will be increased exponentially between retries until a request succeeds.


lib/logstash/outputs/elasticsearch/http_client/pool.rb
```diff
@@ -0,0 +1,337 @@
...
+        attributes = info['attributes']
+        next if attributes && attributes['data'] == 'false'
+
+        matches = addr_str.match(ES1_RE_URL) || addr_str.match(ES2_RE_URL)
+        if matches
+          host = matches[1].empty? ? matches[2] : matches[1]
+          port = matches[3]
+          info.merge :host => host, :port => port, :id => id
+        end
+      end.compact
```
jsvd: sniff! doesn't seem to do any side effect, or am I reading this wrong?

lib/logstash/outputs/elasticsearch/http_client.rb
```diff
@@ -48,40 +44,22 @@ def non_threadsafe_bulk(actions)
...
-      if options[:sniffing]
-        @sniffer_thread = Thread.new do
-          loop do
-            @request_mutex.synchronize { sniff! }
-            sleep (options[:sniffing_delay].to_f || 30)
-          end
-        end
+      end.
+      flatten.
+      reduce("") do |acc,line|
```
jsvd: have you considered doing just 1 call to Json.dump to the whole array?

lib/logstash/outputs/elasticsearch/http_client/manticore_adapter.rb
```diff
@@ -0,0 +1,57 @@
...
+  class ManticoreAdapter
+    attr_reader :manticore, :logger
+
+    def initialize(logger, options)
+      @logger = logger
+      build_client(options || {})
+    end
+
+    # Should just be run once at startup
+    def build_client(options={})
```
jsvd: it should be more explicit that this is a private api, called only by initialize

lib/logstash/outputs/elasticsearch/http_client/pool.rb
```diff
@@ -0,0 +1,337 @@
...
+      DEFAULT_OPTIONS.merge(options).tap do |merged|
+        @healthcheck_path = merged[:healthcheck_path]
+        @scheme = merged[:scheme]
+        @resurrect_interval = merged[:resurrect_interval]
+        @auth = merged[:auth]
+        @sniffing = merged[:sniffing]
+        @sniffer_delay = merged[:sniffer_delay]
+      end
+
+      # Override the scheme if one is explicitly set in urls
```
jsvd: can we remove this functionality perhaps, and stop supporting the config option :scheme?

lib/logstash/outputs/elasticsearch/http_client_builder.rb
```diff
@@ -1,14 +1,21 @@
...
       common_options = {
         :client_settings => client_settings,
-        :sniffing => params["sniffing"],
-        :sniffing_delay => params["sniffing_delay"]
+        :healthcheck_path => params["healthcheck_path"]
       }
 
+      if params["sniffing"]
+        common_options[:sniffing] = true
+        common_options[:sniffer_delay] = params["sniffing__delay"]
```
jsvd: s/__/_

lib/logstash/outputs/elasticsearch/common.rb
```diff
@@ -91,10 +76,21 @@ def retrying_submit(actions)
...
       end
     end
 
+    def next_sleep_interval(current_interval)
+      doubled = current_interval * 2
+      doubled > @retry_max_interval ? @retry_max_interval : doubled
+    end
+
     def submit(actions)
       es_actions = actions.map { |a, doc, event| [a, doc, event.to_hash]}
```
jsvd: I don't see a reason to precompute the actions hash in submit just to avoid the retry mechanism in safe_bulk. I'd rather not have an inlined begin/rescue block in safe_bulk

lib/logstash/outputs/elasticsearch/http_client/pool.rb
```diff
@@ -0,0 +1,337 @@
...
+
+    def stop_resurrectionist
+      @resurrectionist.join
+    end
+
+    def perform_request(method, path, params={}, body=nil)
+      with_connection do |url|
+        resp = perform_request_to_url(url, method, path, params, body)
+
+        raise BadResponseError.new(resp) if resp.code > 399
```
jsvd: This means that any error >= 400 is not retried?

lib/logstash/outputs/elasticsearch/common.rb
```diff
@@ -159,37 +155,40 @@ def get_event_type(event)
...
-      )
-
-      @logger.debug("Failed actions for last bad bulk request!", :actions => actions)
-
-      raise e
+      sleep_interval = @retry_initial_interval
+      begin
+        url, response = @client.bulk(es_actions)
+        response
+      rescue Manticore::SocketException, Manticore::SocketTimeout => e
```
jsvd: in https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/390/files#diff-885c6dfd0ed36754f38c5c925b275036R53 we define a list of exceptions known to mean unreachable host, and this list is then used in https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/390/files#diff-ba322aad25863c7bb2706a30288a4416R208.

This means SocketException will never reach this, while HostUnreachableError might

lib/logstash/outputs/elasticsearch/common_configs.rb
```diff
@@ -145,8 +135,11 @@ def self.included(mod)
       # if enabled, script is in charge of creating non-existent document (scripted update)
       mod.config :scripted_upsert, :validate => :boolean, :default => false
 
+      # Set initial interval between bulk retries. Doubled on each retry up to `retry_max_interval`
```
jsvd: the unit of time should be described here.
Also as an initial retry interval, I'd suggest something much lower like 0.01 or 0.1 seconds

lib/logstash/outputs/elasticsearch/http_client.rb
```diff
@@ -17,25 +17,21 @@ def initialize(options={})
...
+      if template_exists?(name) && !force
+        @logger.debug("Found existing Elasticsearch template. Skipping template management", :name => name)
+        return
       end
+      template_put(name, template)
     end
 
     def bulk(actions)
-      @request_mutex.synchronize { non_threadsafe_bulk(actions) }
+      non_threadsafe_bulk(actions)
```
jsvd: I know this is old code but I think we could just move the code from non_threadsafe_bulk here and document, if necessary, its limitations

lib/logstash/outputs/elasticsearch/http_client/pool.rb
```diff
@@ -0,0 +1,337 @@
...
+        end
+      end
+    end
+
+    def resurrect_dead!
+      # Try to keep locking granularity low such that we don't affect IO...
+      @state_mutex.synchronize { @url_info.select {|url,meta| meta[:dead] } }.each do |url,meta|
+        begin
+          @logger.info("Checking url #{url} with path #{@healthcheck_path} to see if node resurrected")
+          perform_request_to_url(url, "GET", @healthcheck_path)
```
jsvd: since we're not performing any validation on the response, I wonder if we could switch to HEAD.

lib/logstash/outputs/elasticsearch/http_client/pool.rb
```diff
@@ -0,0 +1,337 @@
...
+
+      update_urls(urls)
+      start_resurrectionist
+      start_sniffer if @sniffing
+    end
+
+    def close
+      @state_mutex.synchronize { @stopping = true }
+
+      logger.debug  "Stopping resurrectionist"
```
jsvd: s/resurrectionist/sniffer

lib/logstash/outputs/elasticsearch/http_client/pool.rb
jsvd: ensure BadResponseError is catched correctly

lib/logstash/outputs/elasticsearch/common.rb
andrewvc: You're 100% right. I'm switching this to only rescue `::LogStash::Outputs::ElasticSearch::HttpClient::Pool::HostUnreachableError` and have added `Manticore::SocketTimeout` to the list of exceptions rescued there

lib/logstash/outputs/elasticsearch/http_client/pool.rb
andrewvc: Let's leave this for a more wholistic PR that better addresses the URL situation.

lib/logstash/outputs/elasticsearch/http_client.rb
andrewvc: Not really possible with this format since we need the newlines in specific places.

lib/logstash/outputs/elasticsearch/http_client/pool.rb
andrewvc: Yep. I'm going to remove this error class entirely.

lib/logstash/outputs/elasticsearch/http_client/manticore_adapter.rb
andrewvc: I'm just going to move this code straight into 'initialize'

lib/logstash/outputs/elasticsearch/http_client/pool.rb
andrewvc: good catch, will fix. I'll split it into a pure and impure method.

andrewvc:
@jsvd I force pushed a mostly squashed and rebased off master version that incorporates most of your feedback. I will be added additional tests soon.

lib/logstash/outputs/elasticsearch/http_client/pool.rb
andrewvc: Fixed. Now `check_sniff` does the pure stuff.

andrewvc:
Ran some performance tests. It is ~20% more efficient than the current plugin CPU wise. Doesn't make a huge impact on the wall clock but should help out some other filters!

```
# Old plugin. 1 worker per core
~/p/logstash ((3d875c8...)) $ curl -XDELETE 'http://127.0.0.1:9200/*'; time bin/logstash -e 'input { generator { count => 3000000 } } output { elasticsearch { workers => 8} }'
{"acknowledged":true}--- jar coordinate com.fasterxml.jackson.core:jackson-annotations already loaded with version 2.7.1 - omit version 2.7.0
--- jar coordinate com.fasterxml.jackson.core:jackson-databind already loaded with version 2.7.1 - omit version 2.7.1-1
Settings: Default pipeline workers: 8
Pipeline main started
Pipeline main has been shutdown
stopping pipeline {:id=>"main", :level=>:warn}
      135.62 real       385.20 user        27.90 sys

# New Plugin
~/p/logstash ((3d875c8...)) $ curl -XDELETE 'http://127.0.0.1:9200/*'; time bin/logstash -e 'input { generator { count => 3000000 } } output { elasticsearch { }}'
{"acknowledged":true}--- jar coordinate com.fasterxml.jackson.core:jackson-annotations already loaded with version 2.7.1 - omit version 2.7.0
--- jar coordinate com.fasterxml.jackson.core:jackson-databind already loaded with version 2.7.1 - omit version 2.7.1-1
Settings: Default pipeline workers: 8
Pipeline main started
Pipeline main has been shutdown
stopping pipeline {:id=>"main", :level=>:warn}
      125.13 real       303.91 user        25.56 sys

```

I also did a custom setup on the cloud benchmarker and saw these numbers on the apache logs benchmark:

```
# old plugin
results/benchmarks/benchmark_apachelogs_http.conf.results: 2170 events/sec . (390762 total events)
# new plugin
results/benchmarks/benchmark_apachelogs_http.conf.results: 2031 events/sec . (365737 total events)
```

So maybe 6% faster there. Small, but I'll take it!

andrewvc:
@jsvd this is ready for another round of reviews. I squashed the last version you reviewed. All subsequent commits are new.

This now has:

1. Additional specs
2. Fixed a lack of 'sleep' calls in the infinite loops in pool.rb causing a big slowdown
3. Many small bugs fixed
4. Many tests fixed.

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 4.0.0
+ - Make this plugin threadsafe. Workers no longer needed or supported
+ - Add pool_max and pool_max_per_route options
```
talevy: mention the http_client here

lib/logstash/outputs/elasticsearch/http_client/pool.rb
talevy: I agree. out of the scope of this PR, but we should revisit the scheme logic that went on

andrewvc: I'm going to make a PR to logstash/core that adds a 'url' type. Maybe then we can ditch all non-URL config?

jsvd:
`Requires elasticsearch-ruby from this branch: elastic/elasticsearch-ruby#296`
is this still true?

lib/logstash/outputs/elasticsearch/common.rb
```diff
@@ -168,38 +164,42 @@ def get_event_type(event)
...
-        :client_config => @client.client_options,
-      )
-
-      @logger.debug("Failed actions for last bad bulk request!", :actions => actions)
-
-      raise e
+    def safe_bulk(actions)
+      sleep_interval = @retry_initial_interval
+      begin
+        es_actions = actions.map {|action_type, params, event| [action_type, params, event.to_hash]}
```
jsvd: should we move this line outside of the begin block to avoid recomputing this on retry? I see no downside to it

lib/logstash/outputs/elasticsearch/common.rb
```diff
@@ -168,38 +164,42 @@ def get_event_type(event)
...
+        response = @client.bulk(es_actions)
+        response
+      rescue ::LogStash::Outputs::ElasticSearch::HttpClient::Pool::HostUnreachableError
+        # If we can't even connect to the server let's just print out the URL (:hosts is actually a URL)
+        # and let the user sort it out from there
+        @logger.error(
+          "Attempted to send a bulk request to Elasticsearch configured at '#{@client.client_options[:hosts]}',"+
```
jsvd: I don't think `client_options` exists anymore?

lib/logstash/outputs/elasticsearch/common.rb
andrewvc: @jsvd the overhead is minimal and it really makes understanding the flow much simpler. Before it was always confusing (actions + es_actions). I think @suyograo requested this change originally? Or was it @talevy ?

lib/logstash/outputs/elasticsearch/http_client/pool.rb
```diff
@@ -0,0 +1,339 @@
...
+
+
+        # Only connect to nodes that serve data
+        # this will skip connecting to client, tribe, and master only nodes
+        # Note that if 'attributes' is NOT set, then that's just a regular node
+        # with master + data + client enabled, so we allow that
+        attributes = info['attributes']
+        next if attributes && attributes['data'] == 'false'
+
+        matches = addr_str.match(ES1_RE_URL) || addr_str.match(ES2_RE_URL)
```
jsvd: the names of the regexp constant don't match the ones in lines 129/130

lib/logstash/outputs/elasticsearch/http_client/pool.rb
```diff
@@ -0,0 +1,339 @@
...
+    # Sniffs the cluster then updates the internal URLs
+    def sniff!
+      update_urls(check_sniff)
+    end
+
+    ES1_SNIFF_RE_URL  = /\[([^\/]*)?\/?([^:]*):([0-9]+)\]/
+    ES2_SNIFF_RE_URL  = /([^\/]*)?\/?([^:]*):([0-9]+)/
+    # Sniffs and returns the results. Does not update internal URLs!
+    def check_sniff
+      resp = perform_request('GET', '_nodes')
```
jsvd: the manticore_adapter perform_request accepts verbs as symbols and here we're passing a string

jsvd: also, perform_request returns not the manticore response but an array of [url, resp]. therefore this line needs to be more like:

`url, resp = perform_request(:get, '_nodes')`

jsvd:
I haven't been able yet to make sniffing work, I added some comments about the exceptions I'm seeing and their fix, but currently at:

```
% bin/logstash -f ../config
--- jar coordinate com.fasterxml.jackson.core:jackson-annotations already loaded with version 2.7.1 - omit version 2.7.0
--- jar coordinate com.fasterxml.jackson.core:jackson-databind already loaded with version 2.7.1 - omit version 2.7.1-1
Settings: Default pipeline workers: 4
Pipeline main started
Error while performing sniffing {:error_message=>"Only URI objects may be passed in!", :class=>"ArgumentError", :backtrace=>["/Users/joaoduarte/new_plugins/logstash-output-elasticsearch/lib/logstash/outputs/elasticsearch/http_client/pool.rb:217:in `normalize_url'", "org/jruby/RubyArray.java:2414:in `map'", "/Users/joaoduarte/new_plugins/logstash-output-elasticsearch/lib/logstash/outputs/elasticsearch/http_client/pool.rb:233:in `update_urls'", "/Users/joaoduarte/new_plugins/logstash-output-elasticsearch/lib/logstash/outputs/elasticsearch/http_client/pool.rb:126:in `sniff!'", "/Users/joaoduarte/new_plugins/logstash-output-elasticsearch/lib/logstash/outputs/elasticsearch/http_client/pool.rb:120:in `start_sniffer'", "/Users/joaoduarte/new_plugins/logstash-output-elasticsearch/lib/logstash/outputs/elasticsearch/http_client/pool.rb:104:in `until_stopped'", "/Users/joaoduarte/new_plugins/logstash-output-elasticsearch/lib/logstash/outputs/elasticsearch/http_client/pool.rb:120:in `start_sniffer'"], :level=>:warn}
^CSIGINT received. Shutting down the agent. {:level=>:warn}
stopping pipeline {:id=>"main", :level=>:warn}

```

jsvd:
My current manual test against a 4 node cluster (node1 to node4) leads to data loss when:

0. logstash is indexing 1 million events from generator w/ sniffing true
1. cluster is green
2. ctrl+c the node that is both the master and the one logstash was pointing to (node1)
3. wait for the cluster to become green again
4. join the node1 again

wait for logstash to terminate and the do `GET /logstash-*/_count` which yielded 947879

lib/logstash/outputs/elasticsearch/common.rb
suyograo: wasn't me. probably @talevy

talevy: twas me

jsvd:
@andrewvc I think that's enough for a first round a review? I can redo/continue whenever you want.
By the way, I used the following patch on the code to make it work:

```diff
diff --git a/lib/logstash/outputs/elasticsearch/common.rb b/lib/logstash/outputs/elasticsearch/common.rb
index e8d4495..3e744eb 100644
--- a/lib/logstash/outputs/elasticsearch/common.rb
+++ b/lib/logstash/outputs/elasticsearch/common.rb
@@ -170,11 +170,11 @@ module LogStash; module Outputs; class ElasticSearch;
         es_actions = actions.map {|action_type, params, event| [action_type, params, event.to_hash]}
         response = @client.bulk(es_actions)
         response
-      rescue ::LogStash::Outputs::ElasticSearch::HttpClient::Pool::HostUnreachableError
+      rescue ::LogStash::Outputs::ElasticSearch::HttpClient::Pool::HostUnreachableError => e
         # If we can't even connect to the server let's just print out the URL (:hosts is actually a URL)
         # and let the user sort it out from there
         @logger.error(
-          "Attempted to send a bulk request to Elasticsearch configured at '#{@client.client_options[:hosts]}',"+
+          "Attempted to send a bulk request to Elasticsearch configured at localhost," + # '#{@client.client_options[:hosts]}',"+
             " but Elasticsearch appears to be unreachable or down!",
           :error_message => e.message,
           :class => e.class.name
diff --git a/lib/logstash/outputs/elasticsearch/http_client/pool.rb b/lib/logstash/outputs/elasticsearch/http_client/pool.rb
index 63542eb..b3592ed 100644
--- a/lib/logstash/outputs/elasticsearch/http_client/pool.rb
+++ b/lib/logstash/outputs/elasticsearch/http_client/pool.rb
@@ -130,7 +130,7 @@ module LogStash; module Outputs; class ElasticSearch; class HttpClient;
     ES2_SNIFF_RE_URL  = /([^\/]*)?\/?([^:]*):([0-9]+)/
     # Sniffs and returns the results. Does not update internal URLs!
     def check_sniff
-      resp = perform_request('GET', '_nodes')
+      url, resp = perform_request(:get, '_nodes')
       parsed = LogStash::Json.load(resp.body)
       parsed['nodes'].map do |id,info|
         # TODO Make sure this works with shield. Does that listed
@@ -146,13 +146,13 @@ module LogStash; module Outputs; class ElasticSearch; class HttpClient;
         attributes = info['attributes']
         next if attributes && attributes['data'] == 'false'
 
-        matches = addr_str.match(ES1_RE_URL) || addr_str.match(ES2_RE_URL)
+        matches = addr_str.match(ES1_SNIFF_RE_URL) || addr_str.match(ES2_SNIFF_RE_URL)
         if matches
           host = matches[1].empty? ? matches[2] : matches[1]
           port = matches[3]
           info.merge :host => host, :port => port, :id => id
         end
-      end.compact
+      end.compact.map {|url_hash| URI.parse("http://#{url_hash['http_address']}") }
     end
 
     def stop_sniffer
@@ -176,7 +176,7 @@ module LogStash; module Outputs; class ElasticSearch; class HttpClient;
       @state_mutex.synchronize { @url_info.select {|url,meta| meta[:dead] } }.each do |url,meta|
         begin
           @logger.info("Checking url #{url} with path #{@healthcheck_path} to see if node resurrected")
-          perform_request_to_url(url, "HEAD", @healthcheck_path)
+          perform_request_to_url(url, :head, @healthcheck_path)
           # If no exception was raised it must have succeeded!
           logger.warn("Resurrected connection to dead ES instance at #{url}")
           @state_mutex.synchronize { meta[:dead] = false }
```

lib/logstash/outputs/elasticsearch/http_client/pool.rb
andrewvc: Good call. It is now fixed :)

andrewvc:
@jsvd I've pushed additional changes and fixed the sniffing!

It should be ready for another round of review. All tests passing on my end

jsvd:
When I kill a node during ingestion, I get:

Attempted to send a bulk request to Elasticsearch configured at '["http://localhost:9201/"]', but an error occurred and it failed! Are you sure you can reach elasticsearch from this machine using the configuration provided? {:error_message=>"127.0.0.1:9204 failed to respond", :error_class=>"Manticore::ClientProtocolException",

The message is a bit misleading, as it refers 2 different es instances: 9201 and 9204 (I killed the latter)

jsvd:
I'm now testing with a 5 node cluster, and with a 5p/1r index I get data loss if I kill 2 nodes, making the cluster red for a while.
From what I understand, ES will fail the writes so I think LS isn't behaving correctly here.

lib/logstash/outputs/elasticsearch.rb
```diff
@@ -130,14 +132,35 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
...
+  # This sets the maximum number of open connections the output will create.
+  # Setting this too low may mean frequently closing / opening connections
+  # which is bad.
+  config :pool_max, :validate => :number, :default => 1000
+
+  # While the output tries to reuse connections efficiently we have a maximum per endpoint.
+  # This sets the maximum number of open connections per endpoint the output will create.
+  # Setting this too low may mean frequently closing / opening connections
+  # which is bad.
+  config :pool_max_per_route, :validate => :number, :default => 100
```
jsvd: I think the terms here can get a bit confusing if you aren't aware of the underlying httpclient library being used.
In that sense it's not obvious how a user should set these parameters for their 3, 10, 100 node cluster.

During ingestion, how many distinct endpoints (routes) are going to be hit?
a) The number of nodes (hosts)?
b) number of nodes * 2 (_bulk for ingest and _nodes for sniff)? 

We should clarify here the meaning of endpoint or route (and perhaps use only 1 of the terms), and also give a few hints on how to scale this value with the cluster size?

lib/logstash/outputs/elasticsearch/http_client.rb
jsvd: this is still an issue:

```
logstash-output-elasticsearch (git)-[exponential-threadsafe] % grep client_options . -r
./lib/logstash/outputs/elasticsearch/common.rb:          "Attempted to send a bulk request to Elasticsearch configured at '#{@client.client_options[:hosts]}',"+
```

jsvd:
One thing I find annoying is that, during an error such as an 503, I always see duplicate error messages from logstash, even though one is an error and the other a warning:

```
[503] {"error":{"root_cause":[{"type":"cluster_block_exception","reason":"blocked by: [SERVICE_UNAVAILABLE/2/no master];"}],"type":"cluster_block_exception","reason":"blocked by: [SERVICE_UNAVAILABLE/2/no master];"},"status":503} {:class=>"Elasticsearch::Transport::Transport::Errors::ServiceUnavailable", :backtrace=>["/Users/joaoduarte/experiments/test_es_output/logstash-2.3.1/vendor/bundle/jruby/1.9/gems/elasticsearch-transport-1.0.15/lib/elasticsearch/transport/transport/base.rb:146:in `__raise_transport_error'", "/Users/joaoduarte/experiments/test_es_output/logstash-2.3.1/vendor/bundle/jruby/1.9/gems/elasticsearch-transport-1.0.15/lib/elasticsearch/transport/transport/base.rb:256:in `perform_request'", "/Users/joaoduarte/experiments/test_es_output/logstash-2.3.1/vendor/bundle/jruby/1.9/gems/elasticsearch-transport-1.0.15/lib/elasticsearch/transport/transport/http/manticore.rb:54:in `perform_request'", "/Users/joaoduarte/experiments/test_es_output/logstash-2.3.1/vendor/bundle/jruby/1.9/gems/elasticsearch-transport-1.0.15/lib/elasticsearch/transport/client.rb:125:in `perform_request'", "/Users/joaoduarte/experiments/test_es_output/logstash-2.3.1/vendor/bundle/jruby/1.9/gems/elasticsearch-api-1.0.15/lib/elasticsearch/api/actions/bulk.rb:87:in `bulk'", "/Users/joaoduarte/experiments/test_es_output/logstash-2.3.1/vendor/bundle/jruby/1.9/gems/logstash-output-elasticsearch-2.5.5-java/lib/logstash/outputs/elasticsearch/http_client.rb:53:in `non_threadsafe_bulk'", "/Users/joaoduarte/experiments/test_es_output/logstash-2.3.1/vendor/bundle/jruby/1.9/gems/logstash-output-elasticsearch-2.5.5-java/lib/logstash/outputs/elasticsearch/http_client.rb:38:in `bulk'", "org/jruby/ext/thread/Mutex.java:149:in `synchronize'", "/Users/joaoduarte/experiments/test_es_output/logstash-2.3.1/vendor/bundle/jruby/1.9/gems/logstash-output-elasticsearch-2.5.5-java/lib/logstash/outputs/elasticsearch/http_client.rb:38:in `bulk'", "/Users/joaoduarte/experiments/test_es_output/logstash-2.3.1/vendor/bundle/jruby/1.9/gems/logstash-output-elasticsearch-2.5.5-java/lib/logstash/outputs/elasticsearch/common.rb:163:in `safe_bulk'", "/Users/joaoduarte/experiments/test_es_output/logstash-2.3.1/vendor/bundle/jruby/1.9/gems/logstash-output-elasticsearch-2.5.5-java/lib/logstash/outputs/elasticsearch/common.rb:101:in `submit'", "/Users/joaoduarte/experiments/test_es_output/logstash-2.3.1/vendor/bundle/jruby/1.9/gems/logstash-output-elasticsearch-2.5.5-java/lib/logstash/outputs/elasticsearch/common.rb:86:in `retrying_submit'", "/Users/joaoduarte/experiments/test_es_output/logstash-2.3.1/vendor/bundle/jruby/1.9/gems/logstash-output-elasticsearch-2.5.5-java/lib/logstash/outputs/elasticsearch/common.rb:29:in `multi_receive'", "org/jruby/RubyArray.java:1653:in `each_slice'", "/Users/joaoduarte/experiments/test_es_output/logstash-2.3.1/vendor/bundle/jruby/1.9/gems/logstash-output-elasticsearch-2.5.5-java/lib/logstash/outputs/elasticsearch/common.rb:28:in `multi_receive'", "/Users/joaoduarte/experiments/test_es_output/logstash-2.3.1/vendor/bundle/jruby/1.9/gems/logstash-core-2.3.1-java/lib/logstash/output_delegator.rb:130:in `worker_multi_receive'", "/Users/joaoduarte/experiments/test_es_output/logstash-2.3.1/vendor/bundle/jruby/1.9/gems/logstash-core-2.3.1-java/lib/logstash/output_delegator.rb:114:in `multi_receive'", "/Users/joaoduarte/experiments/test_es_output/logstash-2.3.1/vendor/bundle/jruby/1.9/gems/logstash-core-2.3.1-java/lib/logstash/pipeline.rb:301:in `output_batch'", "org/jruby/RubyHash.java:1342:in `each'", "/Users/joaoduarte/experiments/test_es_output/logstash-2.3.1/vendor/bundle/jruby/1.9/gems/logstash-core-2.3.1-java/lib/logstash/pipeline.rb:301:in `output_batch'", "/Users/joaoduarte/experiments/test_es_output/logstash-2.3.1/vendor/bundle/jruby/1.9/gems/logstash-core-2.3.1-java/lib/logstash/pipeline.rb:232:in `worker_loop'", "/Users/joaoduarte/experiments/test_es_output/logstash-2.3.1/vendor/bundle/jruby/1.9/gems/logstash-core-2.3.1-java/lib/logstash/pipeline.rb:201:in `start_workers'"], :level=>:warn}
Attempted to send a bulk request to Elasticsearch configured at '["http://localhost:9200/"]', but an error occurred and it failed! Are you sure you can reach elasticsearch from this machine using the configuration provided? {:error_message=>"[503] {\"error\":{\"root_cause\":[{\"type\":\"cluster_block_exception\",\"reason\":\"blocked by: [SERVICE_UNAVAILABLE/2/no master];\"}],\"type\":\"cluster_block_exception\",\"reason\":\"blocked by: [SERVICE_UNAVAILABLE/2/no master];\"},\"status\":503}", :error_class=>"Elasticsearch::Transport::Transport::Errors::ServiceUnavailable", :backtrace=>["/Users/joaoduarte/experiments/test_es_output/logstash-2.3.1/vendor/bundle/jruby/1.9/gems/elasticsearch-transport-1.0.15/lib/elasticsearch/transport/transport/base.rb:146:in `__raise_transport_error'", "/Users/joaoduarte/experiments/test_es_output/logstash-2.3.1/vendor/bundle/jruby/1.9/gems/elasticsearch-transport-1.0.15/lib/elasticsearch/transport/transport/base.rb:256:in `perform_request'", "/Users/joaoduarte/experiments/test_es_output/logstash-2.3.1/vendor/bundle/jruby/1.9/gems/elasticsearch-transport-1.0.15/lib/elasticsearch/transport/transport/http/manticore.rb:54:in `perform_request'", "/Users/joaoduarte/experiments/test_es_output/logstash-2.3.1/vendor/bundle/jruby/1.9/gems/elasticsearch-transport-1.0.15/lib/elasticsearch/transport/client.rb:125:in `perform_request'", "/Users/joaoduarte/experiments/test_es_output/logstash-2.3.1/vendor/bundle/jruby/1.9/gems/elasticsearch-api-1.0.15/lib/elasticsearch/api/actions/bulk.rb:87:in `bulk'", "/Users/joaoduarte/experiments/test_es_output/logstash-2.3.1/vendor/bundle/jruby/1.9/gems/logstash-output-elasticsearch-2.5.5-java/lib/logstash/outputs/elasticsearch/http_client.rb:53:in `non_threadsafe_bulk'", "/Users/joaoduarte/experiments/test_es_output/logstash-2.3.1/vendor/bundle/jruby/1.9/gems/logstash-output-elasticsearch-2.5.5-java/lib/logstash/outputs/elasticsearch/http_client.rb:38:in `bulk'", "org/jruby/ext/thread/Mutex.java:149:in `synchronize'", "/Users/joaoduarte/experiments/test_es_output/logstash-2.3.1/vendor/bundle/jruby/1.9/gems/logstash-output-elasticsearch-2.5.5-java/lib/logstash/outputs/elasticsearch/http_client.rb:38:in `bulk'", "/Users/joaoduarte/experiments/test_es_output/logstash-2.3.1/vendor/bundle/jruby/1.9/gems/logstash-output-elasticsearch-2.5.5-java/lib/logstash/outputs/elasticsearch/common.rb:163:in `safe_bulk'", "/Users/joaoduarte/experiments/test_es_output/logstash-2.3.1/vendor/bundle/jruby/1.9/gems/logstash-output-elasticsearch-2.5.5-java/lib/logstash/outputs/elasticsearch/common.rb:101:in `submit'", "/Users/joaoduarte/experiments/test_es_output/logstash-2.3.1/vendor/bundle/jruby/1.9/gems/logstash-output-elasticsearch-2.5.5-java/lib/logstash/outputs/elasticsearch/common.rb:86:in `retrying_submit'", "/Users/joaoduarte/experiments/test_es_output/logstash-2.3.1/vendor/bundle/jruby/1.9/gems/logstash-output-elasticsearch-2.5.5-java/lib/logstash/outputs/elasticsearch/common.rb:29:in `multi_receive'", "org/jruby/RubyArray.java:1653:in `each_slice'", "/Users/joaoduarte/experiments/test_es_output/logstash-2.3.1/vendor/bundle/jruby/1.9/gems/logstash-output-elasticsearch-2.5.5-java/lib/logstash/outputs/elasticsearch/common.rb:28:in `multi_receive'", "/Users/joaoduarte/experiments/test_es_output/logstash-2.3.1/vendor/bundle/jruby/1.9/gems/logstash-core-2.3.1-java/lib/logstash/output_delegator.rb:130:in `worker_multi_receive'", "/Users/joaoduarte/experiments/test_es_output/logstash-2.3.1/vendor/bundle/jruby/1.9/gems/logstash-core-2.3.1-java/lib/logstash/output_delegator.rb:114:in `multi_receive'", "/Users/joaoduarte/experiments/test_es_output/logstash-2.3.1/vendor/bundle/jruby/1.9/gems/logstash-core-2.3.1-java/lib/logstash/pipeline.rb:301:in `output_batch'", "org/jruby/RubyHash.java:1342:in `each'", "/Users/joaoduarte/experiments/test_es_output/logstash-2.3.1/vendor/bundle/jruby/1.9/gems/logstash-core-2.3.1-java/lib/logstash/pipeline.rb:301:in `output_batch'", "/Users/joaoduarte/experiments/test_es_output/logstash-2.3.1/vendor/bundle/jruby/1.9/gems/logstash-core-2.3.1-java/lib/logstash/pipeline.rb:232:in `worker_loop'", "/Users/joaoduarte/experiments/test_es_output/logstash-2.3.1/vendor/bundle/jruby/1.9/gems/logstash-core-2.3.1-java/lib/logstash/pipeline.rb:201:in `start_workers'"], :client_config=>{:hosts=>["http://localhost:9200/"], :ssl=>nil, :transport_options=>{:socket_timeout=>0, :request_timeout=>0, :proxy=>nil, :ssl=>{}}, :transport_class=>Elasticsearch::Transport::Transport::HTTP::Manticore, :logger=>nil, :tracer=>nil, :reload_connections=>false, :retry_on_failure=>false, :reload_on_failure=>false, :randomize_hosts=>false}, :level=>:error}
```

lib/logstash/outputs/elasticsearch/common.rb
jsvd: this is still an issue:

```
logstash-output-elasticsearch (git)-[exponential-threadsafe] % grep client_options . -r
./lib/logstash/outputs/elasticsearch/common.rb:          "Attempted to send a bulk request to Elasticsearch configured at '#{@client.client_options[:hosts]}',"+
```

jsvd:
I'm seeing duplicate messages in logstash during an error:

```
Settings: Default pipeline workers: 4
Pipeline main started
Marking url http://127.0.0.1:9200 as dead. Last error: [LogStash::Outputs::ElasticSearch::HttpClient::Pool::HostUnreachableError] [Manticore::ClientProtocolException] 127.0.0.1:9200 failed to respond {:level=>:warn}
Marking url http://127.0.0.1:9200 as dead. Last error: [LogStash::Outputs::ElasticSearch::HttpClient::Pool::HostUnreachableError] [Manticore::ClientProtocolException] 127.0.0.1:9200 failed to respond {:level=>:warn}
Encountered an unexpected error submitting a bulk request! Will retry. {:error_message=>"undefined method `client_options' for #<LogStash::Outputs::ElasticSearch::HttpClient:0x4d16931e>", :class=>"NoMethodError", :backtrace=>["/Users/joaoduarte/new_plugins/logstash-output-elasticsearch/lib/logstash/outputs/elasticsearch/common.rb:177:in `safe_bulk'", "/Users/joaoduarte/new_plugins/logstash-output-elasticsearch/lib/logstash/outputs/elasticsearch/common.rb:172:in `safe_bulk'", "/Users/joaoduarte/new_plugins/logstash-output-elasticsearch/lib/logstash/outputs/elasticsearch/common.rb:95:in `submit'", "/Users/joaoduarte/new_plugins/logstash-output-elasticsearch/lib/logstash/outputs/elasticsearch/common.rb:71:in `retrying_submit'", "/Users/joaoduarte/new_plugins/logstash-output-elasticsearch/lib/logstash/outputs/elasticsearch/common.rb:23:in `multi_receive'", "org/jruby/RubyArray.java:1653:in `each_slice'", "/Users/joaoduarte/new_plugins/logstash-output-elasticsearch/lib/logstash/outputs/elasticsearch/common.rb:22:in `multi_receive'", "/Users/joaoduarte/experiments/test_es_output/logstash/logstash-core/lib/logstash/output_delegator.rb:132:in `threadsafe_multi_receive'", "/Users/joaoduarte/experiments/test_es_output/logstash/logstash-core/lib/logstash/output_delegator.rb:119:in `multi_receive'", "/Users/joaoduarte/experiments/test_es_output/logstash/logstash-core/lib/logstash/pipeline.rb:345:in `output_batch'", "org/jruby/RubyHash.java:1342:in `each'", "/Users/joaoduarte/experiments/test_es_output/logstash/logstash-core/lib/logstash/pipeline.rb:345:in `output_batch'", "/Users/joaoduarte/experiments/test_es_output/logstash/logstash-core/lib/logstash/pipeline.rb:273:in `worker_loop'", "/Users/joaoduarte/experiments/test_es_output/logstash/logstash-core/lib/logstash/pipeline.rb:234:in `start_workers'"], :level=>:error}
Encountered an unexpected error submitting a bulk request! Will retry. {:error_message=>"undefined method `client_options' for #<LogStash::Outputs::ElasticSearch::HttpClient:0x4d16931e>", :class=>"NoMethodError", :backtrace=>["/Users/joaoduarte/new_plugins/logstash-output-elasticsearch/lib/logstash/outputs/elasticsearch/common.rb:177:in `safe_bulk'", "/Users/joaoduarte/new_plugins/logstash-output-elasticsearch/lib/logstash/outputs/elasticsearch/common.rb:172:in `safe_bulk'", "/Users/joaoduarte/new_plugins/logstash-output-elasticsearch/lib/logstash/outputs/elasticsearch/common.rb:95:in `submit'", "/Users/joaoduarte/new_plugins/logstash-output-elasticsearch/lib/logstash/outputs/elasticsearch/common.rb:71:in `retrying_submit'", "/Users/joaoduarte/new_plugins/logstash-output-elasticsearch/lib/logstash/outputs/elasticsearch/common.rb:23:in `multi_receive'", "org/jruby/RubyArray.java:1653:in `each_slice'", "/Users/joaoduarte/new_plugins/logstash-output-elasticsearch/lib/logstash/outputs/elasticsearch/common.rb:22:in `multi_receive'", "/Users/joaoduarte/experiments/test_es_output/logstash/logstash-core/lib/logstash/output_delegator.rb:132:in `threadsafe_multi_receive'", "/Users/joaoduarte/experiments/test_es_output/logstash/logstash-core/lib/logstash/output_delegator.rb:119:in `multi_receive'", "/Users/joaoduarte/experiments/test_es_output/logstash/logstash-core/lib/logstash/pipeline.rb:345:in `output_batch'", "org/jruby/RubyHash.java:1342:in `each'", "/Users/joaoduarte/experiments/test_es_output/logstash/logstash-core/lib/logstash/pipeline.rb:345:in `output_batch'", "/Users/joaoduarte/experiments/test_es_output/logstash/logstash-core/lib/logstash/pipeline.rb:273:in `worker_loop'", "/Users/joaoduarte/experiments/test_es_output/logstash/logstash-core/lib/logstash/pipeline.rb:234:in `start_workers'"], :level=>:error}
```

jsvd:
Killing a master node still causes data loss on a 3 node cluster, the same doesn't seem to happen in Logstash 2.3.1 vanilla

jsvd:
One thing to be aware about sniffing is that, because the initial connection doesn't resolve hosts to ips, this can happen:

```
Elasticsearch pool adding node @ URL http://localhost:9201 {:level=>:info}
Using mapping template from {:path=>nil, :level=>:info}
Attempting to install template {:manage_template=>{"template"=>"logstash-*", "settings"=>{"index.refresh_interval"=>"5s", "index.number_of_shards"=>3}, "mappings"=>{"_default_"=>{"_all"=>{"enabled"=>true, "omit_norms"=>true}, "dynamic_templates"=>[{"message_field"=>{"match"=>"message", "match_mapping_type"=>"string", "mapping"=>{"type"=>"string", "index"=>"analyzed", "omit_norms"=>true, "fielddata"=>{"format"=>"disabled"}}}}, {"string_fields"=>{"match"=>"*", "match_mapping_type"=>"string", "mapping"=>{"type"=>"string", "index"=>"analyzed", "omit_norms"=>true, "fielddata"=>{"format"=>"disabled"}, "fields"=>{"raw"=>{"type"=>"string", "index"=>"not_analyzed", "ignore_above"=>256}}}}}], "properties"=>{"@timestamp"=>{"type"=>"date"}, "@version"=>{"type"=>"string", "index"=>"not_analyzed"}, "geoip"=>{"dynamic"=>true, "properties"=>{"ip"=>{"type"=>"ip"}, "location"=>{"type"=>"geo_point"}, "latitude"=>{"type"=>"float"}, "longitude"=>{"type"=>"float"}}}}}}}, :level=>:info}
New Elasticsearch output {:class=>"LogStash::Outputs::ElasticSearch", :hosts=>["http://localhost:9201"], :level=>:info}
Starting pipeline {"id"=>"main", "pipeline.workers"=>4, "pipeline.batch.size"=>125, "pipeline.batch.delay"=>5, "pipeline.max_inflight"=>500, :level=>:info}
Pipeline main started
Elasticsearch pool adding node @ URL http://127.0.0.1:9202 {:level=>:info}
Elasticsearch pool adding node @ URL http://127.0.0.1:9201 {:level=>:info}
Elasticsearch pool adding node @ URL http://127.0.0.1:9200 {:level=>:info}
Elasticsearch pool removing node @ URL http://localhost:9201 {:level=>:info}
```

I guess we could resolve the host name at startup time and be done with it? We're already removing the initial uri/host from the pool anyway.

jsvd:
Latest data loss test:

Elasticsearch 2.3.2, 3 node cluster, all on the same machine, minimum.master.nodes set to 2:

node1 - localhost:9202
node2 - localhost:9201
node3 - localhost:9200

Logstash master (current [HEAD](https://github.com/elastic/logstash/commit/bdddab9)), with local logstash-output-elasticsearch git clone using andrewvc/exponential-threadsafe branch at latest [HEAD](https://github.com/andrewvc/logstash-output-elasticsearch/commit/85ad40f)
Changed template so that primary shard count is 3 instead of 5

Logstash config:

```
input { generator { count => 1000000 } }
output {
  elasticsearch {
    hosts => "http://localhost:9201"
    sniffing => true
  }
}
```

##### Test script

1. Start cluster so that node2 (localhost:9201) is master. This means logstash will connect to master
2. Launch logstash 
3. At around 250k events ingested, ctrl+c node2 and start it again
4. Measure /logstash-*/_count after logstash terminates
5. Use scroll request to check the absense of sequence numbers from 0 to 1 million


##### Results

Logstash output:

```
logstash (git)-[master] % bin/logstash -f ../config --log.level=verbose
--- jar coordinate com.fasterxml.jackson.core:jackson-annotations already loaded with version 2.7.1 - omit version 2.7.0
--- jar coordinate com.fasterxml.jackson.core:jackson-databind already loaded with version 2.7.1 - omit version 2.7.1-1
-------- Logstash Settings (* means modified) --------- {:level=>:info}
node.name: "Joaos-MBP.lan" {:level=>:info}
*path.config: "../config" {:level=>:info}
config.test_and_exit: false {:level=>:info}
config.reload.automatic: false {:level=>:info}
config.reload.interval: 3 {:level=>:info}
metric.collect: true {:level=>:info}
path.settings: "/Users/joaoduarte/experiments/test_es_output/logstash/config" {:level=>:info}
pipeline.id: "main" {:level=>:info}
pipeline.workers: 4 {:level=>:info}
pipeline.output.workers: 1 {:level=>:info}
pipeline.batch.size: 125 {:level=>:info}
pipeline.batch.delay: 5 {:level=>:info}
pipeline.unsafe_shutdown: false {:level=>:info}
path.plugins: [] {:level=>:info}
config.debug: false {:level=>:info}
*log.level: "verbose" (default: "warn") {:level=>:info}
version: false {:level=>:info}
help: false {:level=>:info}
log.format: "plain" {:level=>:info}
http.host: "127.0.0.1" {:level=>:info}
http.port: 9600 {:level=>:info}
--------------- Logstash Settings ------------------- {:level=>:info}
starting agent {:level=>:info}
starting pipeline {:id=>"main", :level=>:info}
Elasticsearch pool adding node @ URL http://localhost:9201 {:level=>:info}
Using mapping template from {:path=>nil, :level=>:info}
Attempting to install template {:manage_template=>{"template"=>"logstash-*", "settings"=>{"index.refresh_interval"=>"5s", "index.number_of_shards"=>3}, "mappings"=>{"_default_"=>{"_all"=>{"enabled"=>true, "omit_norms"=>true}, "dynamic_templates"=>[{"message_field"=>{"match"=>"message", "match_mapping_type"=>"string", "mapping"=>{"type"=>"string", "index"=>"analyzed", "omit_norms"=>true, "fielddata"=>{"format"=>"disabled"}}}}, {"string_fields"=>{"match"=>"*", "match_mapping_type"=>"string", "mapping"=>{"type"=>"string", "index"=>"analyzed", "omit_norms"=>true, "fielddata"=>{"format"=>"disabled"}, "fields"=>{"raw"=>{"type"=>"string", "index"=>"not_analyzed", "ignore_above"=>256}}}}}], "properties"=>{"@timestamp"=>{"type"=>"date"}, "@version"=>{"type"=>"string", "index"=>"not_analyzed"}, "geoip"=>{"dynamic"=>true, "properties"=>{"ip"=>{"type"=>"ip"}, "location"=>{"type"=>"geo_point"}, "latitude"=>{"type"=>"float"}, "longitude"=>{"type"=>"float"}}}}}}}, :level=>:info}
New Elasticsearch output {:class=>"LogStash::Outputs::ElasticSearch", :hosts=>["http://localhost:9201"], :level=>:info}
Starting pipeline {"id"=>"main", "pipeline.workers"=>4, "pipeline.batch.size"=>125, "pipeline.batch.delay"=>5, "pipeline.max_inflight"=>500, :level=>:info}
Pipeline main started
Elasticsearch pool adding node @ URL http://127.0.0.1:9202 {:level=>:info}
Elasticsearch pool adding node @ URL http://127.0.0.1:9200 {:level=>:info}
Elasticsearch pool adding node @ URL http://127.0.0.1:9201 {:level=>:info}
Elasticsearch pool removing node @ URL http://localhost:9201 {:level=>:info}
Marking url http://127.0.0.1:9201 as dead. Last error: [LogStash::Outputs::ElasticSearch::HttpClient::Pool::HostUnreachableError] [Manticore::ClientProtocolException] 127.0.0.1:9201 failed to respond {:level=>:warn}
Attempted to send a bulk request to Elasticsearch configured at 'somewhere', but Elasticsearch appears to be unreachable or down! {:error_message=>"[Manticore::ClientProtocolException] 127.0.0.1:9201 failed to respond", :class=>"LogStash::Outputs::ElasticSearch::HttpClient::Pool::HostUnreachableError", :level=>:error}
Checking url http://127.0.0.1:9201 with path / to see if node resurrected {:level=>:info}
Elasticsearch pool removing node @ URL http://127.0.0.1:9201 {:level=>:info}
Input plugins stopped! Will shutdown filter/output workers. {:level=>:info}
Pipeline main has been shutdown
stopping pipeline {:id=>"main", :level=>:warn}
Closing inputs {:level=>:info}
Closed inputs {:level=>:info}
```

Count after test:

```
% curl -XGET "localhost:9200/logstash-*/_count/?pretty"
{
  "count" : 989125,
  "_shards" : {
    "total" : 3,
    "successful" : 3,
    "failed" : 0
  }
}
```

Results from querying elasticsearch for sequence numbers:

```
logstash-2.3.1 % bin/logstash -f check_bulk.config --quiet | sort -n | uniq > numbers && ruby check_numbers.rb
missing 10875!!! too many to show
```

jsvd:
I've just run the same test on logstash-2.3.1, with no data loss, full outputs in gist: https://gist.github.com/jsvd/acd7d920ddefd45d25a894f97c25b020

jsvd:
Results today are amazing, so far no data loss :+1:

jsvd:
Regarding error messages, sometimes it seems either manticore or httpclient leaks through:

```
May 24, 2016 12:45:44 PM org.apache.http.impl.execchain.RetryExec execute
INFO: I/O exception (java.net.SocketException) caught when processing request to {}->http://127.0.0.1:9203: Broken pipe
May 24, 2016 12:45:44 PM org.apache.http.impl.execchain.RetryExec execute
INFO: Retrying request to {}->http://127.0.0.1:9203
Marking url as dead. Last error: [LogStash::Outputs::ElasticSearch::HttpClient::Pool::HostUnreachableError] Elasticsearch Unreachable: [http://127.0.0.1:9203][Manticore::SocketException] Connection refused {:url=>#<URI::HTTP:0x6a379255 URL:http://127.0.0.1:9203>, :error_message=>"Elasticsearch Unreachable: [http://127.0.0.1:9203][Manticore::SocketException] Connection refused", :error_class=>"LogStash::Outputs::ElasticSearch::HttpClient::Pool::HostUnreachableError", :level=>:warn}
Attempted to send a bulk request to elasticsearch' but Elasticsearch appears to be unreachable or down! {:error_message=>"Elasticsearch Unreachable: [http://127.0.0.1:9203][Manticore::SocketException] Connection refused", :class=>"LogStash::Outputs::ElasticSearch::HttpClient::Pool::HostUnreachableError", :will_retry_in_seconds=>2, :level=>:error}
```

while other times:
```
Marking url as dead. Last error: [LogStash::Outputs::ElasticSearch::HttpClient::Pool::HostUnreachableError] Elasticsearch Unreachable: [http://127.0.0.1:9203][Manticore::SocketException] Connection refused {:url=>#<URI::HTTP:0x45875c38 URL:http://127.0.0.1:9203>, :error_message=>"Elasticsearch Unreachable: [http://127.0.0.1:9203][Manticore::SocketException] Connection refused", :error_class=>"LogStash::Outputs::ElasticSearch::HttpClient::Pool::HostUnreachableError", :level=>:warn}
Attempted to send a bulk request to elasticsearch' but Elasticsearch appears to be unreachable or down! {:error_message=>"Elasticsearch Unreachable: [http://127.0.0.1:9203][Manticore::SocketException] Connection refused", :class=>"LogStash::Outputs::ElasticSearch::HttpClient::Pool::HostUnreachableError", :will_retry_in_seconds=>4, :level=>:error}
```

andrewvc:
@jsvd the extra log messages from Apache HC are from having manticore retry idempotent requests (that means sniffing + health checks). 

I'll disable that setting because we have our own retry logic.

lib/logstash/outputs/elasticsearch.rb
```diff
@@ -130,14 +132,36 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
...
+  # Resurrection is the process by which backend endpoints marked 'down' are checked
+  # to see if they have come back to life
+  config :resurrect_delay, :validate => :number, :default => 5
+
   def build_client
     @client = ::LogStash::Outputs::ElasticSearch::HttpClientBuilder.build(@logger, @hosts, params)
   end
 
   def close
+    puts "SUBMIT COUNT: #{@submit_count} AC: #{self.client.action_count} ADAPT: #{self.client.pool.adapter.submitted_times}|#{self.client.pool.adapter.submitted_lines}"
```
talevy: remove print statement?

andrewvc:
@tal @jsvd I think I've addressed all your concerns. Specifically I removed ApacheHC's built-in retries. I also fixed some failing tests and fixed an issue where template installs couldn't proceed (which caused a test failure)

It could use a last pass but its probably good to merge!

jsvd:
LGTM

PR #393, opened 2016-03-14
bump manticore minimum version due to obj leak

### LINKED ISSUE #392: Sniffing in 2.2.2 causes memory leak & OOM
Hi,
   I run few instances of logstash, since upgrade to 2.2.2 i started to notice permanent crashes caused by OOM errors. Bigger heap only increase interval between OOMs. I checked heap dump and found excessive amount of **org.apache.http.config.Registry** objects. 

Out of 3GB I had 887000 Registries taking 1.4GB of memory.
screen shot 2016-03-14 at 8 39 01 am

I started to monitory tenured memory utilization and amount of Registries object instances on the heap for each logstash instance.

**Number of Registry objects 1GB heap (jmap -gchisto)**

screen shot 2016-03-14 at 8 38 16 am

**Tenured memory utilization 1GB heap (jstat -gcutil)**
screen shot 2016-03-14 at 8 41 41 am

Drops from 85% to few percent indicate watchdog killed process as there was no point to waste CPU due to CMS cycles as memory won't be recycled.

You may notice one series of data on those screenshots that looks pretty fine, that's instance with sniffing set to false, all others had sniffing => true

Once disabled sniffing on all instances the problem is gone.

**Disabled Sniffing: Number of Registry objects 1GB heap (jmap -gchisto)**
screen shot 2016-03-14 at 8 54 32 am

jsvd:
see more in https://github.com/cheald/manticore/issues/45

fixes #392

### DIFF
--- a/logstash-output-elasticsearch.gemspec
+++ b/logstash-output-elasticsearch.gemspec
@@ -1,7 +1,7 @@
 Gem::Specification.new do |s|
 
   s.name            = 'logstash-output-elasticsearch'
-  s.version         = '2.5.2'
+  s.version         = '2.5.3'
   s.licenses        = ['apache-2.0']
   s.summary         = "Logstash Output to Elasticsearch"
   s.description     = "Output events to elasticsearch"
@@ -31,7 +31,7 @@ Gem::Specification.new do |s|
 
   if RUBY_PLATFORM == 'java'
     s.platform = RUBY_PLATFORM
-    s.add_runtime_dependency "manticore", '>= 0.5.2', '< 1.0.0'
+    s.add_runtime_dependency "manticore", '>= 0.5.4', '< 1.0.0'
   end
 
   s.add_development_dependency 'logstash-devutils'
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 2.5.3
+ - Bump minimum manticore version to 0.5.4 which fixes a memory leak (#392)
+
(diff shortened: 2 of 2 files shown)


logstash-output-elasticsearch.gemspec
```diff
@@ -31,7 +31,7 @@ Gem::Specification.new do |s|
 
   if RUBY_PLATFORM == 'java'
     s.platform = RUBY_PLATFORM
-    s.add_runtime_dependency "manticore", '>= 0.5.2', '< 1.0.0'
+    s.add_runtime_dependency "manticore", '>= 0.5.4', '< 1.0.0'
```
suyograo: why not pin it to a version?

suyograo: also this might create a dependency issue with http-input?

suyograo: should work: https://github.com/logstash-plugins/logstash-mixin-http_client/blob/master/logstash-mixin-http_client.gemspec#L21

jsvd: and that's why we don't pin a version :D

suyograo:
LGTM

suyograo:
@jsvd can you add more info to the changelog that it actually fixes a memory leak and reference the manticore issue?

jsvd:
done

suyograo:
Thanks. Please merge and publish :)

PR #394, opened 2016-03-15
elasticsearch api expects source to be wrapped in data field

yanjunh:
This is to fix bug discussed at https://github.com/elastic/logstash/issues/4796

When source document contains field named "data", the bulk request is not generated correctly.

### DIFF
--- a/lib/logstash/outputs/elasticsearch/http_client.rb
+++ b/lib/logstash/outputs/elasticsearch/http_client.rb
@@ -44,7 +44,7 @@ module LogStash; module Outputs; class ElasticSearch;
         args, source = update_action_builder(args, source) if action == 'update'
 
         if source && action != 'delete'
-          next [ { action => args }, source ]
+          next [ { action => args.merge({ :data => source }) } ]
         else
           next { action => args }
         end


andrewvc:
Thanks for reporting the issue @yanjunh . This appears to be fallout from our switch to using the official es-ruby client which handles the 'data' key specially. (See https://github.com/elastic/elasticsearch-ruby/blob/master/elasticsearch-api/lib/elasticsearch/api/actions/bulk.rb#L5) for more details.

Confirmed testing with:

```
 curl -XDELETE 'http://localhost:9200/lstest/'; bin/logstash -e 'input { generator { count => 3 add_field => {data => foodata}} } output { elasticsearch { index => lstest action => update document_id => 1 doc_as_upsert => true} }'; curl http://localhost:9200/_refresh; curl 'http://localhost:9200/lstest/_search?pretty'
```

This needs tests, I will submit these in a new PR.

andrewvc:
LGTM

andrewvc:
@yanjunh this is now fixed in master as part of a major refactor. We'll release a 2.x compatible version of what's in master ASAP.

Thanks for bringing attention to this however.

PR #398, opened 2016-03-21
BREAKING CHANGE! New default template for ES 5.0

### LINKED ISSUE #386: Mapping changes in Elasticsearch 5.0 for string field type
Elasticsearch is changing `string` field to split them into two variants: `keyword` and `text`. More info about this here: https://github.com/elastic/elasticsearch/issues/12394

The LS template should be changed to reflect this. `string_fields` rule in template should be split to `text_fields` and `keyword_fields` rule. All the `fields.raw` should be updated to `keyword` type and `@message` should be `text`. 

`@version` should be `keyword` too. 

With this change, LS will be locked with ES version, even with HTTP protocol because this change is not backward compatible.

untergeek:
Do not use this with older versions of Elasticsearch.  It will not work.

This template has been tested with the alpha1 release of Elasticsearch 5.0.0

fixes #386

### DIFF
--- a/lib/logstash/outputs/elasticsearch/elasticsearch-template.json
+++ b/lib/logstash/outputs/elasticsearch/elasticsearch-template.json
@@ -32,7 +32,6 @@
         "@timestamp": { "type": "date" },
         "@version": { "type": "string", "index": "not_analyzed" },
         "geoip"  : {
-          "type" : "object",
           "dynamic": true,
           "properties" : {
             "ip": { "type": "ip" },
--- a/lib/logstash/outputs/elasticsearch/elasticsearch-template.json
+++ b/lib/logstash/outputs/elasticsearch/elasticsearch-template.json
@@ -5,30 +5,32 @@
   },
   "mappings" : {
     "_default_" : {
+      "_all" : {"enabled" : true, "omit_norms" : true},
       "dynamic_templates" : [ {
         "message_field" : {
           "match" : "message",
-          "match_mapping_type" : "text",
+          "match_mapping_type" : "string",
           "mapping" : {
-            "type" : "text", 
-            "norms" : false }
+            "type" : "string", "index" : "analyzed", "omit_norms" : true,
+            "fielddata" : { "format" : "disabled" }
           }
-        }, {
-        "text_fields" : {
+        }
+      }, {
+        "string_fields" : {
           "match" : "*",
-          "match_mapping_type" : "text",
+          "match_mapping_type" : "string",
           "mapping" : {
-            "type" : "text", 
-            "norms" : false,
+            "type" : "string", "index" : "analyzed", "omit_norms" : true,
+            "fielddata" : { "format" : "disabled" },
             "fields" : {
-              "raw" : { "type": "keyword", "ignore_above" : 256}
+              "raw" : {"type": "string", "index" : "not_analyzed", "ignore_above" : 256}
             }
           }
         }
       } ],
       "properties" : {
         "@timestamp": { "type": "date" },
-        "@version": { "type": "keyword" },
+        "@version": { "type": "string", "index": "not_analyzed" },
         "geoip"  : {
           "type" : "object",
           "dynamic": true,
--- a/lib/logstash/outputs/elasticsearch/elasticsearch-template.json
+++ b/lib/logstash/outputs/elasticsearch/elasticsearch-template.json
@@ -5,88 +5,38 @@
   },
   "mappings" : {
     "_default_" : {
-      "_all" : {"enabled" : true, "omit_norms" : true},
       "dynamic_templates" : [ {
         "message_field" : {
           "match" : "message",
-          "match_mapping_type" : "string",
+          "match_mapping_type" : "text",
           "mapping" : {
-            "type" : "string", "index" : "analyzed", "omit_norms" : true,
-            "fielddata" : { "format" : "disabled" }
+            "type" : "text", 
+            "norms" : false }
           }
-        }
-      }, {
-        "string_fields" : {
+        }, {
+        "text_fields" : {
           "match" : "*",
-          "match_mapping_type" : "string",
+          "match_mapping_type" : "text",
           "mapping" : {
-            "type" : "string", "index" : "analyzed", "omit_norms" : true,
-            "fielddata" : { "format" : "disabled" },
+            "type" : "text", 
+            "norms" : false,
             "fields" : {
-              "raw" : {"type": "string", "index" : "not_analyzed", "doc_values" : true, "ignore_above" : 256}
+              "raw" : { "type": "keyword", "ignore_above" : 256}
             }
           }
         }
-      }, {
-        "float_fields" : {
-          "match" : "*",
-          "match_mapping_type" : "float",
-          "mapping" : { "type" : "float", "doc_values" : true }
-        }
-      }, {
-        "double_fields" : {
-          "match" : "*",
-          "match_mapping_type" : "double",
-          "mapping" : { "type" : "double", "doc_values" : true }
-        }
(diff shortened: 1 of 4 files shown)


logstash-output-elasticsearch.gemspec
```diff
@@ -1,7 +1,7 @@
 Gem::Specification.new do |s|
 
   s.name            = 'logstash-output-elasticsearch'
-  s.version         = '2.5.4'
+  s.version         = '3.0.0'
```
suyograo: Lets actually make it 5.0.0 so we immediately know what we're dealing with when a user reports an issue. normally, we want plugins to have their own version life cycle and follow semver, but this may be a special case where it warrants to track ES 5.0. Otherwise it may lead to a lot of confusing users and us

untergeek: Right.  I set it to 3.0.0 before we discussed it, then didn't change it after.

logstash-output-elasticsearch.gemspec
```diff
@@ -1,7 +1,7 @@
 Gem::Specification.new do |s|
 
   s.name            = 'logstash-output-elasticsearch'
-  s.version         = '2.5.4'
+  s.version         = '5.0.0'
```
talevy: is this final `5.0.0` version? should we put an `alpha` or `dev` here for now?

talevy: or `5.0.0-SNAPSHOT`

untergeek: set it to alpha1 to match the ES release level.

untergeek: I set it to alpha1 to match the ES version.

lib/logstash/outputs/elasticsearch/elasticsearch-template.json
```diff
@@ -5,88 +5,37 @@
   },
   "mappings" : {
     "_default_" : {
-      "_all" : {"enabled" : true, "omit_norms" : true},
```
jpountz: Should we keep disabling norms? With the new syntax, this would be `"_all" : {"norms" : false}`

spec/integration/outputs/templates_spec.rb
```diff
@@ -73,7 +73,7 @@
   end
 
   it "make [geoip][location] a geo_point" do
-    results = @es.search(:body => { "filter" => { "geo_distance" => { "distance" => "1000km", "geoip.location" => { "lat" => 0.5, "lon" => 0.5 } } } })
+    results = @es.search(:body => { "query" => { "bool" => { "must" => { "match_all" => {} }, "filter" => { "geo_distance" => { "distance" => "1000km", "geoip.location" => { "lat" => 0.5, "lon" => 0.5 } } } } } })
```
jpountz: you could just do:
`results = @es.search(:body => { "query" => { "geo_distance" => { "distance" => "1000km", "geoip.location" => { "lat" => 0.5, "lon" => 0.5 } } } })` (replacing `filter` with `query`)

jpountz:
This looks great.

lib/logstash/outputs/elasticsearch/elasticsearch-template.json
untergeek: If I recall correctly, when I tested a _new_ 5.0 compatible template (rather than just hoping the existing template would be adapted on the fly) it didn't work with `"norms" : false`.  I got some error or other.

jpountz: Hmm this must be a bug. :( I just tried locally and it worked for me, do you have steps that I can follow to reproduce?

pickypg:
I like that it no longer has the boilerplate dynamic templates for the base types, as it's effectively implied (including with doc values) starting in 2.x. Now that this requires 5.0, it makes total sense!

spec/integration/outputs/templates_spec.rb
suyograo: ah nice, I forget :)

untergeek:
We are going to re-do this with an improved version of the current template that will be backward compatible.  @jpountz has done some awesome things that will automagically translate old templates into new mappings.  We will still prune the old "helper" mappings for building your own numeric types, and since doc_values is default, we'll remove the hard-mappings there as well.

untergeek:
I `PUT` the template to the 5.0.0 alpha RC1 release and inserted a single document:

```
curl -XPUT localhost:9200/logstash-2016.03.30/logs/1 -d '
> { "@timestamp" : "2016-03-30T17:45:00.000Z", "message" : "This is a message field" }
> '
{"_index":"logstash-2016.03.30","_type":"logs","_id":"1","_version":1,"_shards":{"total":2,"successful":1,"failed":0},"created":true}
```

When I pulled the mapping, this is what I see:

```json
{
  "logstash-2016.03.30" : {
    "mappings" : {
      "logs" : {
        "_all" : {
          "enabled" : true,
          "norms" : false
        },
        "dynamic_templates" : [ {
          "message_field" : {
            "match" : "message",
            "mapping" : {
              "fielddata" : {
                "format" : "disabled"
              },
              "index" : "analyzed",
              "omit_norms" : true,
              "type" : "string"
            }
          }
        }, {
          "string_fields" : {
            "match" : "*",
            "mapping" : {
              "fielddata" : {
                "format" : "disabled"
              },
              "fields" : {
                "raw" : {
                  "ignore_above" : 256,
                  "index" : "not_analyzed",
                  "type" : "string"
                }
              },
              "index" : "analyzed",
              "omit_norms" : true,
              "type" : "string"
            }
          }
        } ],
        "properties" : {
          "@timestamp" : {
            "type" : "date",
            "format" : "strict_date_optional_time||epoch_millis"
          },
          "@version" : {
            "type" : "keyword"
          },
          "geoip" : {
            "dynamic" : "true",
            "properties" : {
              "ip" : {
                "type" : "ip"
              },
              "latitude" : {
                "type" : "float"
              },
              "location" : {
                "type" : "geo_point"
              },
              "longitude" : {
                "type" : "float"
              }
            }
          },
          "message" : {
            "type" : "text",
            "norms" : false,
            "fielddata" : false
          }
        }
      },
      "_default_" : {
        "_all" : {
          "enabled" : true,
          "norms" : false
        },
        "dynamic_templates" : [ {
          "message_field" : {
            "match" : "message",
            "mapping" : {
              "fielddata" : {
                "format" : "disabled"
              },
              "index" : "analyzed",
              "omit_norms" : true,
              "type" : "string"
            }
          }
        }, {
          "string_fields" : {
            "match" : "*",
            "mapping" : {
              "fielddata" : {
                "format" : "disabled"
              },
              "fields" : {
                "raw" : {
                  "ignore_above" : 256,
                  "index" : "not_analyzed",
                  "type" : "string"
                }
              },
              "index" : "analyzed",
              "omit_norms" : true,
              "type" : "string"
            }
          }
        } ],
        "properties" : {
          "@timestamp" : {
            "type" : "date",
            "format" : "strict_date_optional_time||epoch_millis"
          },
          "@version" : {
            "type" : "keyword"
          },
          "geoip" : {
            "dynamic" : "true",
            "properties" : {
              "ip" : {
                "type" : "ip"
              },
              "latitude" : {
                "type" : "float"
              },
              "location" : {
                "type" : "geo_point"
              },
              "longitude" : {
                "type" : "float"
              }
            }
          }
        }
      }
    }
  }
}
```

suyograo:
LGTM

suyograo:
@untergeek copied this to https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/402 caused i had a lot of merge conflicts.

untergeek:
Closing this in favor of #402 then.

PR #399, opened 2016-03-24
Remove Unmatched Fields from Template

pickypg:
Currently the template includes specific types, such as "byte" and "float", which are never recognized by
the dynamic template system. The template system instead checks for the widest variant of the data type
(e.g., "long" for integers and "double" for floating point).

This removes those unmatched types to avoid the assumption that they are matched.


pickypg:
This dupes #398. Closing.

untergeek:
@pickypg Those fields were initially added as something new users building their own templates could borrow from.  It was a known thing that they were not functional.  The usefulness has not panned out, so we are removing them.

Issue #408, opened 2016-04-07
Docs say to set retry_on_conflict however the setting does not exist


jjfalling:
https://www.elastic.co/guide/en/logstash/2.2/plugins-outputs-elasticsearch.html#_retry_policy shows the retry policy changed and if the user sees 409s then they should set `retry_on_conflict`, however this setting does not exist.

andrewvc:
This will be addressed in https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/409 apologies for the issue!

PR #409, opened 2016-04-07
Add retry_on_conflict configuration option

andrewvc:
Fixes https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/408 by adding a `retry_on_conflict` operation. This only makes sense for update operations of course.

### DIFF
--- a/lib/logstash/outputs/elasticsearch/common.rb
+++ b/lib/logstash/outputs/elasticsearch/common.rb
@@ -138,6 +138,7 @@ module LogStash; module Outputs; class ElasticSearch;
       if @action == 'update'
         params[:_upsert] = LogStash::Json.load(event.sprintf(@upsert)) if @upsert != ""
         params[:_script] = event.sprintf(@script) if @script != ""
+        params[:_retry_on_conflict] = @retry_on_conflict
       end
       params
     end
--- a/lib/logstash/outputs/elasticsearch/common_configs.rb
+++ b/lib/logstash/outputs/elasticsearch/common_configs.rb
@@ -151,6 +151,11 @@ module LogStash; module Outputs; class ElasticSearch
       # DEPRECATED This setting no longer does anything. If you need to change the number of retries in flight
       # try increasing the total number of workers to better handle this.
       mod.config :retry_max_items, :validate => :number, :default => 500, :deprecated => true
+
+      # The number of times Elasticsearch should internally retry an update/upserted document
+      # See the https://www.elastic.co/guide/en/elasticsearch/guide/current/partial-updates.html[partial updates]
+      # for more info
+      mod.config :retry_on_conflict, :validate => :number, :default => 1
     end
   end
 end end end
--- a/logstash-output-elasticsearch.gemspec
+++ b/logstash-output-elasticsearch.gemspec
@@ -1,7 +1,7 @@
 Gem::Specification.new do |s|
 
   s.name            = 'logstash-output-elasticsearch'
-  s.version         = '2.6.0'
+  s.version         = '2.6.1'
   s.licenses        = ['apache-2.0']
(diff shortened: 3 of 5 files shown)


Gemfile
```diff
@@ -1,3 +1,5 @@
 source 'https://rubygems.org'
 
-gemspec
\ No newline at end of file
+gemspec
+
+#gem 'elasticsearch-transport', :path => '/Users/andrewvc/projects/elasticsearch-ruby/elasticsearch-transport'
```
ph: needed?

lib/logstash/outputs/elasticsearch/common_configs.rb
```diff
@@ -151,6 +151,11 @@ def self.included(mod)
       # DEPRECATED This setting no longer does anything. If you need to change the number of retries in flight
       # try increasing the total number of workers to better handle this.
       mod.config :retry_max_items, :validate => :number, :default => 500, :deprecated => true
+
+      # The number of times Elasticsearch should internally retry an update/upserted document
+      # See the https://www.elastic.co/guide/en/elasticsearch/guide/current/partial-updates.html[partial updates]
+      # for more info
+      mod.config :retry_on_conflict, :validate => :number, :default => 3
```
ph: We might want to use 0, which is the default value of ES?

lib/logstash/outputs/elasticsearch/common.rb
```diff
@@ -138,6 +138,7 @@ def event_action_params(event)
       if @action == 'update'
         params[:_upsert] = LogStash::Json.load(event.sprintf(@upsert)) if @upsert != ""
         params[:_script] = event.sprintf(@script) if @script != ""
+        params[:_retry_on_conflict] = @retry_on_conflict
```
ph: This is send to the elasticsearch ruby client right? I though this options was without an underscore?

https://github.com/elastic/elasticsearch-ruby/blob/master/elasticsearch-api/lib/elasticsearch/api/actions/update.rb#L50 

Maybe I don't point to the right doc? I am a bit confused.

Gemfile
andrewvc: Good catch, will remove.

lib/logstash/outputs/elasticsearch/common.rb
andrewvc: This is different because we use the bulk API. These options are passed directly per https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html#bulk-update

Note the underscore there.

lib/logstash/outputs/elasticsearch/common_configs.rb
andrewvc: I dunno, I think that could surprise users if these happen rarely. I think 3 is sensible. Maybe compromise on 1?

ph: +1  for 1

lib/logstash/outputs/elasticsearch/common.rb
ph: perfect! forgot about that small difference.

logstash-output-elasticsearch.gemspec
```diff
@@ -1,7 +1,7 @@
 Gem::Specification.new do |s|
 
   s.name            = 'logstash-output-elasticsearch'
-  s.version         = '2.5.2'
+  s.version         = '2.5.3'
```
ph: update changelog maybe?

ph:
LGTM, one of the integration test hang though

NandanPhadke:
Current docs still don't show this field. Is this field available in logstash version 2.3.4? https://www.elastic.co/guide/en/logstash/current/plugins-outputs-elasticsearch.html

PR #413, opened 2016-04-19
Add tests for #394

andrewvc:
Adds tests + version bump for #394

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 2.6.2
+ - Fix bug where update index actions would not work with events with 'data' field
+
 ## 2.6.1
  - Add 'retry_on_conflict' configuration option which should have been here from the beginning
  
--- a/spec/integration/outputs/update_spec.rb
+++ b/spec/integration/outputs/update_spec.rb
@@ -53,8 +53,20 @@ describe "Update actions", :integration => true do
       r = @es.get(:index => 'logstash-update', :type => 'logs', :id => "123", :refresh => true)
       insist { r["_source"]["message"] } == 'updated message here'
     end
+
+    # The es ruby client treats the data field differently. Make sure this doesn't
+    # raise an exception
+    it "should update an existing document that has a 'data' field" do
+      subject = get_es_output({ 'document_id' => "123" })
+      subject.register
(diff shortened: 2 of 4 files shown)


talevy:
interesting, I was not aware of this format

ref: https://github.com/elastic/elasticsearch-ruby/blob/20c1b6fec6f5eb5410eaf8599bca6f79608579ef/elasticsearch-api/test/unit/bulk_test.rb#L32
https://github.com/elastic/elasticsearch-ruby/blob/20c1b6fec6f5eb5410eaf8599bca6f79608579ef/elasticsearch-api/lib/elasticsearch/api/utils.rb#L93

andrewvc:
@talevy yeah, I would say it's not the best feature in the client, be c'est la vie.

spec/integration/outputs/update_spec.rb
```diff
@@ -53,8 +53,19 @@ def get_es_output( options={} )
       r = @es.get(:index => 'logstash-update', :type => 'logs', :id => "123", :refresh => true)
       insist { r["_source"]["message"] } == 'updated message here'
     end
+
+    # The es ruby client treats the data field differently. Make sure this doesn't
+    # raise an exception
+    it "should update an existing document that has a 'data' field" do
+      subject = get_es_output({ 'document_id' => "123" })
+      subject.register
+      subject.receive(LogStash::Event.new("data" => "updated message here"))
```
talevy: can we make this event include a sibling field beside `"data"`, just to be more explicit that not all your fields necessarily will be found in `data`

CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## 2.6.2
+ - Fix bug where update index actions would not work with events with 'data' field
```
talevy: this "bug" is a very specific format prerequisite that we should add to the docs

andrewvc: This patch doesn't affect user behavior. It fixes the bug where events with a data field wouldn't work.

Am I missing something?

talevy: or, sorry. looks like this fix does the opposite. removes this from the behavior. ignore above

andrewvc:
Fixed the test to include another field.

talevy:
LGTM

PR #414, opened 2016-04-20
Clean up retry specs

andrewvc:
Try to fix the intermittent test failures by cleaning up specs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
 sudo: false
 env:
-  - INTEGRATION=false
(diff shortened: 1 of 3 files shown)


suyograo:
LGTM

suyograo:
Tests are still failing though

andrewvc:
Thanks for the review @suyograo I'm going to close this actually. The real fix is to merge in #390 . The problem with the specs is the async nature of them I think. So, I'm just going to nix this and keep focussing on #390

Issue #423, opened 2016-05-04
Skip document_already_exists_exception logging


gmoskovicz:
When a document already exists, and we do not want to update the content, we should use the `create` method in the elasticsearch output. This fils up the disk when a lot of documents already exists. In case someone doesn't care about this being logged, we should check the status, and the type of the exception to avoid logging this type of exceptions:

```
 {:status=>409, .................................... "error"=>{"type"=>"document_already_exists_exception", "reason"=>"[facebook][321770180859_10153425431690860]: document already exists", "shard"=>"0", "index"=>"social"}}}, :level=>:warn}
```

I propose a new configuration called `skip_error` which takes a string that is the error type (in this case `document_already_exists_exception`) so we don't log it.

gmoskovicz:
Closed by #424

SuperFunks:
This feature doesnt seem to work in logstash 5.2.2 - also it is not clear if its the statuscode you "whitelist" to blacklist it from the logoutput of if its the error type string.
Tried both - errors like these are still persisted in the log - filling up the disk.

search.transport.TransportService$4@3d82094a on EsThreadPoolExecutor[bulk, queue capacity = 50, org.elasticsearch.common.util.concurrent.EsThreadPoolExecutor@4c64817c[Running, pool size = 18, active threads = 18, queued tasks = 568, completed tasks = 47167567]]"})

gmoskovicz:
@SuperFunks may be broken in 5.2.2 , but you should be able to add the specific error, in this case `es_rejected_execution_exception` in the whitelist.

The documentation is here: https://www.elastic.co/guide/en/logstash/current/plugins-outputs-elasticsearch.html#plugins-outputs-elasticsearch-failure_type_logging_whitelist  . And the test with the example is here: https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/424/files#diff-fcc0c3ab23ddc8b47ed814f1a7b88ea4R45.

PR #424, opened 2016-05-05
Added skip error logging

gmoskovicz:
Fix for logstash-plugins/logstash-output-elasticsearch#423

### DIFF
--- a/lib/logstash/outputs/elasticsearch.rb
+++ b/lib/logstash/outputs/elasticsearch.rb
@@ -133,6 +133,11 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
   # a timeout occurs, the request will be retried.
   config :timeout, :validate => :number
 
+  # Set the Elasticsearch errors in the whitelist that you don't want to log. 
+  # A useful example is when you want to skip all 409 errors
+  # which are `document_already_exists_exception`.
+  config :failure_type_logging_whitelist, :validate => :array, :default => []
+
   # While the output tries to reuse connections efficiently we have a maximum.
   # This sets the maximum number of open connections the output will create.
   # Setting this too low may mean frequently closing / opening connections
--- a/lib/logstash/outputs/elasticsearch/common.rb
+++ b/lib/logstash/outputs/elasticsearch/common.rb
@@ -112,17 +112,18 @@ module LogStash; module Outputs; class ElasticSearch;
       actions_to_retry = []
       bulk_response["items"].each_with_index do |response,idx|
         action_type, action_props = response.first
+
         status = action_props["status"]
-        error  = action_props["error"]
+        failure  = action_props["error"]
         action = actions[idx]
 
         if SUCCESS_CODES.include?(status)
           next
         elsif RETRYABLE_CODES.include?(status)
-          @logger.info "retrying failed action with response code: #{status} (#{error})"
+          @logger.info "retrying failed action with response code: #{status} (#{failure})"
           actions_to_retry << action
-        else
-          @logger.warn "Failed action. ", status: status, action: action, response: response
+        elsif !failure_type_logging_whitelist.include?(failure["type"])
+          @logger.warn "Failed action.", status: status, action: action, response: response
         end
       end
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 4.1.2
+
+- Added a configuration called failure_type_logging_whitelist which takes a list of strings, that are error types from elasticsearch, so we prevent logging WARN if elasticsearch fails with that action. See https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/423
+
 ## 4.1.1
 - Fix bug where setting credentials would cause fatal errors. See https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/441
 
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -25,6 +25,7 @@ Contributors:
 * Tal Levy (talevy)
 * Tom Hodder (tolland)
 * jimmyjones2
+* Gabriel Moskovicz (gmoskovicz)
 
 Note: If you've sent us patches, bug reports, or otherwise contributed to
 Logstash, and you aren't on the list above and want to be, please let us know
--- a/logstash-output-elasticsearch.gemspec
+++ b/logstash-output-elasticsearch.gemspec
@@ -1,7 +1,7 @@
 Gem::Specification.new do |s|
 
   s.name            = 'logstash-output-elasticsearch'
-  s.version         = '4.1.1'
+  s.version         = '4.1.2'
   s.licenses        = ['apache-2.0']
   s.summary         = "Logstash Output to Elasticsearch"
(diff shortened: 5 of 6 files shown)


lib/logstash/outputs/elasticsearch.rb
```diff
@@ -128,6 +128,11 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
   # a timeout occurs, the request will be retried.
   config :timeout, :validate => :number
 
+  # Set the Elasticsearch errors that you don't want to log. By default we log 
+  # every error. A useful example is when you want to skip all 409 errors
+  # which are `document_already_exists_exception`
+  config :skip_errors, :validate => :array, :default  => [""]
```
jordansissel: 'skip' feels like the wrong word here. I don't have any suggestions yet for an alternative.

jordansissel: This should probably use `:default => []` right? (why have an array default to containing 1 element of an empty string?)

lib/logstash/outputs/elasticsearch/common.rb
```diff
@@ -115,7 +115,7 @@ def submit(actions)
         elsif RETRYABLE_CODES.include?(status)
           @logger.info "retrying failed action with response code: #{status} (#{error})"
           actions_to_retry << action
-        else
+        elsif not skip_errors.include?(error["type"])
```
jordansissel: it feels weird to mix "error" and the log level is 'warn'. Maybe we can rename this from 'error' to 'failure' ?

lib/logstash/outputs/elasticsearch.rb
gmoskovicz: @jordansissel what about discard_failures ?

lib/logstash/outputs/elasticsearch/common.rb
gmoskovicz: 👍

lib/logstash/outputs/elasticsearch.rb
jordansissel: Maybe something like `silence_failures => [...]` for a config setting? `silence_failures => [ 409 ]` for example?

gmoskovicz: that looks good.. however we will be skipping responses by status code rather than the elasticsearch error itself. At this point, we know that this is an answer from elasticsearch, and we have the `action_props`  object in `common.rb`. I think that we could skip ES errors than certain status codes. What do you think?

gmoskovicz: does `silence_failures => ["elasticsearch_failure_name"]` sounds good?

andrewvc:
@gmoskovicz where are we on this? Should I keep this issue open?

gmoskovicz:
@andrewvc i need to resume this.. was waiting for more input from @jordansissel

gmoskovicz:
@jordansissel mind reviewing my comments so we can push this?

lib/logstash/outputs/elasticsearch.rb
gmoskovicz: Changed to silence failures!

lib/logstash/outputs/elasticsearch.rb
```diff
@@ -133,6 +133,11 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
   # a timeout occurs, the request will be retried.
   config :timeout, :validate => :number
 
+  # Set the Elasticsearch errors that you don't want to log. By default we log 
+  # every error. A useful example is when you want to skip all 409 errors
+  # which are `document_already_exists_exception`
+  config :silence_failures, :validate => :array, :default  => [""]
```
talevy: maybe add the word `log` somewhere in this setting to make it clear what would be affected?

andrewvc: This should change to use the new typed `list` type. It should be

`config :silence_failures, :validate => :number, :list => true`

andrewvc:
Maybe we should just add an `action => create_if_not_exists` option to handle the main use case here?

lib/logstash/outputs/elasticsearch.rb
jordansissel: I somehow suspect that numbers alone aren't going to be enough for users , but maybe for the initial version of this it's ok.

talevy: I believe the type is the name of the exception in snake_case form.

e.g. `document-already_exists_exception`, so maybe

```
config :failure_type_logging_whitelist, :validate => :string, :list => true
```

whether the default should be `nil` or `[]` is unclear to me. If default were `[]` this would mean that we do not log anything by default

lib/logstash/outputs/elasticsearch/common.rb
```diff
@@ -113,15 +113,15 @@ def submit(actions)
...
         action = actions[idx]
 
         if SUCCESS_CODES.include?(status)
           next
         elsif RETRYABLE_CODES.include?(status)
-          @logger.info "retrying failed action with response code: #{status} (#{error})"
+          @logger.info "retrying failed action with response code: #{status} (#{failure})"
           actions_to_retry << action
-        else
+        elsif not silence_failures.include?(failure["type"])
```
jordansissel: use `!` instead of `not`

gmoskovicz:
I don't know if that's a good option, the actions actually match to single actions in elasticsearch. And the default for CREATE will fail if the document exists. That's why i prefer handling this at the log level rather than at a new action.

lib/logstash/outputs/elasticsearch.rb
talevy: we should also provide more information about some of the popular failure types that are returned by ES during index requests

gmoskovicz: `config :silence_failures, :validate => :number, :list => true`

I am new to this list type, shouldn't this be a list of strings? if so it's `:validate => :string`? and how do you set the defaults?

andrewvc:
@gmoskovicz yes, but that would communicate semantic intent better.

The action for this output can be dynamic, determined in an event field. We might have people wanting to silence logs on creates but not on say updates in that case.

I think this patch is fine on its own but I don't think its the right fix for the create scenario. That said, I have no problems merging it, I just see it being useful in a different way.

gmoskovicz:
i see that @andrewvc . But i also see that people can find different things to mute from Elasticsearch. Once it keeps growing, we could have different type of failures, that we want or not to be logged. Now, we either retry or log the failure. If the failure is just something that, for the use case, it's not important, then you end up filling up the logs with garbage. But open to do what you suggest 😄

lib/logstash/outputs/elasticsearch.rb
jordansissel: don't have any default, it'll make the documentation a bit cleaner, imo.

gmoskovicz:
Fixed everything!

andrewvc:
@gmoskovicz LGTM! Can you bump the minor version and add a note to the changelog? Then I can merge and release.

Thanks so much!

gmoskovicz:
Thanks for helping. @andrewvc added all the required info.

andrewvc:
LGTM

gmoskovicz:
@andrewvc squashed into a single commit

PR #426, opened 2016-05-09
Allow URI paths to work correctly when no 'path' option set.

andrewvc:

This fixes a bug where the default setting of 'path' would break all instances
of paths in 'hosts' URIs. The code would think the user had manually set 'path'.
This removes the default setting for 'path' and adds tests for these cases.

### DIFF
--- a/spec/unit/outputs/elasticsearch_spec.rb
+++ b/spec/unit/outputs/elasticsearch_spec.rb
@@ -18,10 +18,12 @@ describe "outputs/elasticsearch" do
       eso.client.send(:client).transport.options[:hosts].first
     }
 
+    let(:do_register) { true }
+
     around(:each) do |block|
-      eso.register
+      eso.register if do_register
       block.call()
-      eso.close
+      eso.close if do_register
     end
 
     describe "getting a document type" do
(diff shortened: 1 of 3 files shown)


spec/unit/outputs/elasticsearch_spec.rb
```diff
@@ -78,6 +80,55 @@
...
+          let(:options) do
+            o = super()
+            o["path"] = "/override"
+            o["hosts"] = ["http://localhost:9200/mypath/"]
+            o
+          end
+
+          it "should initialize with an error" do
+            expect { eso.register }.to raise_error(LogStash::ConfigurationError)
+          end
```
suyograo: ?What happens in this case? Does `path` override path already set in `hosts`? Can we add another rspec example for validating this?

andrewvc: It raises a ConfigurationError saying that you can't have a path in the URL and in the override.

Would would the other example check? `#register` does not complete in this case.

andrewvc: Maybe I should just add a comment?

suyograo: Ah ok, i missed that. Can you add a comment, and then update the doc above the config line as well

suyograo:
LGTM. 

Minor comment about doc update to address the case where you have path set in both `path` and `hosts`. Please update doc and merge

PR #439, opened 2016-06-17
adding time unit to documentation.

jpcarey:
https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/master/lib/logstash/outputs/elasticsearch/http_client.rb#L71-L72

>     # * `:timeout` - Float. A duration value, in seconds, after which a socket
>     #    operation or request will be aborted if not yet successfull


jordansissel:
LGTM

PR #442, opened 2016-06-24
Fix auth/proxy setup handling.

### LINKED ISSUE #441: es output causes exceptions when using credentials
Because manticore sets up auth+proxy in the same method and triggers this if either one is set (see https://github.com/cheald/manticore/blob/master/lib/manticore/client.rb#L498)

Then, if in Logstash we set credentials, manticore also tries to set the proxy (see https://github.com/cheald/manticore/blob/master/lib/manticore/client.rb#L532)

Manticore does `proxy = options.fetch(:proxy, {})`, and it receives the options from this plugin (see https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/v3.0.2/lib/logstash/outputs/elasticsearch/http_client_builder.rb#L49-L63). If there are no proxy settings passed in this plugin the options hash contains `{:proxy => nil}`, it makes `options.fetch(:proxy, {})` return `nil`.

This then causes:

```
/tmp/logstash-5.0.0-alpha4 % bin/logstash -f ../logstash-5.0.0-alpha4/my.config
--- jar coordinate com.fasterxml.jackson.core:jackson-annotations already loaded with version 2.7.1 - omit version 2.7.0
--- jar coordinate com.fasterxml.jackson.core:jackson-databind already loaded with version 2.7.1 - omit version 2.7.1-1
UNEXPECTED POOL ERROR {:e=>#<NoMethodError: undefined method `[]' for nil:NilClass>, :level=>:warn}
undefined method `[]' for nil:NilClass {:class=>"NoMethodError", :level=>:error}
Pipeline main started
UNEXPECTED POOL ERROR {:e=>#<NoMethodError: undefined method `[]' for nil:NilClass>, :level=>:warn}
An unknown error occurred sending a bulk request to Elasticsearch. We will retry indefinitely {:error_message=>"undefined method `[]' for nil:NilClass", :error_class=>"NoMethodError", :backtrace=>["/private/tmp/logstash-5.0.0-alpha4/vendor/bundle/jruby/1.9/gems/manticore-0.6.0-java/lib/manticore/client.rb:550:in `auth_from_options'", "/private/tmp/logstash-5.0.0-alpha4/vendor/bundle/jruby/1.9/gems/manticore-0.6.0-java/lib/manticore/client.rb:497:in `request_from_options'",
```

andrewvc:
We were accidentally setting `:proxy => nil` when proxy was not set by the user. Manticore just doesn't like this. This fixes this by only setting the `:proxy` key when we actually have proxy configuration data to put there.

Fixes #441

### DIFF
--- a/logstash-output-elasticsearch.gemspec
+++ b/logstash-output-elasticsearch.gemspec
@@ -1,7 +1,7 @@
 Gem::Specification.new do |s|
 
   s.name            = 'logstash-output-elasticsearch'
-  s.version         = '4.1.0'
+  s.version         = '4.1.1'
   s.licenses        = ['apache-2.0']
   s.summary         = "Logstash Output to Elasticsearch"
   s.description     = "This gem is a Logstash plugin required to be installed on top of the Logstash core pipeline using $LS_HOME/bin/logstash-plugin install gemname. This gem is not a stand-alone program"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 4.1.1
+
+- Fix bug where setting credentials would cause fatal errors. See https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/441
+
 ## 4.1.0
 - breaking,config: Removed obsolete config `host` and `port`. Please use the `hosts` config with the `[host:port]` syntax.
 - breaking,config: Removed obsolete config `index_type`. Please use `document_type` instead.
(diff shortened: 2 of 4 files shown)


logstash-output-elasticsearch.gemspec
```diff
@@ -21,7 +21,7 @@ Gem::Specification.new do |s|
 
   s.add_runtime_dependency 'stud', ['>= 0.0.17', '~> 0.0']
   s.add_runtime_dependency 'cabin', ['~> 0.6']
-  s.add_runtime_dependency "logstash-core-plugin-api", "~> 2.0"
+  s.add_runtime_dependency "logstash-core-plugin-api", [">= 2.0", "< 3.0"]
```
suyograo: Why this change here? IMO, we should do this as a mass update for all plugins. Otherwise, PH's script will be confused.

jsvd:
Without  the changes to the gemspec, LGTM

suyograo:
yep LGTM with those changes

suyograo:
Please add changelog as well.

PR #445, opened 2016-06-27
No longer explicitly support or test ES 1.x

andrewvc:
The JSON parsing is too messed update for update queries to work, and some ES1.x queries we use in tests don't work like their 5x equivalents and vice versa.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,4 @@
 ## 4.1.1
-
 - Fix bug where setting credentials would cause fatal errors. See https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/441
 
 ## 4.1.0
--- a/lib/logstash/outputs/elasticsearch.rb
+++ b/lib/logstash/outputs/elasticsearch.rb
@@ -67,7 +67,8 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
   # - delete: deletes a document by id (An id is required for this action)
   # - create: indexes a document, fails if a document by that id already exists in the index.
   # - update: updates a document by id. Update has a special case where you can upsert -- update a
-  #   document if not already present. See the `upsert` option
+  #   document if not already present. See the `upsert` option. NOTE: This does not work and is not supported
+  #   in Elasticsearch 1.x. Please upgrade to ES 2.x or greater to use this feature with Logstash!
   # - A sprintf style string to change the action based on the content of the event. The value `%{[foo]}`
   #   would use the foo field for the action
   #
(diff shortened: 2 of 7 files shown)


suyograo:
LGTM

CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
-## 4.1.1
+## 4.1.2
+- Remove ES 1.x from test matrix. The testing requirements were too severe, and update bugs were too troublesome to correct. See more about the update bugs @ https://github.com/elastic/elasticsearch/issues/19103
```
jordansissel: Maybe indicate something saying how long ES 1.x hasn't worked for this plugin, just in case someone interprets the 4.1.1 version as "Supports ES 1.x" -- I'd like to avoid some confusion :)

suyograo: good point

jordansissel: Alternately, if this update really does abandon ES 1.x, but this plugin v4.1.1 *did* support ES 1.x, we should do a major version bump to 5.0.0 of this plugin to indicate that a compatibility to prior major was lost.

andrewvc: @jordansissel Good point. I'll bump it to 5.0.0. 

I'm not sure exactly some update settings stopped working. Is it worth it to try to untangle that ball? The git commit goes back to April when I checked in the [commit](https://github.com/logstash-plugins/logstash-output-elasticsearch/commit/77c437b9ab8ddc0f3397943a9d36426bc1002ead) that made tests always green.

jordansissel:
This seems a nuclear (pardon the hyperbole) option especially for a version bump in the patch level.

If the problem is updates in ES 1.x, then can we do this instead?
* Document known issue with Elasticsearch that under ES 1.x, sending updates with Logstash won't work.
* Disable just updates for the ES 1.x test
* Update the docs to indicate fully supported versions (2.x, 5.x?)

jordansissel:
Also, as a concept, the thing bearing the version is the behavior of the plugin, in my opinion, so bumping the version for a documentation change and maybe a small acceptance test change feels .... weird.

andrewvc:
@jordansissel I'm down to just not bump the version, I'll make that test 2x and 5x only

jordansissel:
@andrewvc I'm OK with us updating the docs that say "<whatever> are known to not work on ES 1.x." -- ES 1.7.0 is only about a year old, and we tend to do 18-month support for things, don't we? It feels like we can't completely abandon ES 1.x without changing that 18-month-support practice. 

Can we do the smaller-impact and note the bug in ES 1.x and disable `update` tests against ES 1.x?

jordansissel:
User-facing stuff LGTM. Waiting on travis to complete.

PR #450, opened 2016-07-07
Improve Pool Behavior

### LINKED ISSUE #516: Error when specifying user/pass and proxy
Supplying both `user` /  `password` and `proxy` causes an error message in 5.0.

Example config:
```
input {}
filter {}
output {
  elasticsearch {
    hosts => ["192.168.1.1:9200"]
    user => "elastic"
    password => "changeme"
    proxy => "http://random:8080"
  }
}
```

Error:
```
Sending Logstash's logs to /Users/cwurm/Products/Logstash/logstash-5.0.1/logs which is now configured via log4j2.properties
[2016-11-30T09:41:27,902][INFO ][logstash.outputs.elasticsearch] Elasticsearch pool URLs updated {:changes=>{:removed=>[], :added=>["http://~hidden~:~hidden~@192.168.1.1:9200"]}}
[2016-11-30T09:41:27,907][INFO ][logstash.outputs.elasticsearch] Using mapping template from {:path=>nil}
[2016-11-30T09:41:28,029][WARN ][logstash.outputs.elasticsearch] UNEXPECTED POOL ERROR {:e=>#<TypeError: can't convert Symbol into Integer>}
[2016-11-30T09:41:28,031][ERROR][logstash.outputs.elasticsearch] Failed to install template. {:message=>"can't convert Symbol into Integer", :class=>"TypeError"}
[2016-11-30T09:41:28,031][INFO ][logstash.outputs.elasticsearch] New Elasticsearch output {:class=>"LogStash::Outputs::ElasticSearch", :hosts=>["192.168.1.1:9200"]}
[2016-11-30T09:41:28,033][INFO ][logstash.pipeline        ] Starting pipeline {"id"=>"main", "pipeline.workers"=>4, "pipeline.batch.size"=>125, "pipeline.batch.delay"=>5, "pipeline.max_inflight"=>500}
[2016-11-30T09:41:28,035][INFO ][logstash.pipeline        ] Pipeline main started
[2016-11-30T09:41:28,123][INFO ][logstash.agent           ] Successfully started Logstash API endpoint {:port=>9600}
[2016-11-30T09:41:31,058][WARN ][logstash.agent           ] stopping pipeline {:id=>"main"}
```

It's not possible to process any data.

Seems to be in the similar part of the software as https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/441

### LINKED ISSUE #517: Error when specifying proxy as hash
The [documentation states](https://www.elastic.co/guide/en/logstash/current/plugins-outputs-elasticsearch.html#plugins-outputs-elasticsearch-proxy) that it's possible to specify the proxy configuration as a hash.

However, this does not seem to be the case at the moment (5.0, but also 2.3 and probably others), at least not the way it's documented.

Example config:
```
input {}
filter {}
output {
  elasticsearch {
    hosts => ["192.168.1.1:9200"]
    user => "elastic"
    password => "changeme"
    proxy => { host: "random" port: 8080 scheme: "http"}
  }
}
```

Error message (run with `-t`):
```
Sending Logstash's logs to /Users/cwurm/Products/Logstash/logstash-5.0.1/logs which is now configured via log4j2.properties
[2016-11-30T09:45:23,775][FATAL][logstash.runner          ] The given configuration is invalid. Reason: Expected one of #, => at line 8, column 20 (byte 149) after output {
  elasticsearch {
    hosts => ["192.168.1.1:9200"]
    user => "elastic"
    password => "changeme"
    proxy => { host
```

### LINKED ISSUE #529: Resurrection doesn't take the "path" parameter into account
It seems that the resurrection feature always tries to query elasticsearch with a "HEAD /", event though a path, like "/es", is explicitly defined in the output configuration. So it doesn't work.

- Version: 5.1.1
- Operating System: Centos 7
- Config File (if you have sensitive info, please remove it):
```
 elasticsearch {
    hosts => ["myserver.domain.local:443"]
    path => "/es"
    index => "logstash-%{type}-%{+YYYY.MM.dd}"
    ssl => true
    ssl_certificate_verification => false
    template_overwrite => true
    template => "/etc/logstash/templates/elasticsearch-template.json"
    flush_size => 1000
    document_type => "%{type}"
}
```

- Sample Data:
These warning appear in the logs :
`[2016-12-15T14:54:44,388][WARN ][logstash.outputs.elasticsearch] Attempted to resurrect connection to dead ES instance, but got an error. {:url=>#<URI::HTTPS:0x1bffd71d URL:https://myserver.domain.local/es/>, :error_type=>LogStash::Outputs::ElasticSearch::HttpClient::Pool::BadResponseCodeError, :error=>"Got response code '403' contact Elasticsearch at URL 'https://myserver.domain.local/'"}`

Related reverse proxy logs:
```
172.16.x.y - - [15/Dec/2016:14:51:18 +0100] "POST /es/_bulk HTTP/1.1" 200 879
172.16.x.y - - [15/Dec/2016:14:51:18 +0100] "HEAD / HTTP/1.1" 403 -
172.16.x.y - - [15/Dec/2016:14:51:18 +0100] "POST /es/_bulk HTTP/1.1" 200 506
```

- Steps to Reproduce:
Put elasticsearch behind a reverse proxy, but not behind /.

andrewvc:
This PR does a few things to improve the Output

* This switches our URI objects to the new SafeURI object which makes not leaking users/passwords easy.
* Switches the proxy type to use the SafeURI type. This means that hashes can no longer be used as proxy configs. This is a breaking change.
* Fixes many tests that broke with the SafeURI change to be less flimsy
* Shares URL normalization logic between the pool and the http client.

Fixes #517 
Fixes #516 
Fixes #529


lib/logstash/outputs/elasticsearch/http_client.rb
```diff
@@ -27,16 +27,49 @@ class HttpClient
...
     # }
 
-    def initialize(options={})
+    def initialize(options={})      
       @logger = options[:logger]
       # Again, in case we use DEFAULT_OPTIONS in the future, uncomment this.
       # @options = DEFAULT_OPTIONS.merge(options)
       @options = options
+
+      normalize_hosts!
```
andrewvc: Generally +` on this, but we can't move much to the SafeURI constructor since it is in core. Changes there are hard to get in, and then we have cross-version issues :(

lib/logstash/outputs/elasticsearch/http_client.rb
```diff
@@ -27,16 +27,49 @@ class HttpClient
...
+    HOSTNAME_PORT_REGEX=/\A(?<hostname>([A-Za-z0-9\.\-]+)|\[[0-9A-Fa-f\:]+\])(:(?<port>\d+))?\Z/   
+    def normalize_hosts!
+      @options[:hosts] = @options[:hosts].map do |host|
+        h_string = host.to_s
+
+        if HOSTNAME_PORT_REGEX.match(h_string)
+          # Making the scheme relative with // makes the host:port parse correctly
+          ::LogStash::Util::SafeURI.new("//#{h_string}")
+        else
+          host
```
andrewvc: I think this whole block of code should probably be removed since the `SafeURI` constructor just does the right thing here. I believe the `host` variable is already a `SafeURI` at this point. Excellent catch!

lib/logstash/outputs/elasticsearch/http_client.rb
```diff
@@ -125,14 +156,85 @@ def close
     # * `:path` - String. The leading path for prefixing Elasticsearch
     #   requests. This is sometimes used if you are proxying Elasticsearch access
     #   through a special http path, such as using mod_rewrite.
-    def build_pool(options)
-      hosts = options[:hosts] || ["127.0.0.1"]
-      client_settings = options[:client_settings] || {}
-      timeout = options[:timeout] || 0
+    def calculate_property(uris, property, default, sniff_check)
```
andrewvc: argh, `git rebase!` good catch

lib/logstash/outputs/elasticsearch/http_client.rb
```diff
@@ -125,14 +156,85 @@ def close
...
+      if sniff_check && values.size > 1
+          raise LogStash::ConfigurationError, "Cannot have multiple values for #{property} in hosts when sniffing is enabled!"
+      end
+
+      uri_value = values.first
+
+      default = nil if default == '' # Blanks are as good as nil
+      uri_value = nil if uri_value == ''
+
+      if default && uri_value && (default != uri_value)
```
andrewvc: Earlier we check that `values` has a length of one, you can't mix and match the value described here. Then we check that the first (and by now we know only) value is also the same as the value in the URI.

lib/logstash/outputs/elasticsearch/http_client.rb
```diff
@@ -125,14 +156,85 @@ def close
...
+        raise LogStash::ConfigurationError, "SSL option was explicitly set to #{ssl_options[:enabled]} but a URL was also declared with a scheme of '#{explicit_scheme}'. Please reconcile this"
+      end
+
+      calculated_scheme # May be nil if explicit_scheme is nil!
+    end
+
+    def port
+      # Users can't set this, so we always set the default to nil,
+      # We have an or because we always want to return something
+      calculate_property(uris, :port, nil, sniffing) || 9200
```
andrewvc: Good catch!

lib/logstash/outputs/elasticsearch/http_client.rb
```diff
@@ -125,14 +156,85 @@ def close
...
+    end
+    
+    DEFAULT_URIS = [::LogStash::Util::SafeURI.new("http://127.0.0.1")]
+    def uris
+      @options[:hosts] || DEFAULT_URIS
```
andrewvc: I'll have to check, but I believe this makes testing much much easier. Let me confirm before I say that for sure.

lib/logstash/outputs/elasticsearch/http_client.rb
```diff
@@ -152,102 +254,55 @@ def build_pool(options)
...
-        safe_url = ::LogStash::Outputs::ElasticSearch::SafeURL.without_credentials(url)
-        raise LogStash::ConfigurationError, "A path '#{url.path}' has been explicitly specified in the url '#{safe_url}', but you also specified a path of '#{path}'. This is probably a mistake, please remove one setting."
-      end
-
-      if path
-        url.path = path  # The URI library cannot stringify if it holds a nil
-      end
-
-      if parameters
+    def host_to_url(h, parameters={})
```
andrewvc: That's a good idea. I think I actually just found a bug in your parameterized URLs patch where a sniffed host would *not* get the parameters appended. Combining these methods would remove that class of bug.

lib/logstash/outputs/elasticsearch/http_client.rb
```diff
@@ -125,14 +156,85 @@ def close
...
+
+      uri_value || default
+    end
+
+    def sniffing
+      @options[:sniffing]
+    end
+
+    def user
+      calculate_property(uris, :user, @options[:user], sniffing)
```
jsvd: just a cosmetic change, and now that I think about it, nevermind :)

lib/logstash/outputs/elasticsearch/http_client.rb
```diff
@@ -27,16 +27,49 @@ class HttpClient
     #   :setting => value
     # }
 
-    def initialize(options={})
+    def initialize(options={})
```
jsvd: unecessary extra whitespace

lib/logstash/outputs/elasticsearch/http_client/pool.rb
```diff
@@ -254,7 +253,7 @@ def normalize_url(uri)
     def update_urls(new_urls)
       # Normalize URLs
       new_urls = new_urls.map(&method(:normalize_url))
-
+
```
jsvd: unecessary extra whitespace

lib/logstash/outputs/elasticsearch/http_client.rb
andrewvc: That's a good point. That said this logic hasn't changed since the original. Can we address this in a follow-up PR? I agree that we should do something  more consistent here.

lib/logstash/outputs/elasticsearch/http_client.rb
andrewvc: Well, apparently its there for no reasons. I removed it and no test failures :)

lib/logstash/outputs/elasticsearch/common.rb
```diff
@@ -212,8 +212,7 @@ def safe_bulk(actions)
         retry unless @stopping.true?
       rescue ::LogStash::Outputs::ElasticSearch::HttpClient::Pool::BadResponseCodeError => e
         if RETRYABLE_CODES.include?(e.response_code)
-          safe_url = ::LogStash::Outputs::ElasticSearch::SafeURL.without_credentials(e.url)
-          log_hash = {:code => e.response_code, :url => safe_url}
+          log_hash = {:code => e.response_code, :url => e.url}
```
jsvd: can we maybe log the url.to_s or similar? otherwise unecessary object ids leak to the logs (visual garbage):

```
[2016-12-15T12:06:43,688][ERROR][logstash.outputs.elasticsearch] Attempted to send a bulk request to elasticsearch but received a bad HTTP response code! {:code=>503, :url=>#<URI::HTTP:0x4189e152 URL:http://127.0.0.1:9202/_bulk>}
```

jsvd:
I'm not sure I understand how the connection pool is being used to index documents. I started a 3 node cluster of ES and the following logstash config on my mbp dual core+HT:

```
input { generator {} }
output {
  elasticsearch {
    hosts => ["localhost:9200"]
    sniffing => true
    pool_max => 50
    pool_max_per_route => 50
  }
}
```

And I get 6 connections indexing documents:

```
% nettop -m tcp -n -c -t loopback -p 11168 -l 1 -jinterface,state,bytes_in,bytes_out -ktime,rx_dupe,rx_ooo,re-tx,tc_class,tx_win,tc_mgt,cc_algo,P,C,R,W
                                                              interface         state        bytes_in       bytes_out   rtt_avg   rcvsize
java.11168                                                                                     16 MiB         578 MiB
   tcp4 127.0.0.1:58554<->127.0.0.1:9200                            lo0     CloseWait        6643 B           188 KiB   0.22 ms   392 KiB
   tcp4 127.0.0.1:9600<->*:*                                        lo0        Listen
   tcp4 127.0.0.1:58557<->127.0.0.1:9200                            lo0     CloseWait          18 KiB         582 KiB   0.22 ms   380 KiB
   tcp4 127.0.0.1:58558<->127.0.0.1:9200                            lo0     CloseWait          32 KiB        1025 KiB   0.22 ms   366 KiB
   tcp4 127.0.0.1:58559<->127.0.0.1:9200                            lo0     CloseWait          46 KiB        1499 KiB   0.22 ms   351 KiB
   tcp4 127.0.0.1:61442<->127.0.0.1:9202                            lo0   Established        2818 KiB          96 MiB   0.22 ms   387 KiB
   tcp4 127.0.0.1:61443<->127.0.0.1:9200                            lo0   Established        2815 KiB          96 MiB   0.22 ms   390 KiB
   tcp4 127.0.0.1:61444<->127.0.0.1:9201                            lo0   Established        2771 KiB          95 MiB   0.22 ms   434 KiB
   tcp4 127.0.0.1:61450<->127.0.0.1:9202                            lo0   Established        2769 KiB          95 MiB   0.22 ms   436 KiB
   tcp4 127.0.0.1:61451<->127.0.0.1:9201                            lo0   Established        2744 KiB          94 MiB   0.22 ms   461 KiB
   tcp4 127.0.0.1:61452<->127.0.0.1:9200                            lo0   Established        2798 KiB          96 MiB   0.22 ms   407 KiB
```

I see 6 connections active in the pool. Is this because for each node we do indexing and sniffing, so we end up with 2 connections per node?

andrewvc:
@jsvd Yes, that's it exactly. The sniffing happens concurrently and in a round robin fashion. You should expect, under load, to see the # of connections stabilize somewhere near NUM_WORKERS+NUM_NODES connections. This is obviously not a deterministic process.

BTW, the tests are failing after rebasing onto master, I'll try and fix that today.

andrewvc:
@jsvd Actually, I got the formula wrong. The # connections is hard to calculate because latency matters. We try to minimize concurrent connections, yet also have the same # of connections to all nodes in ES so we can round robin. So for a three node cluster you would typically see values of 3 connections 6 connections, or 9 connections. But nothing *not* divisible by 3.

andrewvc:
@jsvd the calculate_property thing with port actually was done for a reason. I added a better comment to the code to indicate this. The previous comment was confusing.

andrewvc:
@jsvd ok, fixed all the comments AFAICT. Ready for another review :)

jsvd:
I'm not sure what is happening but, in my tests, the credentials (user/pass) are dropped from the request, resulting in 401 unauthorized.

andrewvc:
@jsvd I did mess with the adapter class. That is likely the cause! Will look later today.

lib/logstash/outputs/elasticsearch/http_client.rb
```diff
@@ -105,34 +138,86 @@ def close
...
+    def user
+      calculate_property(uris, :user, @options[:user], sniffing)
+    end
+
+    def password
+      calculate_property(uris, :password, @options[:password], sniffing)
+    end
+
+    def path
+      calculated = calculate_property(uris, :path, client_settings[:path], sniffing)
```
jsvd: I'd rather call the result of this "path_string" or "path", makes:

`calculated = "/#{calculated}" if calculated && !calculated.start_with?("/")`

more eligible:

`path.prepend('/') if path && !path.start_with?("/")`

lib/logstash/outputs/elasticsearch/http_client.rb
```diff
@@ -105,34 +138,86 @@ def close
...
+
+      if calculated_scheme && explicit_scheme && calculated_scheme != explicit_scheme
+        raise LogStash::ConfigurationError, "SSL option was explicitly set to #{ssl_options[:enabled]} but a URL was also declared with a scheme of '#{explicit_scheme}'. Please reconcile this"
+      end
 
+      calculated_scheme # May be nil if explicit_scheme is nil!
+    end
+
+    def port
+      # We don't set the 'default' here because the default is what the user
```
jsvd: 👍

lib/logstash/outputs/elasticsearch/http_client.rb
```diff
@@ -105,34 +138,86 @@ def close
...
+    def uris
+      @options[:hosts]
+    end
+
+    def client_settings
+      @options[:client_settings] || {}
+    end
+
+    def ssl_options
+      @options[:client_settings] ? @options[:client_settings][:ssl] : {}
```
jsvd: since we define

```ruby
def client_settings
  @options[:client_settings] || {}
end
```

this means we can do:

```ruby
def ssl_options
  client_settings.fetch(:ssl, {})
  # or
  client_settings[:ssl] || {}
end
```

lib/logstash/outputs/elasticsearch/http_client/pool.rb
```diff
@@ -228,26 +224,25 @@ def start_resurrectionist
...
         end
       end
     end
 
     def stop_resurrectionist
-      @resurrectionist.join
+      @resurrectionist.join if @resurrectionist
     end
 
     def resurrectionist_alive?
```
jsvd: this method is never used, show we remove it?

lib/logstash/outputs/elasticsearch/http_client/pool.rb
```diff
@@ -270,26 +265,19 @@ def perform_request_to_url(url, method, path, params={}, body=nil)
...
+      u
     end
 
     def update_urls(new_urls)
       return if new_urls.nil?
       
       # Normalize URLs
       new_urls = new_urls.map(&method(:normalize_url))
-
+
```
jsvd: unnecessary white space

jsvd:
functionality-wise it LGTM, just pointed out a few aesthetic suggestions

lib/logstash/outputs/elasticsearch/http_client.rb
andrewvc: Good call!

lib/logstash/outputs/elasticsearch/http_client.rb
andrewvc: That would actually break in some situations.
`calculate_property` can return nils, hence the nil check. Additionally, `prepend` should be named `prepend!` because it mutates whatever is passed into it :(.

Some improvement may be possible, but I'm not so concerned with it.

jsvd:
LGTM

andrewvc:
Added a final commit here to fix proxy issues in #517 and #516 . I added some unit tests and manually tested against an nginx forward proxy.

andrewvc:
This is the forward proxy config I used.

lib/logstash/outputs/elasticsearch/http_client/manticore_adapter.rb
```diff
@@ -31,10 +46,26 @@ def client
...
-      # Convert URI object to string
-      url_and_path = path ? (url + path).to_s  : url.to_s
+      
+      # Create a new SafeURI that we can modify
+      request_uri = ::LogStash::Util::SafeURI.new(url.uri.clone)
+      
+      # Combine the paths using the minimal # of /s
+      # First, we make sure the path is relative so URI.join does
+      # the right thing
+      relative_path = path.start_with?("/") ? path[1..-1] : path
```
jsvd: we don't enforce path to be non null. somewhere up in the stack (with `healthcheck!`) this is causing exceptions to occur in the test run

andrewvc: Ah, damn. Good catch, will fix!

andrewvc:
Fixed the failing specs

lib/logstash/outputs/elasticsearch/http_client/manticore_adapter.rb
```diff
@@ -31,10 +46,28 @@ def client
...
+      # We excise auth info from the URL in case manticore itself tries to stick
+      # sensitive data in a thrown exception or log data
+      if request_uri.user
+        params[:auth] = { :user => request_uri.user, :password => request_uri.password, :eager => true }
+        request_uri.user = nil
+        request_uri.password = nil
+      end
 
-      resp = @manticore.send(method.downcase, url_and_path, params)
+      #require 'pry'; binding.pry
```
jsvd: leftover debug statement

lib/logstash/outputs/elasticsearch/http_client/manticore_adapter.rb
```diff
@@ -45,13 +78,12 @@ def perform_request(url, method, path, params={}, body=nil)
...
       if resp.code < 200 || resp.code > 299 && resp.code != 404
-        safe_url = ::LogStash::Outputs::ElasticSearch::SafeURL.without_credentials(url)
-        raise ::LogStash::Outputs::ElasticSearch::HttpClient::Pool::BadResponseCodeError.new(resp.code, safe_url + path, resp.body)
+        raise ::LogStash::Outputs::ElasticSearch::HttpClient::Pool::BadResponseCodeError.new(resp.code, request_uri, body)
       end
 
       resp
     end
-
+
```
jsvd: unnecessary whitespace

andrewvc:
@jsvd just fixed the whitespace / debug things you found.

jsvd:
LGTM

BasemMohamed:
Is the following error related to this issue?!

`[2017-01-29T16:14:07,481][INFO ][logstash.pipeline        ] Starting pipeline {"id"=>"main", "pipeline.workers"=>32, "pipeline.batch.size"=>125, "pipeline.batch.delay"=>5, "pipeline.max_inflight"=>4000}
[2017-01-29T16:14:07,512][INFO ][logstash.pipeline        ] Pipeline main started
[2017-01-29T16:14:07,532][INFO ][logstash.agent           ] Successfully started Logstash API endpoint {:port=>9600}
[2017-01-29T16:14:12,472][INFO ][logstash.outputs.elasticsearch] Running health check to see if an Elasticsearch connection is working {:url=>#<URI::HTTP:0x417a2945 URL:http://localhost:9200>, :healthcheck_path=>"/"}
[2017-01-29T16:14:12,478][WARN ][logstash.outputs.elasticsearch] Attempted to resurrect connection to dead ES instance, but got an error. {:url=>#<URI::HTTP:0x417a2945 URL:http://localhost:9200>, :error_type=>LogStash::Outputs::ElasticSearch::HttpClient::Pool::BadResponseCodeError, :error=>"Got response code '401' contact Elasticsearch at URL 'http://localhost:9200/'"}
[2017-01-29T16:14:17,479][INFO ][logstash.outputs.elasticsearch] Running health check to see if an Elasticsearch connection is working {:url=>#<URI::HTTP:0x7babbe64 URL:http://localhost:9200>, :healthcheck_path=>"/"}
[2017-01-29T16:14:17,482][WARN ][logstash.outputs.elasticsearch] Attempted to resurrect connection to dead ES instance, but got an error. {:url=>#<URI::HTTP:0x7babbe64 URL:http://localhost:9200>, :error_type=>LogStash::Outputs::ElasticSearch::HttpClient::Pool::BadResponseCodeError, :error=>"Got response code '401' contact Elasticsearch at URL 'http://localhost:9200/'"}
[2017-01-29T16:14:22,483][INFO ][logstash.outputs.elasticsearch] Running health check to see if an Elasticsearch connection is working {:url=>#<URI::HTTP:0xa1d47b8 URL:http://localhost:9200>, :healthcheck_path=>"/"}
`

andrewvc:
That error means that log stash does not have permission to GET / on the es
server. The health check is failing. You must fix your elastic search
permissions

Issue #401, opened 2016-03-28
Use correct templates based on Elasticsearch versions


suyograo:
ES 5.0 has numerous mapping changes as described here https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/386. Much effort has been put into making this a backward compatible change recently, but we'd still like users to migrate to the new mapping settings and not carry the old cruft around. 

As such, we should:
* Ship both ES 2.x (old) and ES 5.0 (new) templates
* Enhance this plugin to detect if ES 5.0 is running (when the first attempt to upload template is made), and if so, use the 5.0 template version. If ES version 2.x is running, use the other template. 

**Note: We should do this only when default template is used and the existing template* configs should apply to both the versions**

We may need to introduce a new config option to select different paths for 5.0 and 2.x templates on disk.

pickypg:
I came back to this issue to double check that this part was here:

> We should do this only when default template is used.

I just wanted to call it out for whoever implements it because otherwise it will break some things. :)

suyograo:
Thanks @pickypg . Made it stick out :)

tsg:
What does "force upgrade" mean? Overwrite it?

suyograo:
@tsg when I mean't force upgrade, I actually meant forcing users to upgrade to a 5.0 template if they are running against ES 5.0. If they already have a `logstash` template, we wouldn't overwrite it.

acchen97:
After discussion with @suyograo and @jordansissel we plan to backport this logic for 2.4 as well. Therefore, both LS 2.4 and 5.0 will have compatible default templates with both ES 2.x and 5.0.

suyograo:
@jsvd moving this to @purbon

suyograo:
@jsvd will start on this. 

For ref: beats did something similar as well: https://github.com/elastic/beats/pull/1993

suyograo:
Fixed in #455

PR #455, opened 2016-07-20
choose default template based on es version

jsvd:
during template installation, query ES to check the version and if the default template is to be used, choose which one based on ES major version

TODO:
- [x] have different templates for 2x and 5x

fixes https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/401


suyograo:
Whats up with the integration tests?

jsvd:
Missing the templates :)

suyograo:
@jsvd here's the default template for 2.x: https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/b85e72e8c160100a84a6ebc341fda88f58de0999/lib/logstash/outputs/elasticsearch/elasticsearch-template.json

For 5.x you can copy what we have now in master.

jsvd:
@suyograo I think we need an es 1.x default template too 😆 https://travis-ci.org/logstash-plugins/logstash-output-elasticsearch/builds/146423290

untergeek:
LGTM™ and tests pass for me locally.

untergeek:
Tests pass for me locally, but I notice that the Travis CI tests are not passing, and that’s because the template testing is ​_only_​ for ES major versions 2 and 5: https://travis-ci.org/logstash-plugins/logstash-output-elasticsearch/jobs/146423305

do we want to remove ES 1.7.x testing from Travis?

suyograo:
we should use the 2.x template in 1.x as well. That has worked in the past. 1.x is still popular, so I'd like to continue testing against it

untergeek:
Will do another quick PR to add in support for v1 to use the v2 template.

PR #457, opened 2016-07-26
ES 1.x gets the ES 2.x template

### LINKED ISSUE #456: Add support for ES 1.x to use the ES 2.x template
This is what's currently being done anyway.

untergeek:
This is the current default behavior anyway

fixes #456

### DIFF
--- a/lib/logstash/outputs/elasticsearch/template_manager.rb
+++ b/lib/logstash/outputs/elasticsearch/template_manager.rb
@@ -30,7 +30,8 @@ module LogStash; module Outputs; class ElasticSearch
     end
 
     def self.default_template_path(es_major_version)
-      default_template_name = "elasticsearch-template-es#{es_major_version}x.json"
+      es_major_version == "1" ? template_version = "2" : template_version = es_major_version
+      default_template_name = "elasticsearch-template-es#{template_version}x.json"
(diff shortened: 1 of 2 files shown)


lib/logstash/outputs/elasticsearch/template_manager.rb
```diff
@@ -30,7 +30,8 @@ def self.install(client, template_name, template, template_overwrite)
     end
 
     def self.default_template_path(es_major_version)
-      default_template_name = "elasticsearch-template-es#{es_major_version}x.json"
+      es_major_version == "1" ? template_version = "2" : template_version = es_major_version
```
ph: I think you meant this

```ruby
template_version = es_major_version == "1" ?  "2" :  es_major_version
```

I was a bit confused with the assignment in the ternary ops.

ph:
LGTM, I havent tested with an actual ES 1.7 but I am sure it resolve #456

untergeek:
Thanks for fixing the ternary operator.  I'm still trying to cross over my python domain knowledge into Ruby... 😮

ph:
> Thanks for fixing the ternary operator. I'm still trying to cross over my python domain knowledge into Ruby... 😮

Don't worry don't tell anyone I use semicolon in ruby now because of java..

ph:
LGTM

PR #459, opened 2016-07-26
let ES 5.x handle unanalyzed dynamically mapped strings

talevy:
All string fields in ES (as of https://github.com/elastic/elasticsearch/pull/17188) will have a special dynamic mapping defined. They will be associated with the `text` type and have an unanalyzed multi-field `keyword`.

Historically, Logstash has used the `.raw` multi-field for a similar mapping strategy. Now that this is the default in ES, there is no reason to stray away from the ES defaults.

This PR attempts to keep consistency between the new string-types and default multi-field naming.


suyograo:
Closed in favor of #462

PR #462, opened 2016-08-04
Update template to use ES 5.x mapping

### LINKED ISSUE #386: Mapping changes in Elasticsearch 5.0 for string field type

suyograo:
Fixes #386


untergeek:
LGTM™.  Let's see if the tests pass 😄

untergeek:
Super excited about this, btw.  Definitely going to tweak some upgraders, but +1 for the clean start.

suyograo:
hmm, I have to use some rspec tag trickery to fix this.

jordansissel:
Let's block this until https://github.com/elastic/elasticsearch/issues/19812 is resolved.

suyograo:
Aye. Stalling

jordansissel:
A few of us (@suyograo, @andrewvc, @talevy, etc) had discussions about this offline, and there were two main areas for discussion about the `keyword` name:

1. The `keyword` name in this template was chosen to be consistent with Elasticsearch's use of the word. Further, the name `keyword` is maybe not the best word.
2. The change from `raw` to `keyword` may negatively impact users who upgrade LS from 2.x to 5.x.

FOr the first part, we are discussing in https://github.com/elastic/elasticsearch/issues/19812.

For the second part, we might leave this field named as `raw` to relieve stress on the upgrade experience. We may move to be consistent with Elasticsearch at a later date after we figure out how to minimize pains of upgrades.

untergeek:
In all honesty, I'd multi-field and have both raw and keyword.  This forces users to adjust the template themselves if they're space conscious, and gives us both reverse and forward compatibility, should Kibana standardize on `.keyword` in future releases.

suyograo:
> In all honesty, I'd multi-field and have both raw and keyword. 

I am -1 on this. I wouldn't want users to incur extra disk space because of this. This multi-field is for every string, which is quite a bit.

jordansissel:
Did some historical digging. Logstash v1.3.0 (December 2013) is the first version which shipped a template. Changing `.raw` has the opportunity to break schema with possibly 2.5 years of events stored in Elasticsearch with Logstash.

We discussed this last week somewhere I forget, but we had some thoughts for Logstash 5.0 timeframe:

Premise:

1.By default,  Logstash will only install the template if there is no template already there.
2. We want to ship a template that more closely aligns with Elasticsearch's naming and defaults.
3. Elasticsearch now makes JSON strings dynamically mapped as both `text` and a multifield `.keyword` as keyword mapping type.

Therefore, I summarize a proposal as follows:

1. We change the default template to change `.raw` to `.keyword`
2. Advertise loudly in the changelog and release notes about this change.
3. Document an upgrade path for existing 2.x users to upgrade to the new template, including instructions on how to reindex data to effect a rename of `.raw` to `.keyword`

suyograo:
@jordansissel I've changed this PR to reflect the proposal in https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/462#issuecomment-238376557.  Can you take a look please? /cc @talevy

.travis.yml
```diff
@@ -3,9 +3,9 @@ sudo: false
 jdk: oraclejdk8
 env:
 - INTEGRATION=false
-- INTEGRATION=true ES_VERSION=2.2.0
 - INTEGRATION=true ES_VERSION=1.7.5
-- INTEGRATION=true ES_VERSION=5.0.0-alpha1
+- INTEGRATION=true ES_VERSION=2.3.4
+- INTEGRATION=true ES_VERSION=5.0.0-alpha4
```
untergeek: Shouldn't this be alpha5 now?

talevy: bump version to `alpha5`?

suyograo: Will do

jsvd: This should be increased to alpha5 since it has been released.

untergeek:
LGTM™

.travis.yml
untergeek: Hahahaha.  All of us saw the same thing 👍

lib/logstash/outputs/elasticsearch/elasticsearch-template-es5x.json
```diff
@@ -20,24 +20,23 @@
           "match" : "*",
           "match_mapping_type" : "string",
           "mapping" : {
-            "type" : "string", "index" : "analyzed", "omit_norms" : true,
-            "fielddata" : { "format" : "disabled" },
+            "type" : "text", "norms" : false,
             "fields" : {
-              "raw" : {"type": "string", "index" : "not_analyzed", "ignore_above" : 256}
+              "keyword" : { "type": "keyword" }
```
talevy: I thought the idea was to let ES run through its defaults in the case that a stringed field is being dynamically mapped

spec/integration/outputs/pipeline_spec.rb
```diff
@@ -1,6 +1,6 @@
 require_relative "../../../spec/es_spec_helper"
 
-describe "Ingest pipeline execution behavior", :integration => true, :version_5x => true do
+describe "Ingest pipeline execution behavior", :integration => true, :version_greater_than_5x => true do
```
talevy: `greater_than`? and equal to?

lib/logstash/outputs/elasticsearch/elasticsearch-template-es5x.json
suyograo: Yes, but we need `"norms" : false,` to be set, which ES does not by default. That's why I had to explicitly add this rule for strings.

suyograo: See 2 lines above.

spec/integration/outputs/pipeline_spec.rb
suyograo: good call

lib/logstash/outputs/elasticsearch/elasticsearch-template-es5x.json
talevy: ah, I see. sounds good

talevy:
lgtm

PR #463, opened 2016-08-07
Added config param build_update_action to control update_action_builder

sxmichael:
This is necessary when the input is already in es_bulk format, to allow just passing the content through without the need to parse all the special cases at input (doc_as_upsert, script, params) and re-creating it.


karmi:
Hi @sxmichael, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/463.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

untergeek:
For this change, it is be important to add Rspec tests to validate the new argument, and that the output is as expected.  Is this something you can do, @sxmichael, or do you need assistance?

sxmichael:
@untergeek just to clarify - do you mean to add rspec test to validate the cases when passed build_update_action=false? If yes, I can do it. Else, could you please clarify which tests do you mean?

untergeek:
@sxmichael That's exactly what I mean.  Add tests for the changes you've made.

sxmichael:
@untergeek added test, please check. Currently tests are failing not-related to my commit (I think so).

sxmichael:
@untergeek I've added the test. PLease let me know if it's ok. Thanks

sxmichael:
@untergeek @karmi any update?

andrewvc:
I don't believe it is a goal for this output to support the native ES output bulk format as input. Why do you need this? Why not just talk directly to elasticsearch without logstash?

sxmichael:
Hi @andrewvc , sorry for the delay. The basic idea here is to be able to replace long time ago deprecated river funtionality. We have application writing it's data in bulk api format to rabbitmq and wanted to leverage logstash to ship the data to es. There are several benefits here, the most major ones are ability to withstand the data spikes, no need to implement retry logic on application side (in case of version conflicts or request throttling) and ability to build bigger bulks (application could output data in item by item or small chunks and give logstash build bigger bulk messages and send to es).

PR #464, opened 2016-08-16
Remove empty event action params

sxmichael:
In event_action_params() clean params from nil and empty values - when using dynamic fields (like document_id => "%{[@metadata][document_id]}"), empty or nil values could be produced for document_id and parent fields (in es_bulk metadata part), which is not valid es_bulk metadata format, thus those values should be cleaned.

I've added remove_empty_action_params configuration option to control this behavior (default is false).
Test is added as well.


sxmichael:
@untergeek @karmi any update?

spec/unit/outputs/elasticsearch_spec.rb
```diff
@@ -262,6 +262,53 @@
...
+
+    context "when remove_empty_action_params enabled" do
+      let(:options) { super.merge("remove_empty_action_params" => true) }
+
+      it "should interpolate the requested id and parent values" do
+        params = eso.event_action_params(event_full)
+        expect(params).to include(:_id => "mytestid", :parent => "mytestparent")
+      end
+
+      it "should interpolate the requested id and parent values and leave them as is (sprintf behavior, not sure if correct one) when they missing" do
```
andrewvc: This behavior isn't great, but I don't know what else we can do here. I think this should be documented in the option docs.

spec/unit/outputs/elasticsearch_spec.rb
```diff
@@ -262,6 +262,53 @@
...
+        params = eso.event_action_params(event_full)
+        expect(params).to include({:_id => "mytestid", :parent => "mytestparent"})
+      end
+
+      it "should interpolate the requested id and parent values and leave them as is (sprintf behavior, not sure if correct one) when they missing" do
+        params = eso.event_action_params(event_partial_missing)
+        expect(params).to include(:_id => "%{myid}", :parent => "%{myparent}")
+      end
+
+      it "should interpolate the requested id and parent values and leave them as is (sprintf behavior, not sure if correct one) when they empty" do
```
andrewvc: This just repeats the previous test. I think you wanted to test that this should interpolate the requested id and parent values.

sxmichael: Yeas, those are two identical tests, just to see the actual difference in behavior what happens when proposed remove_empty_action_params param set to true and to false.

sxmichael:
Hey @andrewvc could this be merged?

PR #471, opened 2016-08-29
Change default timeout from ∞ to 60 seconds

### LINKED ISSUE #470: Set default timeout to 1 minute
Currently we do not have a timeout for HTTP connections by default. This should be set to 1 minute for a sane value in the odd case where connections are dropped on the receiving end but never receives a FIN.

andrewvc:
Fixes #470


untergeek:
LGTM™

(that was a short change!)

PR #472, opened 2016-08-29
Add stale connection check / Bump to 5.1.0

### LINKED ISSUE #468: Add Stale Connection Check
Sometimes people have bad networks that drop idle connections. For these people we should add the stale connection check as we do here: https://github.com/logstash-plugins/logstash-mixin-http_client/blob/master/lib/logstash/plugin_mixins/http_client.rb#L106

andrewvc:
Add 'check_connection_timeout' parameter and bump to 5.1.0

This also releases the new timeout defaults

Fixes #468

### DIFF
--- a/lib/logstash/outputs/elasticsearch.rb
+++ b/lib/logstash/outputs/elasticsearch.rb
@@ -177,7 +177,10 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
   # How long to wait before checking if the connection is stale before executing a request on a connection using keepalive.
   # You may want to set this lower, if you get connection errors regularly
   # Quoting the Apache commons docs (this client is based Apache Commmons):
-  # 'Defines period of inactivity in milliseconds after which persistent connections must be re-validated prior to being leased to the consumer. Non-positive value passed to this method disables connection validation. This check helps detect connections that have become stale (half-closed) while kept inactive in the pool.'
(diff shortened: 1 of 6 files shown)


lib/logstash/outputs/elasticsearch.rb
```diff
@@ -173,6 +173,13 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
   # Resurrection is the process by which backend endpoints marked 'down' are checked
   # to see if they have come back to life
   config :resurrect_delay, :validate => :number, :default => 5
+  
+  # How long to wait before checking if the connection is stale before executing a request on a connection using keepalive.
+  # You may want to set this lower, if you get connection errors regularly
+  # Quoting the Apache commons docs (this client is based Apache Commmons):
+  # 'Defines period of inactivity in milliseconds after which persistent connections must be re-validated prior to being leased to the consumer. Non-positive value passed to this method disables connection validation. This check helps detect connections that have become stale (half-closed) while kept inactive in the pool.'
```
untergeek: You may want to put some line breaks in here to keep from having to scroll left/right in the code.

untergeek:
LGTM™, other than having to side scroll a lot to read the documentation

andrewvc:
@untergeek thanks! I just updated the docs to have more breaks, I'll push after tests pass 😎

PR #474, opened 2016-09-08
revert to use `.raw` instead of `.keyword`

### LINKED ISSUE #466: Continue to use `.raw` instead of `.keyword`
Hi all

We discussed the issue of renaming `.raw` to `.keyword` fields in the Elasticsearch FixItFriday again today.  I thought that field aliases (https://github.com/elastic/elasticsearch/issues/17511) might possibly be a solution to help with the transition.  We even considered adding a dedicated hack to automatically check for a `.raw` field if a `.keyword` field didn't exist.

Unfortunately we didn't come up with a good solution.  

We're aware of the impact that this rename will have on our users and want to suggest that you reconsider the change in https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/462

In Elasticsearch, naming the field `.keyword` makes sense given our variety of use cases, but Logstash solves a much more focused problem.  A `.raw` field HAS made sense in the past and will continue to make sense in Logstash's future.  There's no real reason to change - I think it is OK to deviate from the Elasticsearch default here.

It's certainly an easy way of removing a large obstacle from the upgrade path.

/cc @tsg @monicasarbu

suyograo:
In PR https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/462#issuecomment-238376557
we used `.keyword` to match ES's change in default behavior.

Both approaches, using `.keyword` to match ES for consistency and using `.raw` to match
existing logstash data has pros/cons. I think at this time it is wise to err on the side of
easy-upgrade-path for users so reverting template back to using `.raw` as before

Fixes #466


jordansissel:
I think @clintongormley's notes about the impact on various user groups are good reference here: https://github.com/elastic/elasticsearch/issues/18195#issuecomment-238841499

I believe existing users (with or without default settings) will be unaffected by the .keyword change until they elect to change from .raw to .keyword. New users will start with .keyword. This seems safe-enough to me.

I hope moving from .raw to .keyword is not a painful transition. I am confident that we understand the impact to each group of users. 

As discussed offline, I"ll close this. <3

T3rm1:
I have this problem now. Upgraded from 5 to 6 and with the new default template `.keyword` is used instead of `.raw`. So every new index uses keyword while tons of old indices use raw. What do you suggest to do in that case?

andrewvc:
@T3rm1 have you considered using a custom template? That way you can control the field names exactly.

PR #475, opened 2016-09-13
Hide user and password from the url.

jordansissel:
Previously, a 403 error would log this:

> Got response code '403' contact Elasticsrearch at URL 'http://foo:bar@localhost:9200/'

This caused a leak in credentials.


suyograo:
LGTM

PR #478, opened 2016-09-15
Add "version" to 5.x template

### LINKED ISSUE #477: Add "version" to 5.x template
Templates (and pipelines) now support a `"version"` integer field that can added to them for the type of need that Logstash has: when to replace the existing template. (https://github.com/elastic/elasticsearch/issues/20171)

We should add the `"version"` to the existing Logstash template, which should allow us to simplify replacing it, as needed, in the future as well as the 2.x release.

```http
GET /_template/logstash?filter_path=*.version
```

This will return an empty JSON object for non-versioned templates.

```json
{}
```

and it will return whatever version is set for templates that do set it:

```json
{"logstash":{"version":20160915}}
```

pickypg:
This adds a version value of today's date (`20160915`) as the version. Any change to the template should increase the version number (change to that date), thus making it easier to detect changes.

As this is the first release of 5.x, it really is somewhat moot, but any future release may also choose to build in variable for the 5.x change so that the file only needs to be loaded _if_ the version actually needs to be replaced.

Closes #477

### DIFF
--- a/lib/logstash/outputs/elasticsearch/elasticsearch-template-es5x.json
+++ b/lib/logstash/outputs/elasticsearch/elasticsearch-template-es5x.json
@@ -1,5 +1,6 @@
 {
   "template" : "logstash-*",
+  "version" : 20160915,
   "settings" : {
     "index.refresh_interval" : "5s"
   },
--- a/lib/logstash/outputs/elasticsearch/elasticsearch-template-es5x.json
+++ b/lib/logstash/outputs/elasticsearch/elasticsearch-template-es5x.json
@@ -1,6 +1,6 @@
 {
   "template" : "logstash-*",
-  "version" : 20160915,
+  "version" : 50001,
   "settings" : {
     "index.refresh_interval" : "5s"
   },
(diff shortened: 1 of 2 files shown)


jordansissel:
Should we include the ES major version in the template version?

My question comes from the fact that with Logstash we try hard to support older versions of Elasticsearch. In Logstash 5.0.0, we will support Elasticsearch 2.0 and 5.0 with different templates. For Logstash 6.0, I can see us having two templates (ES 5 and ES 6) also that are maintained independently. And dates alone will not necessarily be indicative of how to prioritize one template over another.

Thoughts?

pickypg:
Seems reasonable to me. Logstash is in somewhat of a unique situation here.

I would think that the version can be safely prefixed, but this creates an INT_MAX issue if we want to include the minor version as well: `5x20160915` where `x` is the minor release (`0` in the first release of course). As the Logstash plugin should be expected to work with 5.0+, perhaps we shouldn't worry about minor releases... or we can fall back to `YY` instead of `YYYY`?

jordansissel:
I think if we version the template per version of ES, and do something like ES.TEMPLATEVERSION, but since the `version` field is an Integer, use 0 instead of a period, and end up with something like this:


* 50001 -- ES 5.x, template version 1 (zero padded `001`).
* 20004 -- ES 2.x template version 4 (zero padded `004`).

Thoughts?

pickypg:
Love it. Version cannot be added until 5.x though, so older variants don't get the benefit unfortunately.

suyograo:
LGTM

PR #480, opened 2016-09-21
use path_match instead of match in the message_field field template

jsvd:
this prevents fields like "message.error" or "other.message" to also not get a .raw field.
By using path_match, only "message" will not have a .raw


jordansissel:
Lgtm.

talevy:
accidentally made duplicate https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/481.

jsvd:
Closed in favor of https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/481

PR #481, opened 2016-09-21
update templates to be more strict on `message` match

talevy:
Currently, all fields named `message` are mapped to be excluded
from the `.keyword` and `.raw`. This is rather strict, and need only
be applied to the top-level `message` field. All nested `**.*.message`
fields should be treated like any old field.

### DIFF
--- a/lib/logstash/outputs/elasticsearch/elasticsearch-template-es5x.json
+++ b/lib/logstash/outputs/elasticsearch/elasticsearch-template-es5x.json
@@ -12,8 +12,8 @@
           "path_match" : "message",
           "match_mapping_type" : "string",
           "mapping" : {
-            "type" : "string", "index" : "analyzed", "norms" : false,
-            "fielddata" : { "format" : "disabled" }
+            "type" : "text",
+            "norms" : false
           }
         }
       }, {
--- a/lib/logstash/outputs/elasticsearch/elasticsearch-template-es5x.json
+++ b/lib/logstash/outputs/elasticsearch/elasticsearch-template-es5x.json
@@ -9,7 +9,7 @@
       "_all" : {"enabled" : true, "norms" : false},
       "dynamic_templates" : [ {
         "message_field" : {
-          "match" : "message",
+          "path_match" : "message",
           "match_mapping_type" : "string",
           "mapping" : {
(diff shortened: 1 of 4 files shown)


lib/logstash/outputs/elasticsearch/elasticsearch-template-es5x.json
```diff
@@ -9,7 +9,7 @@
       "_all" : {"enabled" : true, "norms" : false},
       "dynamic_templates" : [ {
         "message_field" : {
-          "match" : "message",
+          "path_match" : "message",
           "match_mapping_type" : "string",
           "mapping" : {
             "type" : "string", "index" : "analyzed", "norms" : false,
```
suyograo: @talevy I missed this when I was doing #402 -- 

`"type" : "string", "index" : "analyzed", "norms" : false,`

should be

`"type" : "text", "index" : "analyzed", "norms" : false,`

`text` not `string` for 5.0

suyograo: Also, can you remove this line https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/master/lib/logstash/outputs/elasticsearch/elasticsearch-template-es5x.json#L16

jordansissel:
Duplicate of https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/480, but this one has specs.

talevy:
😞 didn't realize a PR was already out for this. sorry jwow. I'm fine with closing this in favor of the other

lib/logstash/outputs/elasticsearch/elasticsearch-template-es5x.json
talevy: re: your first comment wrt `text` vs `string`. I will change it.

from the documentation, I was kind of shocked this didn't fail https://www.elastic.co/guide/en/elasticsearch/reference/5.0/string.html.

After looking at the code, looks like ES optimistically upgrades this into a `text` mapping since we also provide `"index": "analyzed"`. Still, I can update

suyograo:
Sorry @jsvd this one has tests and fixes some other stuff :)

suyograo:
LGTM

jsvd:
LGTM!

PR #483, opened 2016-09-27
Hide credentials in exceptions and log messages.

### LINKED ISSUE #482: Passwords in logstash logs partially visible
I'm using logstash version 5.0.0 beta 1. I have a output configuration listed as:

> output {
> elasticsearch {
> hosts => ["xxx.xxx.local:9200"]
> index => "land-%{+YYYY.MM}"
> user => "elastic"
> password => "changeme"
> }
> }

For some reason its failing to connect to ES v5.0.0 beta 1 (not saying this is a bug, I'm investigating this separately) however I spotted in the logs for logstash I can see that the password has been hidden in some places but not others. Unclear if that is intentional or not? See below:

`[2016-09-22T22:42:31,953][WARN ][logstash.outputs.elasticsearch] Marking url as dead. Last error: [LogStash::Outputs::ElasticSearch::HttpClient::Pool::HostUnreachableError] Elasticsearch Unreachable: [http://~hidden~:~hidden~@xxx.xxx.local:9200][Manticore::SocketTimeout] Read timed out {:url=>#<URI::HTTP:0x5eecb836 URL:http://elastic:changeme@xxx.xxx.local:9200>, :error_message=>"Elasticsearch Unreachable: [http://~hidden~:~hidden~@xxx.xxx.local:9200][Manticore::SocketTimeout] Read timed out", :error_class=>"LogStash::Outputs::ElasticSearch::HttpClient::Pool::HostUnreachableError"}`

jordansissel:
Fixes #482

### DIFF
--- a/lib/logstash/outputs/elasticsearch/common.rb
+++ b/lib/logstash/outputs/elasticsearch/common.rb
@@ -208,7 +208,8 @@ module LogStash; module Outputs; class ElasticSearch;
         retry unless @stopping.true?
       rescue ::LogStash::Outputs::ElasticSearch::HttpClient::Pool::BadResponseCodeError => e
         if RETRYABLE_CODES.include?(e.response_code)
-          log_hash = {:code => e.response_code, :url => e.url}
+          safe_url = ::LogStash::Outputs::ElasticSearch::SafeURL.without_credentials(e.url)
+          log_hash = {:code => e.response_code, :url => safe_url}
           log_hash[:body] = e.body if @logger.debug? # Generally this is too verbose
           @logger.error("Attempted to send a bulk request to elasticsearch but received a bad HTTP response code!", log_hash)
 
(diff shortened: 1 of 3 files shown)


andrewvc:
This is weird. The SafeURI class was added to logstash core so that we would't need to do stuff like use the SafeURL class. The idea was to make it secure by default.

I wonder if this object has been converted to a regular URI class. by accident at some point?

andrewvc:
[SafeURI](https://github.com/elastic/logstash/blob/master/logstash-core/lib/logstash/util/safe_uri.rb)

andrewvc:
I'll add that I'd prefer to figure out what happened here and have it be secure by default rather than force us to try to capture every logging use case of a URI which is more fragile as code is changed.

andrewvc:
Oh, I guess I never finished / merged this PR https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/450, WDYT about this approach @jordansissel ? I like the implicit safety of it.

suyograo:
@andrewvc Can we go with @jordansissel approach for 5.0 RC1 and then fix it for a point release using your PR #450 ? This will fix #482 and we can take time to finish up #450

andrewvc:
@suyograo #450 is very close, the main issue was dependency errors in LS core causing unit tests to fail.

andrewvc:
@suyograo hmmm, nm that, I'm OK to go with @jordansissel 's solution for now.

andrewvc:
LGTM

PR #486, opened 2016-10-03
Remove docker dependency

suyograo:
`longeshoreman` depends on `docker-api` which now wants to use ruby 2.0. We don't use longeshoreman anymore in our tests, so removing all together.


untergeek:
That clears a few crufty hacks out, now doesn't it?

LGTM™

PR #489, opened 2016-10-04
Breaking: Pass tests for ES 5.0.0-beta1

### LINKED ISSUE #488: Scripted upserts don't work anymore in Elasticsearch 5.0.0-beta1
@suyog has details, please update this issue ;P

suyograo:
Breaking change to set lang to painless by default mirroring ES
Groovy related tests still pass

Removed one test in create_spec because it does not make sense
as ES is mandating passing of id with op_type=create

For the scripting tests, passing groovy manually to make tests work

Fixes #488

### DIFF
--- a/spec/integration/outputs/update_spec.rb
+++ b/spec/integration/outputs/update_spec.rb
@@ -9,7 +9,8 @@ describe "Update actions", :integration => true, :version_greater_than_equal_to_
       "index" => "logstash-update",
       "template_overwrite" => true,
       "hosts" => get_host_port(),
-      "action" => "update"
+      "action" => "update",
+      "script_lang" => "groovy"
     }
     LogStash::Outputs::ElasticSearch.new(settings.merge!(options))
   end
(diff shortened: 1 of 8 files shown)


spec/integration/outputs/update_spec.rb
```diff
@@ -9,7 +9,8 @@ def get_es_output( options={} )
       "index" => "logstash-update",
       "template_overwrite" => true,
       "hosts" => get_host_port(),
-      "action" => "update"
+      "action" => "update",
+      "script_lang" => "groovy"
```
untergeek: So we're using a groovy script here, because it works with older versions of Elasticsearch too?  It'd be nice to get a "painless" script eventually, I think, but it's not a deal breaker now.

untergeek:
LGTM™.  I _would_ like to see a painless script added in the future, of course.

spec/integration/outputs/update_spec.rb
suyograo: Yes, at some point, we should move these tests to painless.

Issue #453, opened 2016-07-13
Deprecate and noop-ize flush_size parameter


andrewvc:
With the newer pipelines that support batch sizes this parameter is basically useless. We should just deprecate it and eventually delete it.

jordansissel:
Deprecation of flush_size appears in https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/497

Issue #382, opened 2016-03-03
Deprecate flush_size and use pipeline batch size as ES bulk request size


suyograo:
There are 2 different knobs to tune batching in Logstash:
* The pipeline batch size which was introduced in 2.2 that controls the overall event batching for *all* filters and outputs.
* Additionally, in ES output we have `flush_size` which controls the bulk request size to Elasticsearch.

Proposal is to deprecate `flush_size` and just use pipeline batch size. It is just confusing to have to tune batching in multiple places. This also plays well with future acking proposal for resiliency. There has been asks for flush size based on bytes, but IMO, this size based buffering should be also at the pipeline level.

A big drawback to this approach is the one-size-fits-all for buffering events in outputs, which may not work well for certain outputs like HDFS, S3, etc. 

Moved this discussion from https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/381#discussion_r54793561

Thoughts?

jsvd:
Relevant discussion on the ES side: https://github.com/elastic/elasticsearch/issues/16011

jordansissel:
Alternate solution is provided by  https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/497

dlgasser:
#497 doesn't seem to be a fix for users of AWS ElasticSearch, which caps bulk requests at 10Mb.

jordansissel:
@dlgasser Does this plugin even work with Amazon Elasticsearch Service? I think Amazon forked this repository a long time ago and are maintaining their own plugin specific to Amazon's Elasticsearch service.

dlgasser:
Yes, it still works if your IAM setup doesn't require connection requests to be signed (i.e. you use network or EC2 role-based access control for the ElasticSearch service)

jordansissel:
I am hesitant to make this tunable. However, @dlgasser can you open a separate issue to discuss your specific issue? (10mb bulk cap)

RishikeshDarandale:
@jordansissel 

> I am hesitant to make this tunable. However, @dlgasser can you open a separate issue to discuss your specific issue? (10mb bulk cap)

I see that [#785](https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/785) issue is open against above.

PR #497, opened 2016-10-26
Break up large bulk requests into multiple requests.

### LINKED ISSUE #164: flush operation can lead to OOM due to huge memory overhead
When it's time for a flush (by either reaching flush_size, flush_interval or teardown) the memory cost of setting up the bulk operation can lead to OOM.

For example, using the following config:

`input { stdin { } } output { elasticsearch { protocol => "http" host => "127.0.0.1" flush_size => 10000 index => "ugh" } }`

And piping in 300.000 messages of 200-400 bytes each will cause logstash to crash:

```
Logstash startup completed
Exception in thread ">output" java.lang.UnsupportedOperationException
    at java.lang.Thread.stop(Thread.java:869)
    at org.jruby.RubyThread.exceptionRaised(RubyThread.java:1221)
    at org.jruby.internal.runtime.RubyRunnable.run(RubyRunnable.java:112)
    at java.lang.Thread.run(Thread.java:745)
```

NOTE: logstash 1.4.2 was able to terminate the task under the same conditions.

### LINKED ISSUE #284: Buffer based on bytes, not doc sizes
This moves the second half of https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/181 here. We should have a buffer sized definable by byte size, not buffer size.

andrewvc:
This PR changes the HTTP Client to try as best as it can to restrict bulk requests to 20MB or less. If a single document is larger than this it will be sent alone in a separate request.

This PR also changes the flush_size to be effectively infinite. If users have explicitly set the `flush_size` it will still work until that option is made obsolete.

Fixes #164, #284, #382, #453, #499

### DIFF
--- a/lib/logstash/outputs/elasticsearch.rb
+++ b/lib/logstash/outputs/elasticsearch.rb
@@ -17,20 +17,20 @@ require "uri" # for escaping user input
 # yet far easier to administer and work with. When using the HTTP protocol one may upgrade Elasticsearch versions without having
 # to upgrade Logstash in lock-step. For those still wishing to use the node or transport protocols please see
 # the <<plugins-outputs-elasticsearch_java,elasticsearch_java output plugin>>.
-#
+# 
 # You can learn more about Elasticsearch at <https://www.elastic.co/products/elasticsearch>
 #
 # ==== Template management for Elasticsearch 5.x
-# Index template for this version (Logstash 5.0) has been changed to reflect Elasticsearch's mapping changes in version 5.0. 
-# Most importantly, the subfield for string multi-fields has changed from `.raw` to `.keyword` to match ES default 
+# Index template for this version (Logstash 5.0) has been changed to reflect Elasticsearch's mapping changes in version 5.0.
+# Most importantly, the subfield for string multi-fields has changed from `.raw` to `.keyword` to match ES default
 # behavior.
 #
 # ** Users installing ES 5.x and LS 5.x **
 # This change will not affect you and you will continue to use the ES defaults.
 #
 # ** Users upgrading from LS 2.x to LS 5.x with ES 5.x **
-# LS will not force upgrade the template, if `logstash` template already exists. This means you will still use 
-# `.raw` for sub-fields coming from 2.x. If you choose to use the new template, you will have to reindex your data after 
+# LS will not force upgrade the template, if `logstash` template already exists. This means you will still use
+# `.raw` for sub-fields coming from 2.x. If you choose to use the new template, you will have to reindex your data after
 # the new template is installed.
 #
 # ==== Retry Policy
@@ -47,6 +47,10 @@ require "uri" # for escaping user input
 #
 # NOTE: 409 exceptions are no longer retried. Please set a higher `retry_on_conflict` value if you experience 409 exceptions.
 # It is more performant for Elasticsearch to retry these exceptions than this plugin.
+#
+# ==== Batch Sizes ====
+# This plugin attempts to send batches of events as a single request. However, if
+# a request exceeds 20MB we will break it up until multiple batch requests. If a single document exceeds 20MB it will be sent as a single request.
 #
 # ==== DNS Caching
 #
@@ -146,7 +150,7 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
   # a timeout occurs, the request will be retried.
   config :timeout, :validate => :number, :default => 60
 
-  # Set the Elasticsearch errors in the whitelist that you don't want to log. 
+  # Set the Elasticsearch errors in the whitelist that you don't want to log.
   # A useful example is when you want to skip all 409 errors
   # which are `document_already_exists_exception`.
   config :failure_type_logging_whitelist, :validate => :array, :default => []
@@ -173,7 +177,7 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
   # Resurrection is the process by which backend endpoints marked 'down' are checked
   # to see if they have come back to life
   config :resurrect_delay, :validate => :number, :default => 5
-  
+
   # How long to wait before checking if the connection is stale before executing a request on a connection using keepalive.
   # You may want to set this lower, if you get connection errors regularly
   # Quoting the Apache commons docs (this client is based Apache Commmons):
--- a/lib/logstash/outputs/elasticsearch/common_configs.rb
+++ b/lib/logstash/outputs/elasticsearch/common_configs.rb
@@ -81,7 +81,7 @@ module LogStash; module Outputs; class ElasticSearch
       # If you specify a number larger than the batch size of your pipeline it will have no effect,
       # save for the case where a filter increases the size of an inflight batch by outputting
       # events.
-      mod.config :flush_size, :validate => :number, :default => 500
+      mod.config :flush_size, :validate => :number, :deprecate => "This setting is no longer necessary as we now try to restrict bulk requests to sane sizes. See the 'Batch Sizes' section of the docs. If you think you still need to restrict payloads based on the number, not size, of events, please open a ticket."
 
       # The amount of time since last flush before a flush is forced.
       #
--- a/lib/logstash/outputs/elasticsearch/http_client.rb
+++ b/lib/logstash/outputs/elasticsearch/http_client.rb
@@ -5,6 +5,21 @@ require 'logstash/outputs/elasticsearch/http_client/pool'
 require 'logstash/outputs/elasticsearch/http_client/manticore_adapter'
 
 module LogStash; module Outputs; class ElasticSearch;
+  # This is a constant instead of a config option because
+  # there really isn't a good reason to configure it.
+  #
+  # The criteria used are:
+  # 1. We need a number that's less than 100MiB because ES
+  #    won't accept bulks larger than that.
+  # 2. It must be large enough to amortize the connection constant
+  #    across multiple requests.
+  # 3. It must be small enough that even if multiple threads hit this size
+  #    we won't use a lot of heap.
+  #
+  # We wound up agreeing that a number greater than 10 MiB and less than 100MiB
+  # made sense. We picked one on the lowish side to not use too much heap.
+  TARGET_BULK_BYTES = 20 * 1024 * 1024 # 20MiB
+
   class HttpClient
     attr_reader :client, :options, :logger, :pool, :action_count, :recv_count
     # This is here in case we use DEFAULT_OPTIONS in the future
@@ -38,9 +53,11 @@ module LogStash; module Outputs; class ElasticSearch;
     def bulk(actions)
       @action_count ||= 0
       @action_count += actions.size
-      
+
       return if actions.empty?
-      bulk_body = actions.collect do |action, args, source|
+
+
+      bulk_actions = actions.collect do |action, args, source|
         args, source = update_action_builder(args, source) if action == 'update'
 
         if source && action != 'delete'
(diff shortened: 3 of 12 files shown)


andrewvc:
I'm thinking 75MB may be too high, but I set it high because if we set it too low we'll reject legit documents that are large. Maybe we need two options `max_bulk_megabytes` for the maximum request and `target_bulk_megabytes` for the size to attempt to hit. We would still send if the bulk size winds up being larger than `target_bulk_megabytes` and only drop / log if the doc is larger than `max_bulk_megabytes`. Thoughts?

lib/logstash/outputs/elasticsearch.rb
```diff
@@ -184,6 +184,16 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
...
 
+  # The target size of an HTTP request in megabytes to send. Elasticsearch has a hard limit of 100MB
+  # after which it will reject the request. If the batch received by this plugin exceeds this amount
+  # it will be split up and sent in multiple requests. A request may exceed this size if there is a single
+  # document larger than it. You can limit an individuals document's size with max_doc_megabytes
+  config :target_bulk_megabytes, :validate => :number, :default => 75
+
+  # The maximum size an individual document can be without being dropped and logged
+  # Elasticsearch will not accept HTTP requests > 100MB. You likely do not want to increase this size
+  config :max_doc_megabytes, :validate => :number, :default => 90
```
jordansissel: I feel like we don't' need this setting. I think we can have the code know the max size that will be rejected by Elasticsearch and just use that internally.

Can you tell me a story of why a user would want to set this?

lib/logstash/outputs/elasticsearch.rb
```diff
@@ -184,6 +184,16 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
   # See https://hc.apache.org/httpcomponents-client-ga/httpclient/apidocs/org/apache/http/impl/conn/PoolingHttpClientConnectionManager.html#setValidateAfterInactivity(int)[these docs for more info]
   config :validate_after_inactivity, :validate => :number, :default => 10000
 
+  # The target size of an HTTP request in megabytes to send. Elasticsearch has a hard limit of 100MB
+  # after which it will reject the request. If the batch received by this plugin exceeds this amount
+  # it will be split up and sent in multiple requests. A request may exceed this size if there is a single
+  # document larger than it. You can limit an individuals document's size with max_doc_megabytes
+  config :target_bulk_megabytes, :validate => :number, :default => 75
```
jordansissel: I greatly prefer base units for settings, so `bytes` instead of `megabytes`.

I'll send you a PR to add bytes to your branch.

lib/logstash/outputs/elasticsearch.rb
jordansissel: https://github.com/andrewvc/logstash-output-elasticsearch/pull/3 moves to use bytes instead of number.

lib/logstash/outputs/elasticsearch/http_client.rb
```diff
@@ -38,23 +40,45 @@ def get_version
...
+                    LogStash::Json.dump(action)
+
+        if as_json.size > @max_doc_bytes
+          @logger.warn("Document too large for ES output! Dropping",
+                        :max_doc_megabytes => @options[:max_doc_megabytes],
+                        :bulk_action_size => as_json.size)
+          next
+        end
+
+        if (bulk_body.size + as_json.size) > @target_bulk_bytes
```
suyograo: Do we have to call `bulk_body.size` every time? Maybe we memoize this size, for the in-flight `bulk_batch`? As in, keep a running total of the current batch size. You only need to calculate size for current doc every time, and then add it to `bulk_batch_size`. Then reset `bulk_batch_size` when the request is complete or when an error happens. WDYT?

andrewvc: This is basically memoized in the internal impl.

lib/logstash/outputs/elasticsearch/common_configs.rb
```diff
@@ -75,13 +75,9 @@ def self.included(mod)
...
       mod.config :hosts, :validate => :array, :default => ["127.0.0.1"]
 
-      # This plugin uses the bulk index API for improved indexing performance.
-      # This setting defines the maximum sized bulk request Logstash will make
-      # You you may want to increase this to be in line with your pipeline's batch size.
-      # If you specify a number larger than the batch size of your pipeline it will have no effect,
-      # save for the case where a filter increases the size of an inflight batch by outputting
-      # events.
-      mod.config :flush_size, :validate => :number, :default => 500
+      # DEPRECATED. We now restrict bulk sizes to a target size of 20MB. Larger events
```
jordansissel: This might scare users.

Maybe say "We now automatically send the best-sized bulk request where the size is chosen to minimize network latency and memory pressure. Internally, this means we will split a bulk request into multiple requests if they would exceed 20MB in size." 

Main goal was to express the intent and also to avoid the word "restrict"

lib/logstash/outputs/elasticsearch/http_client.rb
```diff
@@ -38,23 +40,52 @@ def get_version
...
-      flatten.
-      reduce("") do |acc,line|
-        acc << LogStash::Json.dump(line)
-        acc << "\n"
       end
 
+      bulk_body = ""
+      bulk_responses = []
+      bulk_actions.each do |action|
+        as_json = action.is_a?(Array) ?
```
jordansissel: when would action be an array?

lib/logstash/outputs/elasticsearch/http_client.rb
```diff
@@ -38,23 +40,52 @@ def get_version
...
+
+    def join_bulk_responses(bulk_responses)
+      {
+        "errors" => flatten_one_array(bulk_responses("errors")),
+        "items" => flatten_one_array(bulk_responses("items"))
+      }
+    end
+
+    # Flatten array by one level only
+    def flatten_one_array(arr)
```
jordansissel: missing code?

lib/logstash/outputs/elasticsearch/http_client.rb
```diff
@@ -38,23 +40,52 @@ def get_version
...
       end
 
+      bulk_body = ""
+      bulk_responses = []
+      bulk_actions.each do |action|
+        as_json = action.is_a?(Array) ?
+                    action.map {|line| LogStash::Json.dump(line)}.join("\n") :
+                    LogStash::Json.dump(action)
+
+        if (bulk_body.size + as_json.size) > TARGET_BULK_BYTES
```
ph: should it be bytesize here?

jordansissel: Good catch. My brain is still in ruby 1.8 where utf-8 strings weren't common. +1 to .bytesize

ph: @jordansissel Well, It help when your language make things breaks :)

lib/logstash/outputs/elasticsearch/http_client.rb
andrewvc: @jordansissel this is for *extra* flat arrays... ;)

jordansissel: So flat it's zero-dimensional! Computers are cool!

jordansissel:
@andrewvc  I just scanned a few old issues. Can you make your commits say that they fix #164, #284, #382, #453, #499

jordansissel:
Also https://github.com/elastic/logstash/issues/4000

lib/logstash/outputs/elasticsearch/common_configs.rb
```diff
@@ -75,13 +75,8 @@ def self.included(mod)
...
 
-      # This plugin uses the bulk index API for improved indexing performance.
-      # This setting defines the maximum sized bulk request Logstash will make
-      # You you may want to increase this to be in line with your pipeline's batch size.
-      # If you specify a number larger than the batch size of your pipeline it will have no effect,
-      # save for the case where a filter increases the size of an inflight batch by outputting
-      # events.
-      mod.config :flush_size, :validate => :number, :default => 500
+      # DEPRECATED. We now restrict bulk sizes to a size that balances heap size with reducing round-trips. Very large events will be sent in a single request by themselves.
+      mod.config :flush_size, :validate => :number, :default => 500, :deprecated => "No Longer Needed"
```
andrewvc: You make a good point, but I think this is still a good use of deprecate. In this case we don't want to break configs, but we want people to stop explicitly setting this. We could leave this feature in, but that's bad because it does things no one should want to do.

I'm open to hearing a conflicting opinion here of course, but there's no good reason to set `flush_size` once this PR lands IMHO.

jsvd: I agree with everything but this being a good use of deprecate.
Tagging an option as deprecated and making it a noop can be very dangerous (remember this setting is non default very frequently). I don't feel comfortable in assuming flush_size is useless for everyone and simply swapping the flush heuristic from count to size will be a win for everyone.
Is for this lack of certainty that we have the deprecate -> obsolete cycle, so we can test the waters, provide a fall back for unexpected scenarios and also a smoother migration.

andrewvc: OK, well, I'm fine deprecating it and preserving old behavior for now. But I would like to remove this option before 6.0.

jordansissel: @jsvd my original intent with `deprecate` was, at the configuration interface, that a configuration would still be loaded successfully but we would warn a user about the deprecation. With obsolete, my intent was to fail to load the config but tell the user why.

From this, I think making it :deprecated as a no-op is fine, since the config still runs.

You have a good point about the folks setting flush_size previously, though, but to be honest, all the uses I can think of are basically workarounds for perceived for actual problems with how much we send to Elasticsearch in a single request. This patch hopefully will remove those problems (perceived or actual) with Elasticsearch bulk sizes, and in doing so, will remove the need for that band-aid -- therefore, I think it's OK to immediately turn :flush_size into a no-op.

Thoughts? Am I missing a key reason that isn't a work-around for why someone would set flush_size?

jordansissel: @jsvd in your favor, the most minimal user-experience change would be to mark it deprecated, set the default to nil, but keep the existing behavior if someone sets flush_size directly.

jsvd: +1 I'm glad like to see flush_size go away, but pretty much all production cases I've seen tune that setting in some way, which is why I'm concerned with removing it.

lib/logstash/outputs/elasticsearch/http_client.rb
andrewvc: `update_action_builder` returns either a a single action or an array of them (some bulk things take up two lines, like indexing). Before this was solved via a call to `.flatten`

lib/logstash/outputs/elasticsearch/common_configs.rb
```diff
@@ -81,7 +81,7 @@ def self.included(mod)
       # If you specify a number larger than the batch size of your pipeline it will have no effect,
       # save for the case where a filter increases the size of an inflight batch by outputting
       # events.
-      mod.config :flush_size, :validate => :number, :default => 500
+      mod.config :flush_size, :validate => :number, :default => 99999999, :deprecate => "This setting is no longer necessary as we now try to restrict bulk requests to sane sizes. See the 'Batch Sizes' section of the docs. If you think you still need to restrict payloads based on the number, not size, of events, please open a ticket."
```
jordansissel: can we default to nil? A value of 9999999 looks weird (will show up in the docs) and confuses me, and may confuse users.

lib/logstash/outputs/elasticsearch/http_client.rb
```diff
@@ -5,6 +5,8 @@
 require 'logstash/outputs/elasticsearch/http_client/manticore_adapter'
 
 module LogStash; module Outputs; class ElasticSearch;
+  TARGET_BULK_BYTES = 20 * 1024 * 1024
```
jordansissel: Put a comment here something maybe explaining why we made this a constant instead of a tunable? My thought is that I want to avoid having users yelling at us about a "bad decision" making this not configurable (even though we agreed to make this not a configuration setting)

lib/logstash/outputs/elasticsearch/common_configs.rb
andrewvc: Yeah, good point!

lib/logstash/outputs/elasticsearch/http_client.rb
andrewvc: Will add a comment!

lib/logstash/outputs/elasticsearch/common_configs.rb
```diff
@@ -81,7 +81,7 @@ def self.included(mod)
       # If you specify a number larger than the batch size of your pipeline it will have no effect,
       # save for the case where a filter increases the size of an inflight batch by outputting
       # events.
-      mod.config :flush_size, :validate => :number, :default => 500
+      mod.config :flush_size, :validate => :number, :deprecate => "This setting is no longer necessary as we now try to restrict bulk requests to sane sizes. See the 'Batch Sizes' section of the docs. If you think you still need to restrict payloads based on the number, not size, of events, please open a ticket."
```
jordansissel: Typo, `:deprecate` should be `:deprecated`

reference: https://github.com/elastic/logstash/blob/master/logstash-core/lib/logstash/config/mixin.rb#L66

jordansissel:
Left one comment. Once this typo is fixed, you have my lgtm! :)

PR #501, opened 2016-11-07
Depends on Addressable 2.3.X instead of 2.4.0

ph:
Some development dependencies of the core require to use 2.3.X and wont
work with 2.4


suyograo:
LGTM

PR #502, opened 2016-11-10
Use .bytesize not .size for bulk size checks

andrewvc:
This makes things more accurate when multi-byte chars are in the source.


ph:
LGTM, merge when ci passes.

PR #508, opened 2016-11-14
Add `sprintf` or event dependent configuration when specifying ingest pipeline

suyograo:
Previously, ingest pipeline config was not setup to use
event dependent configuration, but now it is.


andrewvc:
LGTM

PR #511, opened 2016-11-22
Remove mention of elasticsearch_java output plugin 

dedemorton:
elasticsearch_java output is no longer supported in 5.0, so I'm removing it from the doc.


andrewvc:
LGTM. Thanks @dedemorton

PR #513, opened 2016-11-22
Perform healthchecks at startup using state machine

### LINKED ISSUE #507: Healthcheck should be performed at startup, not only on error
Currently healthchecks are only performed on nodes that are dead before they are resurrected. This means that a user can deploy logstash, have it function, then have a node become temporarily unavailable only to discover that it won't be detected as up by logstash because the healthcheck URL is not one Logstash has perms to access.

All nodes should start in an 'unknown' state and transition to 'good' to make this behavior more consistent. Additionally, we should provide a helpful error in the situation where the healthcheck URL fails.

andrewvc:
Refactored the pool.rb URL status to be more state-machine like.
URLs start as 'unknown' then after a healtcheck transition to 'alive'.

This was the cleanest way to fix #507

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 5.4.0
+- Perform healthcheck against hosts right after startup / sniffing
+
 ## 5.3.5
 - Docs: Remove mention of using the elasticsearch_java output plugin because it is no longer supported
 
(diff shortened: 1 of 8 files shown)


jordansissel:
LGTM

jordansissel:
(haven't tested)

andrewvc:
@jordansissel thanks! I ran tests locally, works great :)

andrewvc:
@jordansissel can you review the latest commit I've pushed?

suyograo:
LGTM for 3fdb686

Issue #518, opened 2016-12-02
Don't include sprintf'ed fields if the operation fails


jsvd:
currently this plugin can be configured to pull certain information from the event like document_id, index or document_type. This is implemented using `event.sprintf("%{[@metadata][id]})` for the document_id for example.

If this sprintf fails, the key is still created in the event with the raw string value of `%{[@metadata][id]}`. this created problems as documents will be indexed in elasticsearch with incorrect data.

My suggestion is that we add an extra condition after each sprintf and only modify the event if the sprintf succeeds.

/cc @andrewvc

jsvd:
closing in favor of https://github.com/elastic/logstash/issues/6363

PR #519, opened 2016-12-05
ensure event interpolation does not pollute event data

jsvd:
eventually we could change Event#sprintf so it never leaks a field reference, but that can break existing setups, so I'm including a safer (slower) sprintf that checks if the result of sprintf contains a field reference, and returns `nil`

solves #518


talevy:
I think this change should be done in Event#sprintf, maybe an optional argument for a type of "ignore" flag. what do you think?

jsvd:
I like that idea,

```ruby
class Event
  def sprintf(format, strict=false)
    # ...
  end
end
```

jsvd:
closing in favor of https://github.com/elastic/logstash/issues/6363

PR #523, opened 2016-12-06
Always retry full-request failures except 400.

### LINKED ISSUE #522: Non-retryable response codes from ES will be retried indefinitely
When executing a bulk request, if the server responds with a non-retryable response code the request will in-fact be retried indefinitely due to a bug where `error.respose.body` is invoked, where it should be `error.body` in . In some sense this is actually desirable because bulk should always return 200. See https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/master/lib/logstash/outputs/elasticsearch/common.rb#L223 for where the error occurs.

andrewvc:
It makes no sense to stop retrying a bulk request because the bulk API
should always return 200. In any other situation we may lose data.

400 response codes are still dropped as that usually indicates a
too-large request.

Fixes #522


andrewvc:
Alternate approach. Maybe we *don't* retry 400s, but instead detect the large requests AOT.

jordansissel:
If we assume 400 could mean too-large, we could just retry with a smaller request on 400s?

(Detecting too-large requires we know the maximum size of a request body in Elasticsearch before we send the request, which may not be practical to obtain? I'm open to whatever)

andrewvc:
@jordansissel this only happens if there is a single event larger than the limit. We already attempt to split up requests in this case, but we obviously cannot subdivide a single event.

andrewvc:
The more I think about it the more I think we need a new option `max_request_size`. We just don't send any request larger than that, and that way we don't need a 'special' 400 case.

In the absence of the too-large-request scenario we *should* retry 400s indefinitely because that would indicate either:

1. A bug in the output code
2. A bug in the user infrastructure, a bad proxy say

In either case they wouldn't want to start dropping events in those cases. A stall would be appropriate.

jordansissel:
> In the absence of the too-large-request scenario we should retry 400s indefinitely

++

andrewvc:
@jordansissel OK, so I've changed the logic and updated the description per our discussion. Can I get a review from yourself?

andrewvc:
@jordansissel is there a plugin-api change I need to make now that we're using the `bytes` validation?

lib/logstash/outputs/elasticsearch.rb
```diff
@@ -88,7 +93,7 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
   # - A sprintf style string to change the action based on the content of the event. The value `%{[foo]}`
   #   would use the foo field for the action
   #
-  # For more details on actions, check out the http://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html[Elasticsearch bulk API documentation]
+  # For more details on actions, check out the http://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html[Elasticsearch bthey are:ulk API documentation]
```
jordansissel: typo inserting 'they are:' in the middle of the word 'bulk' ?

lib/logstash/outputs/elasticsearch.rb
```diff
@@ -191,6 +196,28 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
   # have become stale (half-closed) while kept inactive in the pool.'
   # See https://hc.apache.org/httpcomponents-client-ga/httpclient/apidocs/org/apache/http/impl/conn/PoolingHttpClientConnectionManager.html#setValidateAfterInactivity(int)[these docs for more info]
   config :validate_after_inactivity, :validate => :number, :default => 10000
+  
+  # This is the maximum size HTTP request this plugin will attempt to send.
+  #
+  # This plugin will try to send appropriately sized requests, using multiple
+  # requests per batch if the events are very large, but if a single event
+  # is larger than this value sending even that single request will break.
```
jordansissel: "will break" --- it will break what?

lib/logstash/outputs/elasticsearch.rb
```diff
@@ -191,6 +196,28 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
...
+  # 
+  # We check this size after serializing to see whether we should even try to send it
+  # If the event is larger than this we log an error and drop the event.
+  # 
+  # The default is set to work with Elasticsearch's default maximum request size of 100MiB
+  # Note that we've set it by default to 1MiB lower as a fudge factor for any protocol overhead
+  # or other things that could account for a slight difference in the HTTTP request, as our size
+  # check only happens on the request body.
+  #
+  # This is specified as a number now, but we should move it to the new 'bytes' type
```
jordansissel: "This is specified ..." will show up in the docs. I recommend not putting this comment here.

lib/logstash/outputs/elasticsearch.rb
andrewvc: Ah, clearly a complete ViM failure :)

lib/logstash/outputs/elasticsearch.rb
andrewvc: I'll put in a note saying that we will change it in the future.

andrewvc:
@jordansissel can we move this forward to LGTM?

lib/logstash/outputs/elasticsearch.rb
```diff
@@ -191,6 +196,30 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
...
+  # 
+  # We check this size after serializing to see whether we should even try to send it
+  # If the event is larger than this we log an error and drop the event.
+  # 
+  # The default is set to work with Elasticsearch's default maximum request size of 100MiB
+  # Note that we've set it by default to 1MiB lower as a fudge factor for any protocol overhead
+  # or other things that could account for a slight difference in the HTTTP request, as our size
+  # check only happens on the request body.
+  #
+  # This is specified as a number now, but we will be moved to the new 'bytes' type
```
jordansissel: I think this is a TODO item and we don't want ti showing up in the docs (which it would, in its current location in a comment above this)

jordansissel:
Once the comment-that-looks-like-a-todo-item thing is removed, LGTM :)

andrewvc:
I for the life of me can't repro these travis failures. They ONLY happen on 5.1.1, and they definitely don't happen locally for me :(. The lack of SSH into travis makes this even more perplexing.

I tried some debugging last week, and it appeared that ES wasn't accepting TCP connections anymore on travis! 

Bad travis, bad!

@jsvd @jordansissel any ideas here?

andrewvc:
The failure can be seen here: https://travis-ci.org/logstash-plugins/logstash-output-elasticsearch/jobs/185633956

PR #526, opened 2016-12-08
Fix sniffing on 5x ES servers

### LINKED ISSUE #525: 5x Sniffing Broken
Our sniffing code does not work with ES5.x. Our tests were based on 5.0.0.alpha-2 and used a fixture from that version. We should fix this plugin to sniff ES5.x correctly.

andrewvc:
Previously, sniffing was tested against ES 5.0.0-alpha2, which used the
same 'nodes' response as ES 2.x. This commit updates the fixture to be
proper and includes additional fixtures for 1x and 2x ESes.

This also uses the SafeURI class where appropriate to mask passwords.

Fixes #525


lib/logstash/outputs/elasticsearch/http_client/pool.rb
```diff
@@ -237,7 +270,7 @@ def perform_request_to_url(url, method, path, params={}, body=nil)
     end
 
     def normalize_url(uri)
-      raise ArgumentError, "Only URI objects may be passed in!" unless uri.is_a?(URI)
+      raise ArgumentError, "Only SafeURI objects may be passed in!" unless uri.is_a?(URI) || uri.is_a?(LogStash::Util::SafeURI)
```
jordansissel: This allows URI and SafeURI, but the exception indicates only SafeURI is accepted. Confusing?

lib/logstash/outputs/elasticsearch/http_client/pool.rb
```diff
@@ -151,15 +151,48 @@ def sniff!
     ES2_SNIFF_RE_URL  = /([^\/]*)?\/?([^:]*):([0-9]+)/
     # Sniffs and returns the results. Does not update internal URLs!
     def check_sniff
-      url, resp = perform_request(:get, '_nodes')
+      url, resp = perform_request(:get, '_nodes/http')
```
jordansissel: This looks like a behavioral change (selecting only http nodes)?

Also, I can't find docs on the /_nodes/http endpoint. I think I found where it's defined in the Elasticsearch source code, but not in the docs.

CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## 5.4.1
+- Correctly perform healthchecks against ES 5.x nodes
```
jordansissel: My reading of the code makes me believe this changelog entry to be inaccurate.

It changes how we sniff for http nodes, doesn't it, not how we do health checks?

lib/logstash/outputs/elasticsearch/http_client/pool.rb
andrewvc: Good catch!

CHANGELOG.md
andrewvc: Hah, I recently changed the healthcheck in a previous PRe and got some wires crossed. Good catch :)

andrewvc:
@jordansissel I've just pushed and addressed your PR review

lib/logstash/outputs/elasticsearch/http_client/pool.rb
andrewvc: @jordansissel I copied that from @karmi 's code in the elasticsearch ruby client.

https://github.com/elastic/elasticsearch-ruby/blob/master/elasticsearch-transport/lib/elasticsearch/transport/transport/sniffer.rb

andrewvc: Also, it's not a behavioral change, before we had code to just ignore nodes without an HTTP endpoint. This is a bit more resource efficient since we don't need to receive nodes without HTTP. We can't do anything with those nodes anyway :)

jordansissel: Ahh ok, thanks for the clarification. :)

jordansissel:
LGTM

PR #533, opened 2016-12-20
Handle situations when bulk response has no items

### LINKED ISSUE #532: support absense of items key/value in the bulk response
in some situations it can happen that the _bulk index response doesn't contain the `items: []` element in it.
we should tolerate its absense since right now it causes an exception.

suyograo:
Fixes #532

### DIFF
--- a/lib/logstash/outputs/elasticsearch/http_client.rb
+++ b/lib/logstash/outputs/elasticsearch/http_client.rb
@@ -91,7 +91,7 @@ module LogStash; module Outputs; class ElasticSearch;
     def join_bulk_responses(bulk_responses)
       {
         "errors" => bulk_responses.any? {|r| r["errors"] == true},
-        "items" => bulk_responses.reduce([]) {|m,r| m.concat(r["items"])}
+        "items" => bulk_responses.reduce([]) {|m,r| m.concat(r.key?("items")? r["items"]: [])}
       }
     end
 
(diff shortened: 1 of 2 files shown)


lib/logstash/outputs/elasticsearch/http_client.rb
```diff
@@ -91,7 +91,7 @@ def bulk(actions)
     def join_bulk_responses(bulk_responses)
       {
         "errors" => bulk_responses.any? {|r| r["errors"] == true},
-        "items" => bulk_responses.reduce([]) {|m,r| m.concat(r["items"])}
+        "items" => bulk_responses.reduce([]) {|m,r| m.concat(r.key?("items")? r["items"]: [])}
```
jsvd: I suggest instead:
```ruby
"items" => bulk_responses.reduce([]) {|m,r| m.concat(r.fetch("items", []))}
```

spec/unit/outputs/elasticsearch/http_client_spec.rb
```diff
@@ -97,6 +97,49 @@
     end
   end
 
+  describe "bulk response" do
```
jsvd: I suggest `describe "join_bulk_responses" do` instead, this way we're describing the behaviour of this method.

suyograo:
@jsvd updated.

jsvd:
LGTM

PR #534, opened 2016-12-20
[WIP] provide flag for absolute health URL

### LINKED ISSUE #531: add flag to allow healthcheck_path to be interpreted as either relative or absolute path

suyograo:
Fixes #531

need tests


andrewvc:
WDYT about naming the option `absolute_healthcheck_path`. It'd be more consistent with other options IMHO.

suyograo:
@andrewvc sounds good. I renamed it `absolute_healthcheck_path`.

Closed in favor of https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/535

PR #535, opened 2016-12-21
Add option to use an absolute healthcheck path

### LINKED ISSUE #531: add flag to allow healthcheck_path to be interpreted as either relative or absolute path

suyograo:
Fixes #531

### DIFF
--- a/lib/logstash/outputs/elasticsearch.rb
+++ b/lib/logstash/outputs/elasticsearch.rb
@@ -176,7 +176,11 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
   # to service requests. If you have custom firewall rules you may need to change
   # this
   config :healthcheck_path, :validate => :string, :default => "/"
-
+  # When a `healthcheck_path` config is provided, this additional flag can be used to 
+  # specify whether it is an absolute URI or a relative path. 
+  # If this flag is true, `healthcheck_path` is expected to be a fully formed URL 
+  # with any basic auth credentials provided in the URL itself.
+  config :absolute_healthcheck_path, :validate => :boolean, :default => false
   # How frequently, in seconds, to wait between resurrection attempts.
   # Resurrection is the process by which backend endpoints marked 'down' are checked
   # to see if they have come back to life
--- a/lib/logstash/outputs/elasticsearch/http_client/pool.rb
+++ b/lib/logstash/outputs/elasticsearch/http_client/pool.rb
@@ -27,10 +27,13 @@ module LogStash; module Outputs; class ElasticSearch; class HttpClient;
       end
     end
 
-    attr_reader :logger, :adapter, :sniffing, :sniffer_delay, :resurrect_delay, :healthcheck_path
+    attr_reader :logger, :adapter, :sniffing, :sniffer_delay, :resurrect_delay, :healthcheck_path, :absolute_healthcheck_path
+
+    ROOT_URI_PATH = '/'.freeze
 
     DEFAULT_OPTIONS = {
-      :healthcheck_path => '/'.freeze,
+      :healthcheck_path => ROOT_URI_PATH,
+      :absolute_healthcheck_path => false,
       :scheme => 'http',
       :resurrect_delay => 5,
       :sniffing => false,
@@ -46,6 +49,7 @@ module LogStash; module Outputs; class ElasticSearch; class HttpClient;
       @url_normalizer = options[:url_normalizer]
       DEFAULT_OPTIONS.merge(options).tap do |merged|
         @healthcheck_path = merged[:healthcheck_path]
+        @absolute_healthcheck_path = merged[:absolute_healthcheck_path]
         @resurrect_delay = merged[:resurrect_delay]
         @sniffing = merged[:sniffing]
         @sniffer_delay = merged[:sniffer_delay]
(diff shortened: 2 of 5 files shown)


suyograo:
@jsvd @andrewvc my rspec-foo is not great. can you check if I'm doing the right thing in these tests?

lib/logstash/outputs/elasticsearch/http_client/pool.rb
```diff
@@ -225,9 +229,16 @@ def healthcheck!
...
       @state_mutex.synchronize { @url_info.select {|url,meta| meta[:state] != :alive } }.each do |url,meta|
         begin
-          logger.info("Running health check to see if an Elasticsearch connection is working",
-                      url: url, healthcheck_path: healthcheck_path)
-          response = perform_request_to_url(url, :head, healthcheck_path)
+          if @absolute_healthcheck_path
+            absolute_healthcheck_safe_url = ::LogStash::Util::SafeURI.new(healthcheck_path)
+            logger.info("Running health check to see if an Elasticsearch connection is working",
+                        absolute_healthcheck_url: absolute_healthcheck_safe_url.sanitized)
+            response = perform_request_to_url(healthcheck_path, :head, ROOT_URI_PATH)
```
suyograo: @andrewvc should `perform_request_to_url` use the safe URL?

spec/unit/outputs/elasticsearch/http_client/pool_spec.rb
```diff
@@ -44,6 +44,25 @@
       expect(subject).to receive(:healthcheck!).once
       sleep(subject.resurrect_delay + 1)
     end
+
+    context "when using an absolute_healthcheck_path option" do
```
jsvd: I suggest using describe:

```ruby
describe "absolute_healthcheck_path" do
  context "when enabled" do
    # ...
  end
end
```

spec/unit/outputs/elasticsearch/http_client/pool_spec.rb
```diff
@@ -44,6 +44,25 @@
...
+
+      before do
+        pool.start
+      end
+
+      after do
+        pool.close
+      end
+
+      it "should use the healthcheck_path as URL to do a health check" do
```
jsvd: look at this feature, I'd probably test that:

1) when `absolute_healthcheck_path` is enabled, `perform_request_to_url` is called with the `url` argument equal to the url constructed from `healthcheck_path`

2) when disabled, it behaves as before: appending the path to the existing hosts url (again observable in `perform_request_to_url`'s arguments

3) what happens when you pass a full URL and disable `absolute_healthcheck_path` (should raise exception)

4) what happens when you pass a path and enable `absolute_healthcheck_path` (should raise exception)

lib/logstash/outputs/elasticsearch/http_client/pool.rb
```diff
@@ -225,9 +229,16 @@ def healthcheck!
       # Try to keep locking granularity low such that we don't affect IO...
       @state_mutex.synchronize { @url_info.select {|url,meta| meta[:state] != :alive } }.each do |url,meta|
         begin
-          logger.info("Running health check to see if an Elasticsearch connection is working",
-                      url: url, healthcheck_path: healthcheck_path)
-          response = perform_request_to_url(url, :head, healthcheck_path)
+          if @absolute_healthcheck_path
```
jsvd: I suggest deduping this into something like:

```ruby
          if @absolute_healthcheck_path
            url = absolute_healthcheck_safe_url.sanitized
            path = ROOT_URI_PATH
          else
            url = url # superfluous
            path = healthcheck_path
          end
          logger.info("Running health check to see if an Elasticsearch connection is working",
                        :url => url, :healthcheck_path => path)
          response = perform_request_to_url(url, :head, path)
```

lib/logstash/outputs/elasticsearch/http_client/pool.rb
andrewvc: At this point everything uses the SafeURI. We only unwrap it briefly in the ManticoreAdapter *just* before passing it in to Manticore (which does not understand SafeURI)

lib/logstash/outputs/elasticsearch.rb
```diff
@@ -176,7 +176,11 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
   # to service requests. If you have custom firewall rules you may need to change
   # this
   config :healthcheck_path, :validate => :string, :default => "/"
-
+  # When a `healthcheck_path` config is provided, this additional flag can be used to 
+  # specify whether it is an absolute URI or a relative path. 
+  # If this flag is true, `healthcheck_path` is expected to be a fully formed URL 
+  # with any basic auth credentials provided in the URL itself.
+  config :absolute_healthcheck_path, :validate => :boolean, :default => false
```
jsvd: in order for this setting to reach the pool class, it is necessary to pass it down through the http_client_builder and http_client too.

suyograo: Done

suyograo:
@jsvd I wasn't able to get the tests updated.

jsvd:
this seems to be failing because the top level `before :each` of this spec suite introduces a stub on the adapter that is getting in the way of the failed test.
it might be necessary to move [this](https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/master/spec/unit/outputs/elasticsearch/http_client/pool_spec.rb#L15) to localized spec sections (instead of global).

jsvd:
closed in favor https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/538

PR #528, opened 2016-12-09
Add support for external version numbers

vvcephei:
I'm just pushing an initial commit so I can see the tests run in TravisCI.

For some reason, I'm unable to run the tests locally. They fail with this error:

```
$ bundle exec rspec
The signal EXIT is in use by the JVM and will not work correctly on this platform
bundler: failed to load command: rspec (/home/john/repos/baz/logstash-output-elasticsearch/vendor/bundle/jruby/1.9/bin/rspec)
RuntimeError: 

        you might need to reinstall the gem which depends on the missing jar or in case there is Jars.lock then resolve the jars with `lock_jars` command

no such file to load -- org/apache/logging/log4j/log4j-core/2.6.2/log4j-core-2.6.2 (LoadError)
... (stacktrace) ...
```

PR #536, opened 2016-12-21
add version and version_type support

neumaics:
This allows the template to define version and external version type.

Also, PR #528 covers the same functionality--perhaps we can collaborate.


andrewvc:
Agreed that #528 also covers this. #528 seems to have more specs, which it has in its favor.

Maybe you could hop into #528 and add some feedback?

neumaics:
Apologies--I've just got back from holiday and didn't give feedback to #528. I'll close this PR, though.

PR #538, opened 2016-12-27
Add option to use an absolute healthcheck path

### LINKED ISSUE #531: add flag to allow healthcheck_path to be interpreted as either relative or absolute path

jsvd:
Fixes #531

Replaces #535


suyograo:
LGTM

PR #543, opened 2017-01-13
Fix body on bad response codes

andrewvc:
If debug logging is enabled the request body will be logged as well.


jsvd:
cool find, LGTM

PR #545, opened 2017-01-19
No longer require escaping of 'password' option for special chars

andrewvc:
There was a regression at some point where the `password` option required URL escaping. This fixes that issue by escaping the option for you.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## 6.2.2
+- Fixed a bug that forced users to URL encode the `password` option.
+  If you are currently manually escaping your passwords upgrading to this version
+  will break authentication. You should unescape your password if you have implemented
+  this workaround as it will otherwise be doubly encoded.
+  URL escaping is STILL required for passwords inline with URLs in the `hosts` option. 
+
 ## 6.2.1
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+## 6.2.2
+- No longer require URL escaping in passwords given in the `password` option.
```
jordansissel: I think it's worth calling out this as a bug fix.

lib/logstash/outputs/elasticsearch/http_client_builder.rb
```diff
@@ -119,11 +121,27 @@ def self.setup_ssl(logger, params)
 
     def self.setup_basic_auth(logger, params)
       user, password = params["user"], params["password"]
-      return {} unless user && password
+      unsafe_password = password && password.value
+      unsafe_escaped_password = unsafe_password ? CGI.escape(unsafe_password) : nil
+      
+      # TODO: Remove this when we release LS6.0.0
+      if unsafe_password =~ /%[0-9A-Fa-f]{2}/
+        m <<-EOM
```
jordansissel: 👍

jordansissel:
LGTM

PR #546, opened 2017-01-20
Properly decode passwords when setting auth header

andrewvc:
6.2.2 broke passwords that need URL escaping completely. This fixes that PR by properly decoding the passwords before setting auth header.


ph:
I've tested the following case with LS 5.1.2

password: "changeme#"
hosts with escaped **#**: "http://elastic:changeme%23@localhost:9200"

LGTM

PR #547, opened 2017-01-20
Actually do the bump to 6.2.3

andrewvc:
Forgot to actually bump the gemspec in #546


ph:
LGTM so many small things to check :)

PR #548, opened 2017-01-20
Add auth integration tests

andrewvc:
This adds integration tests against an endpoint using both SSL and HTTP Basic auth so as to better catch bugs.

This also removed the buggy warning for escaped chars in the password field. This warning never made sense because if you had those chars in bad versions of logstash it would just crash.


suyograo:
LGTM

PR #550, opened 2017-01-23
Cleanup integration test use of insist + alt HTTP clients

andrewvc:
Removes dev dependency on 'elasticsearch' and 'ftw' gems.
Adds new helpers for testing against ES directly using Manticore.
Converts all legacy `insist` tests to standard `expect`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,8 @@
 ---
+addons:
+  apt:
+    packages:
+      - nginx
 sudo: false
 jdk: oraclejdk8
 env:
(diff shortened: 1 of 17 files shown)


jordansissel:
Specs pass for me locally including integration tests on 5.x

jordansissel:
I forgot about this PR. I'm very much in favor of eliminating `insist` in favor of `expect`. 

History for what it's worth: I wrote insist for C years ago, and still use it, and kind of enhanced it when I moved it to Ruby. I needed it back in the times of RSpec 2 when everything was "foo.should be ...` -- but with RSpec 3 bringing `expect` I am so so so so happy :)

Thank you for this work.

jsvd:
the code has deviated a lot from since this PR was created, I opened #756 to replace this.

PR #551, opened 2017-01-25
Reject default password of 'changeme'

jordansissel:
Safety first!

The intent of this is to force users to actually change the default
password often given to users by administrators. Further, x-pack
has a default password of 'changeme' and so this now requires users to
change this password in Elasticsearch before Logstash will run.


jordansissel:
Fixed a test failure where I forgot to check if `@password` was nil before calling .value on it.

talevy:
LGTM. 

tests seem unrelated to this, but may be caused by this: https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/548/files#diff-147dd12a53887174f4d0c86a301ca6c1R108.

jordansissel:
@talevy I can't reproduce the errors on Travis when I run locally; and some of them pass but others do not. Seems ... funky.

jordansissel:
Rebased against master.

PR #553, opened 2017-02-03
Use the correct Content-Type header

andrewvc:
When we switched away from using the 'elasticsearch' gem to using our own internal implementation we didn't port this behavior. Official clients should send the correct content-type header.

This patch also adds an extra rspec scope for integration tests requiring SSL so that tests will pass again on travis. Those tests broke a couple commits ago and that wasn't detected because at the time travis-ci was broken even though they passed locally with nginx started.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,9 +3,9 @@ sudo: false
 jdk: oraclejdk8
 env:
 - INTEGRATION=false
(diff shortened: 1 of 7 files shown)


jordansissel:
LGTM; I haven't tested.

jsvd:
LGTM

PR #554, opened 2017-02-06
change url handling of health check requests

jsvd:
This commit makes `absolute_healthcheck_path` change how the healthcheck_path is treated: either append it to any existing path (default) or replace any existing path

Also ensures that the healthcheck url contains no query parameters regarless of hosts urls contains them or query_params being set

depends on https://github.com/elastic/logstash/pull/6645


andrewvc:
@jsvd FYI tests are failing. Feel free to assign to me after they pass

jsvd:
@andrewvc as mentioned in the PR, it depends on elastic/logstash#6645

jsvd:
@andrewvc actually, to not depend on a change on core (which would mean this plugin would only be installable on logstash > 5.2.1 and >5.3, I'm going to replace the clone call

jsvd:
to make this change easier to merge, I'll remove the sniffing part of it.

PR #558, opened 2017-02-10
Implement HTTP redirect caching

rayharris-ibm:
Hi. I opened a topic on discuss.elastic.co about a week ago asking for feedback on interest in adding the ability to cache HTTP redirects in filebeat about a week ago. I had planned on discussing adding the feature to this plugin as well. However, as I got no feedback on the original topic, I'm going to go ahead and submit the PR. I've completed a CLA as rayharris-ibm/harrisra@us.ibm.com. And here is the link to the discuss topic:

https://discuss.elastic.co/t/add-ability-to-cache-http-redirects-in-elasticsearch-output/73901

### DIFF
--- a/lib/logstash/outputs/elasticsearch.rb
+++ b/lib/logstash/outputs/elasticsearch.rb
@@ -103,6 +103,9 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
   # not also set this field. That will raise an error at startup
   config :path, :validate => :string
 
+  # Follow and cache HTTP redirects
+  config :cache_redirect, :validate => :boolean, :default => false
+
   # Pass a set of key value pairs as the URL query string. This query string is added
   # to every host listed in the 'hosts' configuration. If the 'hosts' list contains
   # urls that already have query strings, the one specified here will be appended.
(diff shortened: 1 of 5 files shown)


suyograo:
@rayharris-ibm thanks for this patch. I'll review this shortly. Can you please rebase once #563 gets in?

jordansissel:
Some thoughts in no particular order:

* This is very specific to exactly 1 user with a custom in-house system. (so far)

This PR adds a new setting. Every setting we add incurs an additional burden on users for learning to use this plugin. You've laid out the case that IBM would use this internally because of design decisions made with a different internal project within IBM, but is there any examples of someone needing this outside of IBM?

If the answer is no, or if we cannot come up with compelling story for other users finding value with this feature, I am likely to resist your proposal.

* 302 vs 301

This may seem like pedantry, but it feels like if we are going to be caching redirects, we should only cache code 301 (Moved Permanently), not 302 (Found).

* Cache ...

for how long? and what key to cache on (url? host? path?)? on what conditions do we expire entries? do users need an API for expiring the cache manually? Should the cache survive restarts and reloads?

* No tests

There are no tests for this PR. It's a pretty significant change in behavior.

andrewvc:
Thanks for the PR @rayharris-ibm . I think what I would recommend for your specific use case, mentioned in your discuss post, would be to use an HTTP load balancer instead of connecting directly.

I'm against us handling redirects for all the reasons @jordansissel mentioned. Specifically, this significantly increases the complexity of this plugin for a very very niche problem. I think there are good alternatives here. Have you considered sticking an HTTP LB between LS and the backend clusters, and using that to migrate them?

jordansissel:
@rayharris-ibm any thoughts?

jordansissel:
Closing idle.

PR #559, opened 2017-02-14
Added gzip compression support

dinnot:
Signed CLA: https://elasticsearch.na1.echosign.com/public/viewAgreement?tsid=CBFCIBAA3AAABLblqZhA5OlzpZs-E9WCGvdpyuUvHIwxrJQNglMowf-fapfJDZTMGQ03XLOsUjaqxEfO-HSUTWVWRHYv8Y_QN-fGamcoi&

### DIFF
--- a/lib/logstash/outputs/elasticsearch.rb
+++ b/lib/logstash/outputs/elasticsearch.rb
@@ -203,6 +203,9 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
   # See https://hc.apache.org/httpcomponents-client-ga/httpclient/apidocs/org/apache/http/impl/conn/PoolingHttpClientConnectionManager.html#setValidateAfterInactivity(int)[these docs for more info]
   config :validate_after_inactivity, :validate => :number, :default => 10000
 
+  # Enable or disable gzip compression
+  config :use_gzip, :validate => :boolean, :default => false
+
   def build_client
     @client ||= ::LogStash::Outputs::ElasticSearch::HttpClientBuilder.build(@logger, @hosts, params)
   end
--- a/lib/logstash/outputs/elasticsearch.rb
+++ b/lib/logstash/outputs/elasticsearch.rb
@@ -204,7 +204,7 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
   config :validate_after_inactivity, :validate => :number, :default => 10000
 
   # Enable or disable gzip compression
-  config :use_gzip, :validate => :boolean, :default => false
+  config :upload_compression, :validate => :boolean, :default => false
 
   def build_client
     @client ||= ::LogStash::Outputs::ElasticSearch::HttpClientBuilder.build(@logger, @hosts, params)
--- a/lib/logstash/outputs/elasticsearch/http_client.rb
+++ b/lib/logstash/outputs/elasticsearch/http_client.rb
@@ -4,6 +4,8 @@ require "base64"
 require 'logstash/outputs/elasticsearch/http_client/pool'
 require 'logstash/outputs/elasticsearch/http_client/manticore_adapter'
 require 'cgi'
+require 'zlib'
+require 'stringio'
 
 module LogStash; module Outputs; class ElasticSearch;
   # This is a constant instead of a config option because
@@ -22,6 +24,31 @@ module LogStash; module Outputs; class ElasticSearch;
   TARGET_BULK_BYTES = 20 * 1024 * 1024 # 20MiB
 
   class HttpClient
+    # Gzip library from rails ActiveSupport
+    module Gzip
+        class Stream < StringIO
+            def initialize(*)
+            super
+            set_encoding "BINARY"
+            end
+            def close; rewind; end
+        end
+
+        # Decompresses a gzipped string.
+        def self.decompress(source)
+            Zlib::GzipReader.new(StringIO.new(source)).read
+        end
+
+        # Compresses a string using gzip.
+        def self.compress(source, level=Zlib::DEFAULT_COMPRESSION, strategy=Zlib::DEFAULT_STRATEGY)
+            output = Stream.new
+            gz = Zlib::GzipWriter.new(output, level, strategy)
+            gz.write(source)
+            gz.close
+            output.string
+        end
+    end
+
     attr_reader :client, :options, :logger, :pool, :action_count, :recv_count
     # This is here in case we use DEFAULT_OPTIONS in the future
     # DEFAULT_OPTIONS = {
@@ -129,8 +156,13 @@ module LogStash; module Outputs; class ElasticSearch;
     end
 
     def bulk_send(bulk_body)
-      # Discard the URL
-      url, response = @pool.post("_bulk", nil, bulk_body)
+      params = {}
+      if use_gzip then 
+        params[:headers] = {"Content-Encoding" => "gzip"}
+        bulk_body = Gzip::compress(bulk_body)
+      end
+      # Discard the URL 
+      url, response = @pool.post("_bulk", params, bulk_body)
       LogStash::Json.load(response.body)
     end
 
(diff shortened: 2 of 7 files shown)


suyograo:
@dinnot thanks much for the contribution. This is a much requested feature!

Looks like manticore HTTP client which we use internally already supports `compression` handling. See https://github.com/cheald/manticore/blob/master/lib/manticore/client.rb#L140. Also mentioned here https://github.com/cheald/manticore#major-features.

We should enable and use it. 

Also, the setting to turn on compression should be defaulted on Logstash side to be turned off. This plugin can be used to send data to ES 2.x and 5.x. ES 5.x enables compression by default when the right Content-Type is set, but in ES 2.x, you need to enable it.

And, this needs integration tests :) I could provide pointers/help to add one.

lib/logstash/outputs/elasticsearch.rb
```diff
@@ -203,6 +203,9 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
   # See https://hc.apache.org/httpcomponents-client-ga/httpclient/apidocs/org/apache/http/impl/conn/PoolingHttpClientConnectionManager.html#setValidateAfterInactivity(int)[these docs for more info]
   config :validate_after_inactivity, :validate => :number, :default => 10000
 
+  # Enable or disable gzip compression
+  config :use_gzip, :validate => :boolean, :default => false
```
suyograo: How about the name `compression` which takes `true` or `false`. Defaults to `false`

lib/logstash/outputs/elasticsearch/http_client.rb
```diff
@@ -22,6 +24,31 @@ module LogStash; module Outputs; class ElasticSearch;
   TARGET_BULK_BYTES = 20 * 1024 * 1024 # 20MiB
 
   class HttpClient
+    # Gzip library from rails ActiveSupport
+    module Gzip
```
suyograo: No need for this. Handled by manticore and in turn by the HTTP commons library.

lib/logstash/outputs/elasticsearch/http_client.rb
```diff
@@ -129,8 +156,13 @@ def join_bulk_responses(bulk_responses)
     end
 
     def bulk_send(bulk_body)
-      # Discard the URL
-      url, response = @pool.post("_bulk", nil, bulk_body)
+      params = {}
+      if use_gzip then 
+        params[:headers] = {"Content-Encoding" => "gzip"}
```
suyograo: Needs to be `gzip, deflated`. Also this should be a constant.

lib/logstash/outputs/elasticsearch/http_client.rb
dinnot: Hi suyograo. Thank you for the tips, I never worked with neither ruby or es before, so they are really appreciated. I had a look at the manticore compression feature, but that is client-side. So, when compression is true (which is the case by default), it will send a header with "Accept-Encoding: gzip, deflate" to inform the server that it can handle both algos. Then, if the server knows about http compression, it can look at the accept-encoding and decide which algo it wants to use, and sends the data back with the response header "Content-Encoding: algo_chosen". So there is no harm always having the flag on, as the server decides to encode the data based on what the client knows. 

In our case, ElasticSearch app also implements encoding on receiving data on the bulk api, meaning that we have to act like the server would and set the Content-Encoding header to the algorithm we used and compress the content using the same algorithm. This could had been done with encoding-negociation, but I haven't managed to make es sending it's Accept-Encoding header.

lib/logstash/outputs/elasticsearch.rb
dinnot: As per my comment below, should we use upload_compression to make sure is differentiated from the download encoding which is done by manticore automatically trough protocol negotiation?

lib/logstash/outputs/elasticsearch/http_client.rb
dinnot: should I make the whole header {"Content-Encoding" => "gzip"} a constant or just the string gzip from the header value? On the DEFAULT_HEADERS constant I've seen it's doing the whole header.

lib/logstash/outputs/elasticsearch/http_client/manticore_adapter.rb
```diff
@@ -47,7 +47,7 @@ def client
     # @see    Transport::Base#perform_request
     #
     def perform_request(url, method, path, params={}, body=nil)
-      params = (params || {}).merge(@client_params)
+      params = (params || {}).merge(@client_params) { |key, oldval, newval| (oldval.class.to_s == 'Hash' && newval.class.to_s == 'Hash') ? oldval.merge(newval) : newval }
```
andrewvc: The class check should be: `oldval.is_a?(Hash)`. This also needs an explanatory comment.

karmi:
Hi @dinnot, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/559.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

lib/logstash/outputs/elasticsearch/http_client.rb
```diff
@@ -129,8 +156,13 @@ def join_bulk_responses(bulk_responses)
     end
 
     def bulk_send(bulk_body)
-      # Discard the URL
-      url, response = @pool.post("_bulk", nil, bulk_body)
+      params = {}
+      if upload_compression then 
+        params[:headers] = {"Content-Encoding" => "gzip"}
+        bulk_body = Gzip::compress(bulk_body)
```
talevy: Hi @dinnot, what do you think about making the compression here be more streaming oriented. To avoid having two copies of the bulk_body (`bulk_body` + `bulk_actions`), we can just keep a streaming `bulk_body` that is either a compressed, or not compressed StringIO object.

lib/logstash/outputs/elasticsearch/http_client.rb
```diff
@@ -22,6 +24,31 @@ module LogStash; module Outputs; class ElasticSearch;
...
+        class Stream < StringIO
+            def initialize(*)
+            super
+            set_encoding "BINARY"
+            end
+            def close; rewind; end
+        end
+
+        # Decompresses a gzipped string.
+        def self.decompress(source)
```
talevy: unneeded

dinnot:
@talevy I've implemented streams but I can't notice any speed/memory improvements. Also, It seems to be breaking integration tests now. Could you have a look at the code if it's what you were expecting? Cheers

andrewvc:
@dinnot the stream code looks about right, I'm not sure why its breaking integration tests. Thanks for putting in the extra effort on this patch, it is much appreciated.

That said I think streaming is a hard requirement. We know that duplication of events has caused memory issues before. It mostly shows up when users have a large number of large events. For small use cases duplication is trivial, but for some use cases this distinction is critical.

dinnot:
Made some fixes and now the integration tests are passed. @suyograo Can you please take another look at this pr? Would be good if we could use this in production without having to deploy our own forked sourcecode.

suyograo:
@dinnot Thanks. Almost there now. Can you please add an integration test for this patch? You can add a new 

```
describe "with upload compression turned on" do
      let(:config) do
        {
          "hosts" => ["https://#{user}:#{CGI.escape(password)}@localhost:9900"],
          "upload_compression" => true,
          "index" => index
        }
      end

```
after this line: https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/master/spec/integration/outputs/index_spec.rb#L113

You can then verify if the right headers are being sent, like: https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/master/spec/integration/outputs/index_spec.rb#L86

And also, it uses a real ES to send compressed requests and un-compresses them.

Thanks for your patience!

dinnot:
@suyograo  Added integration tests. Can you please have a look if they are ok?

suyograo:
LGTM. Thanks @dinnot

suyograo:
@dinnot @talevy I changed this to include both requests and response compression: https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/563

Also, fixed up the broken integration tests.

PR #562, opened 2017-02-16
introduce customization of bulk, healthcheck and sniffing paths

jsvd:
This commit removes absolute_healthcheck_path and query_parameters
and introduces two new configs settings: `sniffing_path` and `bulk_path`.

The behaviour of the healthcheck, bulk and sniffing paths is now:
- if not set: the default value will be used
- if not set and path is also set: the default is appended to path
- if set: the set value will be used, ignoring the default and path values

These three settings also support query parameters, which will be
carried and included in the request to elasticsearch

### DIFF
--- a/lib/logstash/outputs/elasticsearch/http_client/pool.rb
+++ b/lib/logstash/outputs/elasticsearch/http_client/pool.rb
@@ -28,13 +28,14 @@ module LogStash; module Outputs; class ElasticSearch; class HttpClient;
       end
     end
 
-    attr_reader :logger, :adapter, :sniffing, :sniffer_delay, :resurrect_delay, :healthcheck_path, :absolute_healthcheck_path
+    attr_reader :logger, :adapter, :sniffing, :sniffer_delay, :resurrect_delay, :healthcheck_path, :sniffing_path, :bulk_path
 
     ROOT_URI_PATH = '/'.freeze
 
     DEFAULT_OPTIONS = {
       :healthcheck_path => ROOT_URI_PATH,
-      :absolute_healthcheck_path => false,
+      :sniffing_path => "/_nodes/http",
+      :bulk_path => "/_bulk",
       :scheme => 'http',
       :resurrect_delay => 5,
       :sniffing => false,
@@ -49,8 +50,9 @@ module LogStash; module Outputs; class ElasticSearch; class HttpClient;
       raise ArgumentError, "No URL Normalizer specified!" unless options[:url_normalizer]
       @url_normalizer = options[:url_normalizer]
       DEFAULT_OPTIONS.merge(options).tap do |merged|
+        @bulk_path = merged[:bulk_path]
+        @sniffing_path = merged[:sniffing_path]
         @healthcheck_path = merged[:healthcheck_path]
-        @absolute_healthcheck_path = merged[:absolute_healthcheck_path]
         @resurrect_delay = merged[:resurrect_delay]
         @sniffing = merged[:sniffing]
         @sniffer_delay = merged[:sniffer_delay]
@@ -152,7 +154,7 @@ module LogStash; module Outputs; class ElasticSearch; class HttpClient;
     ES2_SNIFF_RE_URL  = /([^\/]*)?\/?([^:]*):([0-9]+)/
     # Sniffs and returns the results. Does not update internal URLs!
     def check_sniff
-      url, resp = perform_request(:get, '_nodes/http')
+      _, resp = perform_request(:get, @sniffing_path)
       parsed = LogStash::Json.load(resp.body)
       
       nodes = parsed['nodes']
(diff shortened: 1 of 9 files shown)


lib/logstash/outputs/elasticsearch/http_client/pool.rb
```diff
@@ -230,17 +232,11 @@ def healthcheck!
...
-          healthcheck_url = LogStash::Util::SafeURI.new(url.uri.clone)
-          healthcheck_url.query = nil
-          if @absolute_healthcheck_path
-            healthcheck_url.path = ROOT_URI_PATH
-          end
           logger.info("Running health check to see if an Elasticsearch connection is working",
-                        :healthcheck_url => healthcheck_url, :path => path)
-          response = perform_request_to_url(healthcheck_url, :head, path)
+                        :healthcheck_url => url, :path => @healthcheck_path)
+          response = perform_request_to_url(url, :head, @healthcheck_path)
```
andrewvc: Love this cleanup!

lib/logstash/outputs/elasticsearch/http_client_builder.rb
```diff
@@ -26,7 +24,25 @@ def self.build(logger, hosts, params)
       common_options[:timeout] = params["timeout"] if params["timeout"]
 
       if params["path"]
-        client_settings[:path] = "/#{params["path"]}/".gsub(/\/+/, "/") # Normalize slashes
+        client_settings[:path] = "/#{params["path"]}/".gsub(/\/+/, "/")
+      end
+
+      common_options[:bulk_path] = if params["bulk_path"]
+         "/#{params["bulk_path"]}".gsub(/\/+/, "/")
```
andrewvc: Can we split out this gsub into its own function since we're using it everywhere?

spec/unit/outputs/elasticsearch/http_client/pool_spec.rb
```diff
@@ -45,46 +45,27 @@
...
-
-      context "when using query params in the initial url" do
-        let(:initial_urls) { [::LogStash::Util::SafeURI.new("http://localhost:9200?q=s")] }
-        it "is removed from the healthcheck request" do
+      let(:initial_urls) { [::LogStash::Util::SafeURI.new("http://localhost:9200")] }
+
+      context "and not setting healthcheck_path" do
+        it "performs the healthcheck to the root" do
           expect(adapter).to receive(:perform_request) do |url, method, req_path, _, _|
-            expect(method).to eq(:head)
```
andrewvc: Why did we drop the check for the method?

lib/logstash/outputs/elasticsearch/http_client_builder.rb
jsvd: yeah I knew you'd point it out and makes perfect sense. will refactor now

spec/unit/outputs/elasticsearch/http_client/pool_spec.rb
```diff
@@ -45,46 +45,27 @@
...
-            end
-            subject.healthcheck!
-          end
-        end
-
-        context "when disabled" do
-          let(:absolute_healthcheck_path) { false }
-          it "should use the healthcheck_path as a relative path" do
-            expect(adapter).to receive(:perform_request) do |url, method, req_path, _, _|
-              expect(method).to eq(:head)
```
andrewvc: Also, why remove the method check?

spec/unit/outputs/elasticsearch/http_client/pool_spec.rb
jsvd: IMO that was an extra concern not related to this test which was checking the path handling of the health check request, the fact that the request is a HEAD is included in another test

andrewvc:
Great PR! Needs some minor tweaks however.

spec/unit/outputs/elasticsearch/http_client/pool_spec.rb
jsvd: re-added!

jsvd:
@andrewvc thanks for the review, can you give another pass at it?

lib/logstash/outputs/elasticsearch/http_client_builder.rb
```diff
@@ -131,5 +149,10 @@ def self.setup_basic_auth(logger, params)
         :password => unsafe_escaped_password
       }
     end
+
+    private
+    def self.sanitize_url(url)
```
andrewvc: Can we rename this to `dedupe_slashes` or something. Sanitization implies that it removes some sort of secret or dangerous threat. That's how its used in `SafeURI` which makes this name confusing here.

After that its LGTM, no need for another pass.

andrewvc:
@jsvd looking good, I added one final comment, after that's addressed feel free to merge/release!

lib/logstash/outputs/elasticsearch/http_client_builder.rb
jsvd: Sure, I used sanitize as "making sane", or "removing craziness", but I'll rename as you suggested

jsvd:
@andrewvc just a heads up, due to removal of two settings, I'm bumping to 7.0.0

PR #563, opened 2017-02-23
Change the compression feature to include gzip'ing responses and requests

suyograo:
This PR:
* Adds `http_compression` to enable compressing both the request and response.
* Response compression is handled automatically by HTTPCommons, but the correct headers (`Accept-Encoding`) needs to be sent.
* Adds a new file for compressed indexing tests. The previous test attempt was borked
* Bumps version

Follow up from #559

### DIFF
--- a/lib/logstash/outputs/elasticsearch.rb
+++ b/lib/logstash/outputs/elasticsearch.rb
@@ -202,7 +202,7 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
   # See https://hc.apache.org/httpcomponents-client-ga/httpclient/apidocs/org/apache/http/impl/conn/PoolingHttpClientConnectionManager.html#setValidateAfterInactivity(int)[these docs for more info]
   config :validate_after_inactivity, :validate => :number, :default => 10000
 
-  # Enable or disable gzip compression on requests and responses
+  # Enable gzip compression on requests. Note that response compression is on by default for Elasticsearch v5.0 and beyond
   config :http_compression, :validate => :boolean, :default => false
 
   def build_client
--- a/lib/logstash/outputs/elasticsearch.rb
+++ b/lib/logstash/outputs/elasticsearch.rb
@@ -202,8 +202,8 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
   # See https://hc.apache.org/httpcomponents-client-ga/httpclient/apidocs/org/apache/http/impl/conn/PoolingHttpClientConnectionManager.html#setValidateAfterInactivity(int)[these docs for more info]
   config :validate_after_inactivity, :validate => :number, :default => 10000
 
-  # Enable or disable gzip compression
-  config :upload_compression, :validate => :boolean, :default => false
+  # Enable or disable gzip compression on requests and responses
+  config :http_compression, :validate => :boolean, :default => false
 
   def build_client
     @client ||= ::LogStash::Outputs::ElasticSearch::HttpClientBuilder.build(@logger, @hosts, params)
--- a/lib/logstash/outputs/elasticsearch.rb
+++ b/lib/logstash/outputs/elasticsearch.rb
@@ -67,8 +67,8 @@ require "forwardable"
 #
 # This plugin supports request and response compression. Response compression is enabled by default and 
 # for Elasticsearch versions 5.0 and later, the user doesn't have to set any configs in Elasticsearch for 
-# it to send back compressed response. For version before 5.0, `http.compression` must be set to `true` to 
-# take advantage of response compression using this plugin
+# it to send back compressed response. For versions before 5.0, `http.compression` must be set to `true` in 
+# Elasticsearch to take advantage of response compression when using this plugin
 #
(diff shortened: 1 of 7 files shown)


spec/integration/outputs/compressed_indexing_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+require "stringio"
+require "gzip"
+
+RSpec::Matchers.define :a_valid_gzip_encoded_string do
+  match { |data|
+    expect { Zlib::GzipReader.new(StringIO.new(data)).read }.not_to raise_error
+  }
+end
+
+describe "indexing with upload compression", :integration => true, :version_greater_than_equal_to_5x => true do
```
talevy: s/upload/http

spec/integration/outputs/compressed_indexing_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+      result = LogStash::Json.load(response.body)
+      result["hits"]["hits"].each do |doc|
+        expect(doc["_type"]).to eq(type)
+        expect(doc["_index"]).to eq(index)
+      end
+    end
+  end
+
+  it "sets the correct content-encoding header and body is compressed" do
+    expect(subject.client.pool.adapter.client).to receive(:send).
```
talevy: do we want to have a similar check on the adapter upon receiving? to verify that Elasticsearch sent the correct gzip encoding back?

dinnot: I think that would test if elasticsearch server has gzip or deflate support, rather than this code. Idea with Accept-Encoding is that you specify which algorithms the http server knows to decode, but the server isn't forced to encode the data if it doesn't want to.

lib/logstash/outputs/elasticsearch/http_client.rb
```diff
@@ -133,7 +133,7 @@ def join_bulk_responses(bulk_responses)
     end
 
     def bulk_send(body_stream)
-      params = upload_compression ?  {:headers => {"Content-Encoding" => "gzip"}} : {}
+      params = http_compression ?  {:headers => {"Accept-Encoding" => "gzip,deflate", "Content-Encoding" => "gzip"}} : {}
```
dinnot: I don't think it's good for the code to add the Accept-Encoding header in here. In case Manticore drops or adds new algorithms to the set of supported ones we will be sending the wrong headers. Manticore automatically sets the header if the compression is turned on (there's an argument 'compression' which takes a bool and is true by default).

suyograo: @dinnot so you're saying response compression is on by default and we don't have to do anything?

dinnot: Yes, response compression is on by default. Manticore sets the Accept-Encoding header already and elastic can (and should, but if it doesn't it's an es lack of feature) respond with gzip encoded data.

suyograo:
@talevy ready for another round of review.

lib/logstash/outputs/elasticsearch.rb
```diff
@@ -62,6 +62,17 @@
 #
 # Keep in mind that a connection with keepalive enabled will
 # not reevaluate its DNS value while the keepalive is in effect.
+#
+# ==== HTTP Compression
+#
+# This plugin supports request and response compression. Response compression is enabled by default and 
+# for Elasticsearch versions 5.0 and later, the user doesn't have to set any configs in Elasticsearch for 
+# it to send back compressed response. For versions before 5.0, `http.compression` must be set to `true` in
```
talevy: s/response/responses

talevy: also, should we reference the docs here? https://www.elastic.co/guide/en/elasticsearch/reference/5.0/modules-http.html#_settings_2

suyograo: Will do

spec/integration/outputs/compressed_indexing_spec.rb
talevy: I guess the idea here is to verify consistent behavior between default Elasticsearch and manticore, just in case any one of them change how they handle things without Logstash knowing

suyograo: @dinnot I'm having trouble validating if manticore is indeed sending `Accept-Encoding: gzip`.  I want to validate that is on by default..

dinnot: What I did was pointing logstash to a node app instead of an es instance and logging the received headers in that app.

suyograo: I was able to verify manually that manticore sends the right gzip value for `Accept-Encoding`. I was not successful in getting the `[response](https://github.com/cheald/manticore/blob/master/lib/manticore/response.rb#L253)` object in manticore to return the original response headers from ES. Weirdly, when I use `curl -i -H  'Accept-Encoding: gzip,deflate` localhost:9200` I see gzipped data.

I am gonna leave this as is for now since manticore already has a [spec](https://github.com/cheald/manticore/blob/34a00cee57a56148629ed0a47c329181e7319af5/spec/manticore/client_spec.rb#L33) for validating this.

PR #567, opened 2017-02-26
Support Elasticsearch master (6.0.0) version.

suyograo:
* This PR adds support for 6.0 by adding a separate template for 6.x which removes _all.
* It also removes the groovy update tests because groovy has been deprecated in 5.x and removed in 6.0
* Adds infra to build against master branch of ES.


suyograo:
@jsvd thanks for keeping me honest here 👍 . I assumed the update tests were _all_ groovy. 

I've now split the update tests into: `groovy_update`, `painless_update` and update tests without any scripts in them. It does complicate the matrix we have, but it is what it is. 

Also, while testing scripted updates, some of the `sprintf` logic seem to be broken regardless of groovy or painless. I'll open a separate issue for that.

PR #568, opened 2017-02-27
Expose a `#get` method on the http_client to allow custom calls

ph:
This is really handy when you use the elasticsearch output to create the
right client with the right ssl options but you want to make a custom
call to the back.


suyograo:
@ph can you add an integration test please?

ph:
@suyograo Will do, I thought they were more oriented to test plugin behavior than the `http_client`

suyograo:
@ph Yep, I realized we never expose this as a setting... so scratch the integ tests.

suyograo:
LGTM

Issue #496, opened 2016-10-26
Improve handling of response codes from Elasticsearch


ppf2:
LS 5.0

```
[2016-10-26T12:20:22,245][ERROR][logstash.outputs.elasticsearch] Got a bad response code from server, but this code is not considered retryable. Request will be dropped {:code=>403}
```

For errors that are not retryable causing requests till drop immediately, it will be helpful to provide more than just the http status code, but also the http response message and body of content so it's easier for the end user to tell exactly what the response error is.

roadrunner:
would be great if we can specify retryable error codes at input configuration
when consuming from a message queue, i think it's important to be able to re-queue event in certain cases instead of dropping

ppf2:
LS 5.1.1

This will be especially helpful for users using X-pack security.  Elasticsearch / X-pack security provides a very descriptive message (root cause) detailing the exact privilege that is missing from the user role.  But this is completely lost in Logstash logging and only a 403 response code is provided.

```
[2016-12-28T17:09:56,008][WARN ][logstash.outputs.elasticsearch] Attempted to resurrect connection to dead ES instance, but got an error. {:url=>#<URI::HTTPS:0x2c0ba922 URL:https://~hidden~:~hidden~@instance_id.us-east-1.aws.found.io:9243>, :error_type=>LogStash::Outputs::ElasticSearch::HttpClient::Pool::BadResponseCodeError, :error=>"Got response code '403' contact Elasticsearch at URL 'https://~hidden~:~hidden~@instance_id.us-east-1.aws.found.io:9243/'"}
```

Ideally, the output plugin will be passing through the root cause message from Elasticsearch to its own logging, eg.

```
{"error":{"root_cause":[{"type":"security_exception","reason":"action [cluster:monitor/main] is unauthorized for user [~hidden~]"}],"type":"security_exception","reason":"action [cluster:monitor/main] is unauthorized
```

ppf2:
The error message reported is worse and more ambiguous if the user role is missing the `write` privilege for the index Logstash is writing to.  It simply throws an undefined method error instead of reporting the actual authorization issue. 

To reproduce, create a X-pack security user with just index and create_index privileges (but no `write` privilege).

```
[...]

PR #571, opened 2017-03-06
Mark 403 as retryable error

grypyrg:
Mark error `403` to be a retryable error!

(This is also discussed at: https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/496)

With the wrong permissions in X-Pack this give errors similar to: 
```
[ERROR][logstash.outputs.elasticsearch] Got a bad response code from server, but this code is not considered retryable. Request will be dropped {:code=>403, :response_body=>"{\"error\":{\"root_cause\":[{\"type\":\"security_exception\",\"reason\":\"action [indices:data/write/bulk] is unauthorized for user [****]\"}]...
```

instead of having data loss, the request should be retried.

### DIFF
--- a/lib/logstash/outputs/elasticsearch/common.rb
+++ b/lib/logstash/outputs/elasticsearch/common.rb
@@ -4,7 +4,7 @@ module LogStash; module Outputs; class ElasticSearch;
   module Common
     attr_reader :client, :hosts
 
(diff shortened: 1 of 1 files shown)


suyograo:
Handled in https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/583

Issue #576, opened 2017-03-17
ES output failing with long messages


gmoskovicz:
- Version: Elasticsearch 5.2.2
- Operating System: ANY
- Config File (if you have sensitive info, please remove it)

```
input {
    file {
        path => [ "/Users/Gabriel/Documents/ElasticSearch/logstash-5.2.2/file/content.txt" ]
        type => "test"
        start_position => "beginning"
        sincedb_path => "/dev/null"
        ignore_older => 0
    }
}
filter {

}
output {
    elasticsearch {
        hosts => [ "localhost:9200" ]
        index => "test-1"
    }
}
```

- Sample Data:

Content.txt file (unzip it after download): [content.txt.zip](https://github.com/logstash-plugins/logstash-output-elasticsearch/files/850986/content.txt.zip)

- Steps to Reproduce:

[1] Run the above config with `content.txt` file (>50MB 1 line file)
[2] Wait until the message is processed 
[3] Verify that you get the following error:

```
[2017-03-17T12:12:44,260][ERROR][logstash.outputs.elasticsearch] Got a bad response code from
 server, but this code is not considered retryable. Request will be dropped {:code=>400, 
:response_body=>"{\"error\":{\"root_cause\":[{\"type\":\"parse_exception\",\"reason\":\"Failed to
 derive xcontent\"}],\"type\":\"parse_exception\",\"reason\":\"Failed to derive 
xcontent\"},\"status\":400}"}
```

In wireshark you can verify that the operation is sent like this (not the Content Length 0):

```
Hypertext Transfer Protocol
    POST /_bulk HTTP/1.1\r\n
    Connection: Keep-Alive\r\n
    Content-Type: application/json\r\n
    Content-Length: 0\r\n
        [Content length: 0]
    Host: localhost:9200\r\n
    User-Agent: Manticore 0.6.1\r\n
    Accept-Encoding: gzip,deflate\r\n
    \r\n
    [Full request URI: http://localhost:9200/_bulk]
    [HTTP request 4/4]
    [Prev request in frame: 17]
    [Response in frame: 23]
```

Basically when you have a very big file, Logstash is failing to send it correctly as a bulk request.

gmoskovicz:
CC @ajoliveira

jsvd:
[...]

PR #577, opened 2017-03-20
fix bulk_size_bytes handling of single big event

jsvd:
fixes https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/576

### DIFF
--- a/lib/logstash/outputs/elasticsearch/http_client.rb
+++ b/lib/logstash/outputs/elasticsearch/http_client.rb
@@ -116,7 +116,9 @@ module LogStash; module Outputs; class ElasticSearch;
                     action.map {|line| LogStash::Json.dump(line)}.join("\n") :
                     LogStash::Json.dump(action)
         as_json << "\n"
-        bulk_responses << bulk_send(body_stream) if (body_stream.size + as_json.bytesize) > TARGET_BULK_BYTES
+        if (body_stream.size + as_json.bytesize) > TARGET_BULK_BYTES
+          bulk_responses << bulk_send(body_stream) unless body_stream.size == 0
+        end
         stream_writer.write(as_json)
       end
       stream_writer.close if http_compression
@@ -234,7 +236,7 @@ module LogStash; module Outputs; class ElasticSearch;
     end
 
     def http_compression
-      client_settings.fetch(:http_compression, {})
+      client_settings.fetch(:http_compression, false)
     end
 
     def build_adapter(options)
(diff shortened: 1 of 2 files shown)


lib/logstash/outputs/elasticsearch/http_client.rb
```diff
@@ -224,7 +226,7 @@ def ssl_options
     end
 
     def http_compression
-      client_settings.fetch(:http_compression, {})
+      client_settings.fetch(:http_compression, false)
```
jsvd: this is a copy paste mistake. it is not necessary for the PR, but it was making some code paths use http_compression when it shouldn't, since `{}` was being used instead of false.

lib/logstash/outputs/elasticsearch/http_client.rb
```diff
@@ -116,7 +116,9 @@ def bulk(actions)
                     action.map {|line| LogStash::Json.dump(line)}.join("\n") :
                     LogStash::Json.dump(action)
         as_json << "\n"
-        bulk_responses << bulk_send(body_stream) if (body_stream.size + as_json.bytesize) > TARGET_BULK_BYTES
+        if (body_stream.size + as_json.bytesize) > TARGET_BULK_BYTES
+          bulk_responses << bulk_send(body_stream) unless body_stream.size == 0
```
jordansissel: Can you talk me through this fix? We only send if the body is nonzero size, but why would we have a zero-sized body in the first place -- how would this occur?

jsvd:
The algorithm currently sends whatever is in the buffer if the (buffer
size)+(current action size) trips the threshold.
The problem is that in the first step of the iteration on a new call to
bulk the buffer is always empty and if a big events appears, the
conditional kicks in, and the current (empty) buffer is sent.

jordansissel:
@jsvd ahh, that makes sense. Thank you for explaining :)

jsvd:
@jordansissel can you give a 👍 / 👎 / LGTM on this? thanks

lib/logstash/outputs/elasticsearch/http_client.rb
andrewvc: The reason is that on the first iteration that is always the case, we don't add to the stream till after this line.

jordansissel: Still feels like a weird way to phrase it. The surrounding code is difficult for me to read, but the patch doesn't make it any worse, so I am OK merging this.

jsvd:
merged in 8789ab550e20053d01c178a46a70893d687dff8d

PR #579, opened 2017-03-24
Expose a `#post` method on the http client until we have a more reusable in the library

ph:
Xpack is using the ES output http client to make all the connections to the backend and make sure we have the same SSL support, but it needs to be able to make raw request to ES.


suyograo:
LGTM

PR #581, opened 2017-03-25
Add plugin level required to support sending content-type headers

dedemorton:
Adds compatibility note requested in https://github.com/elastic/logstash/issues/6718

@acchen97 Please review.


acchen97:
LGTM. Is there someway we can have this backported all the way back to 2.4 for visibility?

dedemorton:
I can manually tweak the asciidoc. We'll just has to be careful that we don't overwrite the changes if we need to regenerate the asciidoc at some point.

PR #583, opened 2017-04-07
Use a dead letter queue to handle mapping errors.

### LINKED ISSUE #572: Retry on unexpected response codes
cloud.elastic.co's proxy will return HTTP 502 if the cluster is being restarted. Currently, the plugin considers 502 to be *NOT* retryable, which means a cluster restart will cause data loss when the Elasticsearch output targets Cloud.

```
[2017-03-07T01:20:06,418][ERROR][logstash.outputs.elasticsearch] Got a bad response code from server, but this code is not considered retryable. Request will be dropped {:code=>502, :response_body=>"{\"ok\":false,\"message\":\"Unable to connect to the server.\"}"}
```

This is an unexpected problem because we only have targeted Elasticsearch-documented response codes in our work, here, and have never planned for an interfering gateway. Should be an easy fix to align more well with the way elastic cloud and other proxy or load balancer systems work.

suyograo:
Also retry policy has been changed. This is mostly motivated from #572 and the use of DLQ.
We now retry everything except for 400 and 409.

Fixes #572

### DIFF
--- a/lib/logstash/outputs/elasticsearch/common.rb
+++ b/lib/logstash/outputs/elasticsearch/common.rb
@@ -1,10 +1,11 @@
 require "logstash/outputs/elasticsearch/template_manager"
+java_import org.logstash.common.DeadLetterQueueFactory
 
 module LogStash; module Outputs; class ElasticSearch;
   module Common
     attr_reader :client, :hosts
 
-    RETRYABLE_CODES = [429, 503]
+    DLQ_CODES = [400]
     SUCCESS_CODES = [200, 201]
     CONFLICT_CODE = 409
 
@@ -16,6 +17,7 @@ module LogStash; module Outputs; class ElasticSearch;
 
     def register
       @stopping = Concurrent::AtomicBoolean.new(false)
+      @dlq_writer = LogStash::Util::DeadLetterQueueFactory::get(self)
       setup_hosts # properly sets @hosts
       build_client
       install_template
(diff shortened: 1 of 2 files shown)


lib/logstash/outputs/elasticsearch/common.rb
```diff
@@ -128,16 +130,23 @@ def submit(actions)
...
           next
-        elsif RETRYABLE_CODES.include?(status)
-          @logger.info "retrying failed action with response code: #{status} (#{failure})"
+        elsif DLQ_CODES.include?(status)
+          action_event = action[2]
+          @dlq_writer.write(event, "Could not index event to Elasticsearch. status: #{status}, action: #{action}, response: #{response}")
+          next
+        else
+          # only log what the user whitelisted
+          @logger.info "retrying failed action with response code: #{status} (#{failure})" if !failure_type_logging_whitelist.include?(failure["type"])
```
suyograo: I could probably use `unless` here.

jordansissel:
>  We now retry everything except for 400 and 429.

You mean "retry everythign except for 400" ? We need to retry 429

lib/logstash/outputs/elasticsearch/common.rb
```diff
@@ -128,16 +130,23 @@ def submit(actions)
...
         if SUCCESS_CODES.include?(status)
           next
         elsif CONFLICT_CODE == status && VERSION_TYPES_PERMITTING_CONFLICT.include?(action_params[:version_type])
           @logger.debug "Ignoring external version conflict: status[#{status}] failure[#{failure}] version[#{action_params[:version]}] version_type[#{action_params[:version_type]}]"
           next
-        elsif RETRYABLE_CODES.include?(status)
-          @logger.info "retrying failed action with response code: #{status} (#{failure})"
+        elsif DLQ_CODES.include?(status)
+          action_event = action[2]
+          @dlq_writer.write(event, "Could not index event to Elasticsearch. status: #{status}, action: #{action}, response: #{response}")
```
jordansissel: Let's put a TODO item to change this api to be similar to our logging api where we can include context as an object:

Example proposed:

```
dlq.write(event, "Could not index event to Elasticsearch", status: status, action: action, response: response)
```

lib/logstash/outputs/elasticsearch/common.rb
jordansissel: +1 but not necessary since this is a refactor of existing code.

jordansissel:
Next steps:

* Need a logstash-core that supports DLQ manager
* Can we do this change without breaking compatibility? Can we do DLQ optionally? Or backport this to the current-major somehow?

lib/logstash/outputs/elasticsearch/common.rb
```diff
@@ -16,6 +17,7 @@ module Common
 
     def register
       @stopping = Concurrent::AtomicBoolean.new(false)
+      @dlq_writer = LogStash::Util::DeadLetterQueueFactory::get(self)
```
jordansissel: To support bwc, this should check if DLQ even exists. If it doesn't, we could provide a null dlq instance (perhaps one implemented in this plugin so it can work in older Logstash versions)

lib/logstash/outputs/elasticsearch/common.rb
```diff
@@ -16,6 +16,8 @@ module Common
 
     def register
       @stopping = Concurrent::AtomicBoolean.new(false)
+      # To support BWC, we check if DLQ exists in core (< 5.4). If it doesn't, we use nil to resort to previous behavior.
+      @dlq_writer = LogStash::Util.const_defined?(:DeadLetterQueueFactory)? LogStash::Util::DeadLetterQueueFactory::get(self): nil
```
jordansissel: spacing between `?` and `:`

jordansissel:
Lets also add tests to cover the new retry behavior, perhaps tailored to unexpected interference from things like proxies (injecting 502 response codes, etc)

jordansissel:
LGTM.

Next steps:

* Test against master/5.x where we have execution_context
* Also test against master/5.x when DLQ lands.

suyograo:
tested with DLQ.

PR #585, opened 2017-04-11
add a few metrics to the plugin

jsvd:
~~- number of sniffing requests performed~~
~~- number of bulk_requests measured before right before executing the
request~~
~~- number of concurrent bulk_requests~~
~~- number of bulk retries~~

metrics calculated:

```ruby
{
  "bulk_requests": {
     "responses": {  # counts per top level status code returned in the bulk response
       "200": 346300,
       "201": 712,
       "404": 61,
       "429": 36
       # other status codes ...
     },
     "successes": 346819,  # successful bulk requests with no errors (sum of 200 and 201 minus partial_successes)
     "failures": 23,         # no http response from ES. will be retried. see point 1 above for causes
     "partial_successes": 134,  # bulk request successful (status 20X) but contains errors
     "drops": 5,   # bulk request dropped
  },
  "documents": {
       "successes": 173409500,
       "retryable_failures": 450,
       "non_retryable_failures": 12
  },
  "sniffing_count": 621
}
```

this is the context of https://github.com/elastic/logstash/issues/6831

### DIFF
--- a/lib/logstash/outputs/elasticsearch/common.rb
+++ b/lib/logstash/outputs/elasticsearch/common.rb
@@ -25,6 +25,8 @@ module LogStash; module Outputs; class ElasticSearch;
 
       install_template
       check_action_validity
+      @bulk_request_metrics = metric.namespace(:bulk_requests)
+      @document_level_metrics = metric.namespace(:documents)
 
       @logger.info("New Elasticsearch output", :class => self.class.name, :hosts => @hosts.map(&:sanitized).map(&:to_s))
     end
@@ -156,8 +158,15 @@ module LogStash; module Outputs; class ElasticSearch;
 
       # If the response is nil that means we were in a retry loop
       # and aborted since we're shutting down
+      return if bulk_response.nil?
+
       # If it did return and there are no errors we're good as well
-      return if bulk_response.nil? || !bulk_response["errors"]
+      if bulk_response["errors"]
+        @bulk_request_metrics.increment(:with_errors)
+      else
+        @bulk_request_metrics.increment(:successes)
+        return
+      end
 
       actions_to_retry = []
       bulk_response["items"].each_with_index do |response,idx|
@@ -173,8 +182,10 @@ module LogStash; module Outputs; class ElasticSearch;
         # - For a mapping error, we send to dead letter queue for a human to intervene at a later point.
         # - For everything else there's mastercard. Yep, and we retry indefinitely. This should fix #572 and other transient network issues
         if DOC_SUCCESS_CODES.include?(status)
+          @document_level_metrics.increment(:successes)
           next
         elsif DOC_CONFLICT_CODE == status
+          @document_level_metrics.increment(:non_retryable_failures)
           @logger.warn "Failed action.", status: status, action: action, response: response if !failure_type_logging_whitelist.include?(failure["type"])
           next
         elsif DOC_DLQ_CODES.include?(status)
@@ -186,9 +197,11 @@ module LogStash; module Outputs; class ElasticSearch;
           else
             @logger.warn "Could not index event to Elasticsearch.", status: status, action: action, response: response
           end
+          @document_level_metrics.increment(:non_retryable_failures)
           next
         else
           # only log what the user whitelisted
+          @document_level_metrics.increment(:retryable_failures)
           @logger.info "retrying failed action with response code: #{status} (#{failure})" if !failure_type_logging_whitelist.include?(failure["type"])
           actions_to_retry << action
         end
@@ -239,6 +252,7 @@ module LogStash; module Outputs; class ElasticSearch;
 
         # We retry until there are no errors! Errors should all go to the retry queue
         sleep_interval = sleep_for_interval(sleep_interval)
+        @bulk_request_metrics.increment(:failures)
         retry unless @stopping.true?
       rescue ::LogStash::Outputs::ElasticSearch::HttpClient::Pool::NoConnectionAvailableError => e
         @logger.error(
@@ -249,8 +263,10 @@ module LogStash; module Outputs; class ElasticSearch;
         )
         Stud.stoppable_sleep(sleep_interval) { @stopping.true? }
         sleep_interval = next_sleep_interval(sleep_interval)
+        @bulk_request_metrics.increment(:failures)
         retry unless @stopping.true?
       rescue ::LogStash::Outputs::ElasticSearch::HttpClient::Pool::BadResponseCodeError => e
+        @bulk_request_metrics.increment(:failures)
         log_hash = {:code => e.response_code, :url => e.url.sanitized.to_s}
         log_hash[:body] = e.response_body if @logger.debug? # Generally this is too verbose
         message = "Encountered a retryable error. Will Retry with exponential backoff "
@@ -280,6 +296,7 @@ module LogStash; module Outputs; class ElasticSearch;
         @logger.debug("Failed actions for last bad bulk request!", :actions => actions)
 
         sleep_interval = sleep_for_interval(sleep_interval)
+        @bulk_request_metrics.increment(:failures)
         retry unless @stopping.true?
       end
     end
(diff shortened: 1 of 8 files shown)


jordansissel:
@jsvd cla failure? weird...

lib/logstash/outputs/elasticsearch/http_client.rb
```diff
@@ -135,12 +136,16 @@ def join_bulk_responses(bulk_responses)
     def bulk_send(body_stream)
       params = http_compression ? {:headers => {"Content-Encoding" => "gzip"}} : {}
       # Discard the URL 
+      @metric.increment(:bulk_requests)
+      @metric.increment(:concurrent_bulk_requests)
```
jordansissel: What action would a user take based on this 'concurrent_bulk_requests' value? And could that same action not be informed by the other metric (bulk_requests total count)?

jsvd: If the number of concurrent requests is consistently equal to the number of workers then it can hint that back pressure will happen  in the pipeline.
This number can't be extracted from the number of requests. It may be estimated by taking into account the millis spent doing requests and wallclock time, but seems far fetched.

Also there was a request in 
https://github.com/elastic/logstash/issues/6544 for this too

jordansissel: For pipelines where the main latency comes from Elasticsearch writes (or the network between), this value is likely to always be the same as number-of-workers. 

For example, if a single batch take 1ms to process and 50ms to land in Elasticsearch (or maybe worse, if you're using Cloud where your ES cluster is very distant). This scenario has each batch taking 51ms, where 98% of the time is spent talking to Elasticsearch, so roughly 98% of the time this `concurrent_bulk_requests` metric will be equal to the number of pipeline workers.

My assumption is that the network is generally slower than the cpu (in terms of time-spent-per-batch), and given this, I do not believe 'concurrent_bulk_requests' to be a useful value because it can be normal and healthy in many cases to have 'concurrent_bulk_requests' any value from 0 to number-of-workers.

jordansissel: Here's my thought on what a user might want to know:

* Is the ES output stuck?
* Is the ES output slow?

For these questions, concurrent_bulk_requests (CBS) doesn't tell us enough, partly because you need to know more context (Is CBS always max because ES is our main bottleneck? Are event counts going up?)

In terms of the questions I proposed above --

* Is ES stuck?

We can infer this by observing two values today: `in` vs `out` count for the ES output being nonzero *and* having both values stable (not increasing). Across two observations (of reasonable time, say, 10 seconds), if `in` is higher than `out` and neither value is growing, then we know ES output is waiting on Elasticsearch.

* Is ES slow?

We can infer this by observing the `duration_in_millis`/`out` to compute per-event latency through Elasticsearch. If the event latency is higher than expected, then Elasticsearch output is a bottleneck.

jsvd: From what I tested locally, the value of this metric was either 0 or `pipeline.workers`. Because my pipelines were always light on processing, the bottle neck was ES, hence the value == `pipeline.workers`. However alt+tabbing and performing other tasks in my laptop would make this value become 0 temporarily.

I am under the impression that in production it's rare for this value to be anything other than 0-1. Otherwise it would mean the pipeline is under constant backpressure if the event rate is > Batch Size / Time per back, or  in this case 125/ 0.051s = 2450 eps

jsvd: > Is ES stuck?
> We can infer this by observing two values today: in vs out count for the ES output being nonzero and having both values stable (not increasing). Across two observations (of reasonable time, say, 10 seconds), if in is higher than out and neither value is growing, then we know ES output is waiting on Elasticsearch.

++ I think there are many ways to see the pipeline is stalled 

> Is ES slow?
> We can infer this by observing the duration_in_millis/out to compute per-event latency through Elasticsearch. If the event latency is higher than expected, then Elasticsearch output is a bottleneck.

My concern here is that by computing values per event while logstash works pretty much by batch may lead to wrong conclusions. I don't think CBC is also a crystal clear metric, but can help understand the pipeline behaviour if the values are consistently 0 (no problems in `out` increases), number of workers (may signify backpressure). I don't think values in between can have any meaning.

jsvd: btw I'm ok with removing the concurrent bulk requests from this PR and move the discussion to another issue.

acchen97: @allenmchan any feedback here regarding the usage of concurrent bulk requests?

FYI we're discussing the overall ES output stats here: https://github.com/elastic/logstash/issues/6831#issuecomment-295691299

jsvd:
refactored to report metrics described in https://github.com/elastic/logstash/issues/6831#issuecomment-297152712:


```
{
  "bulk_requests": {
     "responses": {  # counts per top level status code returned in the bulk response
       "200": 346300,
       "201": 712,
       "404": 61,
       "429": 36
       # other status codes ...
     },
     "successes": 346819,  # successful bulk requests with no errors (sum of 200 and 201 minus partial_successes)
     "failures": 23,         # no http response from ES. will be retried. see point 1 above for causes
     "partial_successes": 134,  # bulk request successful (status 20X) but contains errors
     "drops": 5,   # bulk request dropped
  },
  "documents": {
       "successes": 173409500,
       "retryable_failures": 450,
       "non_retryable_failures": 12
  },
  "sniffing_count": 621
}
```

A note that the bulk requests bucket counts per status code use a string instead of a number.
@ph can you confirm that we can't have metric names as numbers in core? I couldn't find a way to do it, since internally they become symbols, and `200.to_sym` fails

ph:
> the bulk requests bucket counts per status code use a string instead of a number.
@ph can you confirm that we can't have metric names as numbers in core? I couldn't find a way to do it, since internally they become symbols, and 200.to_sym fails

Correct, they currently can't be number, @jsvd can you create a followup issue on core to fix that for 6.x?

jsvd:
received feedback where we should measure drops at the bulk request level as it should never happen, which led to the creation of [an issue to investigate the current dropping scenarios in bulk handling](https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/595)

jsvd:
I created https://github.com/elastic/logstash/issues/7009 to track the inability to have metrics that are numbers and not strings.

lib/logstash/outputs/elasticsearch/common.rb
```diff
@@ -119,8 +121,15 @@ def submit(actions)
 
       # If the response is nil that means we were in a retry loop
       # and aborted since we're shutting down
+      return if bulk_response.nil?
+
       # If it did return and there are no errors we're good as well
-      return if bulk_response.nil? || !bulk_response["errors"]
+      if bulk_response["errors"]
+        @bulk_request_metrics.increment(:partial_successes)
```
jsvd: can I counter suggest with calling it `with_errors` ? because the bulk request actually returns 200, but will contain errors.

suyograo: Sounds good.

lib/logstash/outputs/elasticsearch/common.rb
```diff
@@ -119,8 +121,15 @@ def submit(actions)
...
       # and aborted since we're shutting down
+      return if bulk_response.nil?
+
       # If it did return and there are no errors we're good as well
-      return if bulk_response.nil? || !bulk_response["errors"]
+      if bulk_response["errors"]
+        @bulk_request_metrics.increment(:with_errors)
+      else
+        @bulk_request_metrics.increment(:successes)
+        return
```
robbavey: Returning here means that document level success metrics are not incremented unless the bulk response contains errors

acchen97:
@ycombinator wanted to make sure you're aware of this augmentation to metrics collection for the ES output. We should make sure this update is reflected respectively in the x-pack metrics shipper and also effectively exposed in the new metrics drawer in the pipeline viewer when it ships.

lib/logstash/outputs/elasticsearch/common.rb
jsvd: well spotted, fixed 👍

jsvd:
@robbavey I added tests to some of the metrics, for both a successful and faulty bulk request

PR #586, opened 2017-04-12
support sniffing_path customization

jsvd:
same with healthcheck path allow customization of sniffing path by introducing `sniffing_path` and `absolute_sniffing_path` configuration parameters

NOTE: this PR should be only applied to 6.x major version

### DIFF
--- a/spec/unit/outputs/elasticsearch/http_client/pool_spec.rb
+++ b/spec/unit/outputs/elasticsearch/http_client/pool_spec.rb
@@ -102,6 +102,42 @@ describe LogStash::Outputs::ElasticSearch::HttpClient::Pool do
         expect(subject.sniffer_alive?).to eql(true)
       end
     end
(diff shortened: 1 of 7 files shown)


spec/unit/outputs/elasticsearch/http_client/pool_spec.rb
```diff
@@ -102,6 +102,42 @@
...
+
+      context "when enabled" do
+        let(:absolute_sniffing_path) { true }
+        it "should use the sniffing_path as the absolute path" do
+          expect(subject).to receive(:perform_request_to_url) do |url, method, req_path, _, _|
+            expect(method).to eq(:get)
+            expect(req_path).to eq(sniffing_path)
+            expect(url.path).to eq("/")
+            response
+          end
```
ph: I should use that block form, I always forget about it.

ph:
Missing changelog and gemspec bump.

PR #591, opened 2017-04-24
Make travis-run.sh script runnable on non-travis envs

andrewvc:
This will set a default value for TRAVIS_BUILD_DIR (./) that
makes sense on a local laptop for testing.

This also renames travis-build.sh to build.sh to reflect the fact
that it can be used on envs other than travis


suyograo:
LGTM. Merge on green

PR #593, opened 2017-04-26
Log 429 errors as debug

andrewvc:
Before these were all logged at the `error` level. This confused users
and caused issues like the one seen here:
https://discuss.elastic.co/t/throttling-processing-of-log-backfill-using-filebeat/83740/3

This patch should result in fewer false alarms

### DIFF
--- a/lib/logstash/outputs/elasticsearch/common.rb
+++ b/lib/logstash/outputs/elasticsearch/common.rb
@@ -247,7 +247,17 @@ module LogStash; module Outputs; class ElasticSearch;
         if RETRYABLE_CODES.include?(e.response_code)
           log_hash = {:code => e.response_code, :url => e.url.sanitized}
           log_hash[:body] = e.body if @logger.debug? # Generally this is too verbose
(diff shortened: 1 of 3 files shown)


lib/logstash/outputs/elasticsearch/common.rb
```diff
@@ -247,7 +247,17 @@ def safe_bulk(actions)
...
           log_hash[:body] = e.body if @logger.debug? # Generally this is too verbose
-          @logger.error("Attempted to send a bulk request to elasticsearch but received a bad HTTP response code!", log_hash)
+          message = "Encountered a retryable error. Will Retry with exponential backoff "
+
+          # We treat 429s as a special case because these really aren't errors, but
+          # rather just ES telling us to back off a bit, which we do.
+          # The other retryable codes are 502 and 503, which are true errors
+          # Even though we retry the user should be made aware of these
+          if e.response_code == 429
+              @logger.debug(message, log_hash)
```
jordansissel: indent isn't quite right.

jordansissel:
LGTM, minor style comment.

PR #594, opened 2017-04-27
fix bwc layer for dlq detection

jsvd:
the bwc layer for dlq detection only checks for the presence of execution_context, not that the execution_context can execute dlq_writer, this PR fixes it.


suyograo:
LGTM

PR #600, opened 2017-05-24
Fix broken $BUILD_PATH in build script

andrewvc:
The build script was only working on travis, this makes it work locally again


suyograo:
LGTM

PR #601, opened 2017-05-24
Fix handling of retryable errors

andrewvc:
This fixes the retryable errors handling in a few key ways:

1. It fixes the missing RETRYABLE_CODES constant
2. It doesn't stop retrying just because logstash is shutting down (we
don't want to lose data here!)
3. It adds tests for this behavior

Right now the output is broken pretty badly, since the missing constant will cause it to crash on those exceptions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 7.3.2
+ - Fix error where a 429 would cause this output to crash
+ - Wait for all inflight requests to complete before stopping
+
 ## 7.3.1
  - Fix the backwards compatibility layer used for detecting DLQ capabilities in logstash core
 
--- a/lib/logstash/outputs/elasticsearch/common.rb
+++ b/lib/logstash/outputs/elasticsearch/common.rb
@@ -4,6 +4,11 @@ module LogStash; module Outputs; class ElasticSearch;
   module Common
     attr_reader :client, :hosts
 
+    # These are codes for temporary recoverable conditions
+    # 429 just means that ES has too much traffic ATM
+    # 503 means it , or a proxy is temporarily unavailable
+    RETRYABLE_CODES = [429, 503]
+
     DLQ_CODES = [400, 404]
     SUCCESS_CODES = [200, 201]
     CONFLICT_CODE = 409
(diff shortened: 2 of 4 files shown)


lib/logstash/outputs/elasticsearch/common.rb
```diff
@@ -4,6 +4,13 @@ module LogStash; module Outputs; class ElasticSearch;
   module Common
     attr_reader :client, :hosts
 
+    # These are codes for temporary recoverable conditions
+    # 429 just means that ES has too much traffic ATM
+    # 503 means it , or a proxy is temporarily unavailable
+    RETRYABLE_CODES = [429, 503]
+
+    # These codes are not recoverable, send them to the DLQ
```
jsvd: It's somewhat out of this scope, but since the comment was added, the description for codes we put in DLQ_CODES doesn't seem consistent, as a 404 return code for an index operation may be fixed by creating the index?
if you agree I suggest leaving this without a comment and restrict the PR to fixing the missing constant + data loss

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 7.4.0
```
jsvd: since the bug was introduced in 7.3.0 and this is a bug fix PR I suggest releasing it as 7.3.2

andrewvc: I did the major bump since I also changed the retry on shutdown behavior to infinite. WDYT?

lib/logstash/outputs/elasticsearch/common.rb
andrewvc: Good point, I'll remove it

CHANGELOG.md
jsvd: can we break this into two prs with a single concern? makes it easier to ship 7.3.1 with the bug fix, and 7.4.0 with the fix + new shutdown strategy

spec/unit/outputs/elasticsearch_spec.rb
```diff
@@ -202,8 +202,39 @@
...
+        )
+      end
+      let(:logger) { double("logger").as_null_object }
+
+      before(:each) do
+        i = 0
+        bulk_param =  [["index", anything, event.to_hash]]
+
+        allow(eso).to receive(:logger).and_return(logger)
+        allow(eso.client).to receive(:bulk).with(bulk_param) do
```
jordansissel: You can `allow(eso.client).to receive(:bulk).with(bulk_param).once` and achieve the same thing -- if I am reading correctly that you are wanting to allow the `bulk` method to be called exactly 1 time?

CHANGELOG.md
jordansissel: think the previous shutdown behavior is a bug, not a feature. If someone asks logstash to shutdown, and we give up retrying because we're shutting down, that is the bug.

andrewvc:
@jsvd @jordansissel I've changed the version to a patch bump instead of a minor bump. Does that address all concerns?

spec/unit/outputs/elasticsearch_spec.rb
andrewvc: Nope, we want it to be called twice, raising an error the first time, not raising one the second.

jordansissel: Ahh, thank you for that clarity. I understand the intent now. Can you add a coment above the line to explain this intent for future readers?

jordansissel:
@andrewvc Most of the tests are passing on Travis. One failed but is likely due to a change in Elasticsearch master (types deprecation?) and seems unrelated to your change, so I am pretending Travis is green.

andrewvc:
@jordansissel I just added the comment if you don't mind taking another look

PR #605, opened 2017-05-26
Refactor use of SafeURI to conform to 9k compatible subset

### LINKED ISSUE #604: IPv6 hosts do not work with SafeURI on JRuby 9k
The `SafeURI` class we provide is backed by the Ruby URI class. The behavior in stdlib changed post ruby 2.0 such that it no longer considers `//[::1]` a valid URL. This is a commonly used URL in the elasticsearch output.

I'll be fixing this by backing `SafeURI` with `java.net.URI` which has no such problem.

andrewvc:
Fixes #604 

Requires https://github.com/elastic/logstash/pull/7236

As part of this work this removes the slash de-duplication that runs when executing HTTP requests. This is really work that doesn't need to be done. ES handles duplicate slashes fine, no reason to complicate code over it. 

The slash de-duplication was trickier to do with the changes to SafeURI, and it never had a point, hence its removal.

### DIFF
--- a/lib/logstash/outputs/elasticsearch/http_client/manticore_adapter.rb
+++ b/lib/logstash/outputs/elasticsearch/http_client/manticore_adapter.rb
@@ -84,19 +84,19 @@ module LogStash; module Outputs; class ElasticSearch; class HttpClient;
     end
 
     def format_url(url, path=nil)
-
-      request_uri = url.uri
+      request_uri = url.clone
 
       if path
         # Combine the paths using the minimal # of /s
         # First, we make sure the path is relative so URI.join does
         # the right thing
         relative_path = path && path.start_with?("/") ? path[1..-1] : path
-        # because URI.join obliterates the query parameter, we need to save it
-        # and restore it after this operation
-        query = request_uri.query
-        request_uri = URI.join(request_uri, relative_path)
-        request_uri.query = query if query
+
+        if !request_uri.path
+          request_uri.path = path
+        else
+          request_uri.path = "#{request_uri.path}/#{relative_path}"
+        end
       else
         request_uri = request_uri.clone
       end
@@ -106,8 +106,7 @@ module LogStash; module Outputs; class ElasticSearch; class HttpClient;
       request_uri.user = nil
       request_uri.password = nil
 
-      # Wrap this with a safe URI defensively against careless handling later
-      ::LogStash::Util::SafeURI.new(request_uri)
+      request_uri
     end
 
     def close
--- a/spec/unit/outputs/elasticsearch/http_client/manticore_adapter_spec.rb
+++ b/spec/unit/outputs/elasticsearch/http_client/manticore_adapter_spec.rb
@@ -19,15 +19,19 @@ describe LogStash::Outputs::ElasticSearch::HttpClient::ManticoreAdapter do
   describe "auth" do
     let(:user) { "myuser" }
     let(:password) { "mypassword" }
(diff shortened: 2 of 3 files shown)


andrewvc:
I should mention that reviewers should manually run these specs with the current logstash 5.x branch as the logstash core version to make sure the changes from https://github.com/elastic/logstash/pull/7236 still work here.

lib/logstash/outputs/elasticsearch/http_client/manticore_adapter.rb
```diff
@@ -84,19 +84,19 @@ def perform_request(url, method, path, params={}, body=nil)
...
-        # because URI.join obliterates the query parameter, we need to save it
-        # and restore it after this operation
-        query = request_uri.query
-        request_uri = URI.join(request_uri, relative_path)
-        request_uri.query = query if query
+
+        if !request_uri.path
+          request_uri.path = path
+        else
+          request_uri.path = request_uri.path + "/" + relative_path
```
jsvd: this is on the critical path for requests, so I suggest using string interpolation to create less strings

lib/logstash/outputs/elasticsearch/http_client/manticore_adapter.rb
```diff
@@ -84,19 +84,19 @@ def perform_request(url, method, path, params={}, body=nil)
...
-        # because URI.join obliterates the query parameter, we need to save it
-        # and restore it after this operation
-        query = request_uri.query
-        request_uri = URI.join(request_uri, relative_path)
-        request_uri.query = query if query
+
+        if !request_uri.path
+          request_uri.path = path
+        else
+          request_uri.path = "#{request_uri.path}/#{relative_path}"
```
jsvd: I see that URI.join goes to great lengths to "work", using convert_to_uri on the first argument ,then using URI::Generic#merge on the other arguments, which does [a lot](https://docs.ruby-lang.org/en/2.1.0/URI/Generic.html#method-i-merge).

Off the top of my head I see no issues with just using normal string joining, do you think some problems could appear here?

andrewvc: Not really, at least not for the purposes we have.

andrewvc: Do you have a specific concern?

BTW, I think its a feature that we aren't doing extra work we don't actually need anymore.

jsvd: > Do you have a specific concern?

Unknowns.. breaking things that aren't tested and didn't think about upfront. URI, like File paths are tricky things, so it makes sense most transformation of data in a URI to be done by the library instead of outside of it (e.g. Using `File.join(a, b)` instead of `#{a}/#{b}`)

andrewvc: @jsvd we don't want to use `File.join` because that will join things using `File::SEPARATOR` which can be platform specific.

If we haven't created the test, and the integration tests don't catch it, I'm OK with that risk.

andrewvc:
@jsvd I think I've addressed all concerns. The only failing test is against ES master which is a known issue.

jsvd:
@andrewvc what of https://travis-ci.org/logstash-plugins/logstash-output-elasticsearch/jobs/239615958#L951 ?

ph:
Oh, I will rebase the branch to have that commit.

PR #606, opened 2017-06-05
Bump to 7.3.3

andrewvc:
This releases 9k + 9K URI compatibility https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/605


jsvd:
we need to add some logstash-core dependency constraints, since a user may install this plugin on logstash 5.3.x, for example.

andrewvc:
@jsvd that is actually fine! It's still compatible :)

andrewvc:
@jsvd you can tell because the the tests all passed on #605 even though we haven't released a new gem.

PR #607, opened 2017-06-06
Fix 9k path concat

andrewvc:
This allows for query strings in the bulk path to work correctly.
The implementation required extra parsing of the query string to work
correctly.

This also restores some of the slash de-duplication logic removed when
we switched over to java.net.URI compatibility since it simplifies the
logic here.

### DIFF
--- a/lib/logstash/outputs/elasticsearch/http_client/manticore_adapter.rb
+++ b/lib/logstash/outputs/elasticsearch/http_client/manticore_adapter.rb
@@ -66,7 +66,7 @@ module LogStash; module Outputs; class ElasticSearch; class HttpClient;
 
       request_uri = format_url(url, path)
 
-      resp = @manticore.send(method.downcase, request_uri.to_s, params)
+      resp = @manticore.send(method.downcase, request_uri, params)
 
       # Manticore returns lazy responses by default
       # We want to block for our usage, this will wait for the repsonse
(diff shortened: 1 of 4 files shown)


lib/logstash/outputs/elasticsearch/http_client/manticore_adapter.rb
```diff
@@ -83,30 +83,30 @@ def perform_request(url, method, path, params={}, body=nil)
...
-      end
-
+      
       # We excise auth info from the URL in case manticore itself tries to stick
       # sensitive data in a thrown exception or log data
       request_uri.user = nil
       request_uri.password = nil
 
-      request_uri
+      return request_uri.to_s unless path_and_query && path_and_query != nil
```
ph: Can't we just do this ?

```ruby
 return request_uri.to_s if path_and_query.nil?

File: lib/logstash/outputs/elasticsearch/http_client/manticore_adapter.rb [hunk 2]
```diff
@@ -83,30 +83,30 @@ def perform_request(url, method, path, params={}, body=nil)
...
+      
+      parsed_path_and_query = java.net.URI.new(path_and_query)
+      
+      query = request_uri.query
+      parsed_query = parsed_path_and_query.query
+      
+      new_query_parts = [request_uri.query, parsed_path_and_query.query].select do |part|
+        part && !part.empty? # Skip empty nil and ""
+      end
```
ph: We can just compact the array?
``` 
[request_uri.query, parsed_path_and_query.query].compact.select { |part| !part.empty? }
```

ph: sadly we don't have `present?` in all ruby version.

ph:
Make sure we bump the version + changelog.

lib/logstash/outputs/elasticsearch/http_client/manticore_adapter.rb
andrewvc: Hmmm. There are some situations where people pass in a "". I suppose in that case the rest of the code wouldn't hurt though. I'll do that.

lib/logstash/outputs/elasticsearch/http_client/manticore_adapter.rb
andrewvc: @ph that's a rails-ism, not in plain ruby.

andrewvc:
@ph I believe I've addressed all issues here

lib/logstash/outputs/elasticsearch/http_client/manticore_adapter.rb
ph: I know its an active-thingy :)

PR #609, opened 2017-06-08
fix incorrect name of event to be written to dlq

talevy:


suyograo:
@talevy LGTM. Can you bump version and changelog?

PR #610, opened 2017-06-12
introduce ignore_above in string_fields default mapping for keywords

### LINKED ISSUE #588: bring back ignore_above in the default template for keyword fields?
The default [template for 2.x used ignore_above for the .raw](https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/f4cf5b67b6776d847df32f38d4fd95f835e109e0/lib/logstash/outputs/elasticsearch/elasticsearch-template-es2x.json#L26) fields. 
In the 5.x template, [this is not being used anymore](https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/f4cf5b67b6776d847df32f38d4fd95f835e109e0/lib/logstash/outputs/elasticsearch/elasticsearch-template-es5x.json#L26). 
This could lead to rejections when the content of a `.keyword` field exceeds Lucene’s term byte-length limit of 32766.

I'm not sure if we should bring it back, as it would silently accept the field without it being indexed or stored and the rejection might raise awareness to the user that he probably does not want a long text as a keyword field, but there might be use cases where the old behavior was intended.
Just want to start a discussion, if we should introduce [ignore_above](https://www.elastic.co/guide/en/elasticsearch/reference/current/ignore-above.html) again or keep it like it is.

jsvd:
fixes #588

Since the 5.x and 6.x templates override the default keyword mapping
in elasticsearch which defaults to `"ignore_above": 256`, the templates
should include this clause otherwise the burden of dealing with this
situation passes instead to Lucene.

Also, looking at the history I see no rationale for removing it in the first place, maybe it was a mistake?

### DIFF
--- a/lib/logstash/outputs/elasticsearch/elasticsearch-template-es5x.json
+++ b/lib/logstash/outputs/elasticsearch/elasticsearch-template-es5x.json
@@ -23,7 +23,7 @@
           "mapping" : {
             "type" : "text", "norms" : false,
             "fields" : {
-              "keyword" : { "type": "keyword" }
+              "keyword" : { "type": "keyword", "ignore_above": 256 }
             }
           }
         }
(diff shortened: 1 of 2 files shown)


suyograo:
LGTM

PR #616, opened 2017-07-07
Fix bug where logs would log URI class name instead of the actual URI

andrewvc:
Since the upgrade of the SafeURI class the SafeURI#sanitized method returns a java object,
which is fine, but jruby does not invoke #to_s on those objects automatically, so we must be
explicit to have those values show in the logs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,12 @@
+##  7.3.8
+ - Fix bug where java class names were logged rather than actual host names in various scenarios
+
 ## 7.3.7
  - Properly support characters needing escaping in users / passwords across multiple SafeURI implementions (pre/post LS 5.5.1)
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+NEXT(CHANGEME)
```
original-brownbear: @andrewvc why not just up the version to `7.3.6` here and publish the fix? :)

andrewvc:
Good call, I was planning on doing another PR right after this one but that
isn't happening anymore

andrewvc:
@original-brownbear this is now updated with a real version bump. I also added a changelog entry for 7.3.6 which never got one.

original-brownbear:
@andrewvc thanks LGTM :) (pending Travis :))

andrewvc:
Rebased to fix version conflicts...

PR #620, opened 2017-07-12
[WIP] Fix escaping

### LINKED ISSUE #618: Passwords needing escaping are broken
It seems that https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/605 caused there to be broken behavior here, causing situations like https://discuss.elastic.co/t/error-start-logstash-after-update-from-5-4-3-to-5-5/92418 where the password field would not be automatically URL escaped.

This behavior should be restored.

andrewvc:
Fixes #618


andrewvc:
Actually, this needs some spec fixes before review...

andrewvc:
closing in favor of https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/621

PR #621, opened 2017-07-13
Fix escaping behavior and add tests around it

### LINKED ISSUE #618: Passwords needing escaping are broken

andrewvc:
This PR takes over from #620 with a new approach. This fixes #618 . 

This depends on https://github.com/elastic/logstash/pull/7687 to actually run. Until that is merged and released integration tests here will fail. PR review should be done manually against that branch for both the `INTEGRATION` and `SECURE_INTEGRATION` branches for now.

This branch fixes escaping for the `user` parameter and also adds integration tests for secure scenarios, where user/password are used, including with odd escaped values. This all tests SSL to boot.


andrewvc:
I should add that the best way to test the specs locally is with:

```
export INTEGRATION=true
export SECURE_INTEGRATION=true
export ES_VERSION=5.5
killall -9 java # Kill any active ES instances
ci/run.sh
```

PR #623, opened 2017-07-18
Remove SECURE_INTEGRATION tests against master

andrewvc:
These specs don't make sense because x-pack master isn't publicly available. We should be well covered by the tests against the current released branch.

Alternatively we could use a different testing approach with a different HTTPS/Auth provider in the future, but since we're testing the client, not x-pack, this should all be fine unless we use something other than HTTP auth in the future.


suyograo:
LGTM

Issue #626, opened 2017-07-21
Logstash loses records on 500 status code 


dmarkhas:
One of the ES nodes in our cluster crashed today, and while the cluster kept functioning some of the bulk insert requests performed by logstash failed with the below error:

[2017-07-19T04:47:46,560][WARN ][logstash.outputs.elasticsearch] Failed action. {:status=>500, :action=>["index", {:id=>"19e9c5dc-f23e-4adc-ad46-176869872768", :index=>"computing.network_latency.08f694be-0619-485c-a3a4-0d7dd25cf1ef-2017.29", :type=>"doc", :routing=>nil}, 2017-07-19T04:45:42.038Z %{host} %{message}], :response=>{"index"=>{"index"=>"computing.networklatency.08f694be-0619-485c-a3a4-0d7dd25cf1ef-2017.29", "type"=>"doc", "id"=>"19e9c5dc-f23e-4adc-ad46-176869872768", "status"=>500, "error"=>{"type"=>"node_not_connected_exception", "reason"=>"[data1-iil-005][10.184.95.5:9300] Node not connected"}}}}

This is obviously a temporary issue, so I would expect Logstash to retry - but it didn't, since 500 isn't a retryable code as far as logstash is concerned. Either way, this data was lost. We use Logstash 5.4, but even if we were using the DLQ feature (which is not clear to me in which version it is available), it wouldn't have helped since the elasticsearch output DLQ's only 400 and 404 errors.

Is this something that should be fixed in Logstash, or in ElasticSearch? While Logstash is correct to assume 500 means a permanent failure, this particular exception is somewhat of a gray area.

andrewvc:
Hmmm, we should retry indefinitely, IMHO, on 500 errors, I'll investigate this.

andrewvc:
This will be fixed in #640

andrewvc:
By the way, which JRE/JDK version are you using? Java TLS support differs between them.

andrewvc:
@dmarkhas btw, I also just want to apologize for the bad behavior there. That is definitely a nasty bug.

dmarkhas:
We use JDK 1.8.0.92

I completely understand that you cannot always predict with 100% accuracy what kind of errors ES will return and which of them are retriable.
When I said that the data was lost that was not accurate - since we use a custom version of the plugin (as I mentioned in #572), we didn't lose any data.

dmarkhas:
[...]

PR #628, opened 2017-07-30
Extend retry HTTP codes

dropje86:
When using a service like Elastic Cloud there are more possible
  recoverable HTTP error codes because there is a proxy in between
  the ES data nodes and the client.
  Extend the list of retry codes with these errors.

### DIFF
--- a/lib/logstash/outputs/elasticsearch.rb
+++ b/lib/logstash/outputs/elasticsearch.rb
@@ -54,6 +54,7 @@ require "forwardable"
 # - Network errors (Inability to connect)
 # - 404 (Not found) Indicating an upstream proxy has no matching route
 # - 429 (Too many requests) Graceful backoff indication by Elastic
+# - 500 (Internal Server Error) when an ES node disconnect occurs this is temporarily returned
 # - 502 (Bad gateway) During shutdown a proxy might return this
 # - 503 (Service unavailable) Proxy returns an unavailable service
 # - 504 (Gateway timed out) Proxy backend timed out
--- a/lib/logstash/outputs/elasticsearch.rb
+++ b/lib/logstash/outputs/elasticsearch.rb
@@ -51,9 +51,12 @@ require "forwardable"
 #
 # The following errors are retried infinitely:
 #
-# - Network errors (inability to connect)
-# - 429 (Too many requests) and
-# - 503 (Service unavailable) errors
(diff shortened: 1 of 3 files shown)


dropje86:
I understand that there is a DLQ now but these return codes are common in situations where a reverse proxy sits in front of Elastic during "graceful" shutdowns, upgrades etc.. 404's are either a misconfiguration in Logstash -> ES endpoints or temporary missing routes in the proxy. It would also be fine if retry and DLQ codes were configurable but I feel these are generic sane defaults.

jordansissel:
@dropje86 I haven't looked at the code (CLA not signed?), but about your 404 comment. Elasticsearch itself returns 404 in certain situations and we decided this means we should DLQ it (for example, writing to an index that doesn't exist, I think, will return 404 for the status.

dropje86:
Hi @jordansissel CLA is signed and shows okay on my side. The 404 is very unlikely to be returned with writing to a non-existing index as in 99% of the use cases `action.auto_create_index` is enabled on Elasticsearch (also the default).

dropje86:
According to this issue: https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/626 it might also be a good idea to include 500 (internal server error) in the retry codes.

dropje86:
Looking at the status codes again for me it doesn't make real sense to have `CONFLICT_CODE` in there. I'd say that is a perfect use case for DLQ apart from a 400 (Bad Request), or am I missing something?

andrewvc:
First, let me say this is a great PR.

I think there's some complication because status codes are used in two places:

1. As the code of the HTTP response
2. As the code, per-event, in the bulk API

In these two contexts they mean two different things.

In the case of a 404 it means that the endpoint is missing (which is not an ES error, only a proxy error) in the context of the HTTP request, while it means an individual doc does not exist (useful for an update operation) in the context of a single event in the bulk request.

I agree that a 404 at the request level should be retried, but our handling of a 404 at the doc level is currently correct. We'll need to split this error code list into two different sets of error codes depending on *where* the error happens.

andrewvc:
I've rebased and added a commit on top of this in #640 to address this important issue. Would love your thoughts there @dropje86

andrewvc:
Closing this since this was merged in #640 and a new gem just published in the 8.x series. A new 7.4.0 version will be published soon as well with this backported

Issue #602, opened 2017-05-25
Parent/Child specs failing on latest ES master


andrewvc:
Builds like: https://travis-ci.org/logstash-plugins/logstash-output-elasticsearch/jobs/236051767
are failing with errors like:

```

org.elasticsearch.transport.RemoteTransportException: [KkXNHqC][127.0.0.1:9300][indices:data/read/search[phase/query]]
Caused by: org.elasticsearch.index.query.QueryShardException: [has_parent] query configured 'parent_type' [4627443913_parent] is not a valid type
```

We need to fix this! I suspect this is related to the removal of types in ES 6.0

suyograo:
@andrewvc I'll take this from you while you focus on other things..

jsvd:
I've done some investigation about these failures and you're right that the main problem is that elasticsearch 5.6 and 6 will only support a single type.
This has an implication in the parent/child feature of ES, where it will no longer be done through the use of multiple types, but through a `join` data type. More about this can be found inhttps://www.elastic.co/guide/en/elasticsearch/reference/5.x/parent-join.html

This implies that the current tests about parent/child must be confined to <= 5.x through tagging and new tests with the join data type must be added and run with both 5.6 and 6.x

jsvd:
an update the support of the `parent` parameter with elasticsearch 6.0.0: the syntax in the bulk request for indexing child documents has gone from

```json
{"index":{"_id":null,"_index":"1112","_type":"doc","_routing":"test"}}
{"foo": {"name": "child_value", "parent": "test"},"message":"Hello World!"}
```

to

```json
{"index":{"_id":null,"_index":"1112","_type":"doc","_routing":null,"parent":"test"}}
{"foo":"child_value","message":"Hello World!"}
```

which means the join definition must be in the _source instead of header. The es output doesn't currently have the ability to do this, so more work is required in the presence of a child document index operation

PR #632, opened 2017-08-04
Fix parent/child support for elasticsearch 6.x and support join type field

jsvd:
This PR adds a new parameter `join` to support elasticsearch 6.x join field type and associated changes in the parent/child feature, along with a test for it. It also excludes existing parent/child specs from running if the tested ES is 6.x or higher.

The specs are currently failing since scripting also requires changes to make this plugin compatible with ES 6.x, but I believe these are out of scope for this PR.

fixes https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/602

### DIFF
--- a/lib/logstash/outputs/elasticsearch/common.rb
+++ b/lib/logstash/outputs/elasticsearch/common.rb
@@ -174,8 +174,15 @@ module LogStash; module Outputs; class ElasticSearch;
         params[:pipeline] = event.sprintf(@pipeline)
       end
 
-     if @parent
-        params[:parent] = event.sprintf(@parent)
+      if @parent
+        if @join_field
+          join_value = event.get(@join_field)
+          parent_value = event.sprintf(@parent)
+          event.set(@join_field, { "name" => join_value, "parent" => parent_value })
+          params[:_routing] = event.sprintf(@parent)
+        else
+          params[:parent] = event.sprintf(@parent)
+        end
       end
 
       if @action == 'update'
--- a/spec/integration/outputs/parent_spec.rb
+++ b/spec/integration/outputs/parent_spec.rb
@@ -1,7 +1,8 @@
 require_relative "../../../spec/es_spec_helper"
 require "logstash/outputs/elasticsearch"
 
-shared_examples "a parent indexer" do
+context "when using elasticsearch 5.x and before", :integration => true, :version_less_than_equal_to_5x => true do
+  shared_examples "a type based parent indexer" do
     let(:index) { 10.times.collect { rand(10).to_s }.join("") }
     let(:type) { 10.times.collect { rand(10).to_s }.join("") }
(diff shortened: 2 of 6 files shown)


spec/integration/outputs/parent_spec.rb
```diff
@@ -43,29 +44,122 @@
...
 
-describe "(http protocol) index events with static parent", :integration => true do
-  it_behaves_like 'a parent indexer' do
-    let(:parent) { "test" }
-    let(:config) {
-      {
-        "hosts" => get_host_port,
-        "index" => index,
-        "parent" => parent
+  describe "(http protocol) index events with static parent" do
```
andrewvc: Why is this no longer an integration test? It is in the integration test folder.

lib/logstash/outputs/elasticsearch/common.rb
```diff
@@ -174,8 +174,15 @@ def event_action_params(event)
         params[:pipeline] = event.sprintf(@pipeline)
       end
 
-     if @parent
-        params[:parent] = event.sprintf(@parent)
+      if @parent
```
andrewvc: What's the story here WRT 5.x ES servers?

This change means that we no longer support parent/child in versions below 6.0

There are two paths forward:

1. We maintain separate 5.x and 6.x compat branches in this output using the gemspec to enforce which versions work with which logstash
2. We do version sniffing inside the plugin and support two different code paths for the two different parent/child impls.

Thoughts?

spec/integration/outputs/parent_spec.rb
jsvd: I just moved the integration declaration to the context that envelops this describe block. that should be enough, right?

lib/logstash/outputs/elasticsearch/common.rb
jsvd: we still support it, as long as you don't use the join_field option, then the parent option behaves as it did before.

jsvd:
@andrewvc I believe I have addressed your concerns

lib/logstash/outputs/elasticsearch/common.rb
andrewvc: Ohhhh, I misread the code. Very nice!

spec/integration/outputs/parent_spec.rb
andrewvc: Ah, I missed that, much better now :)

lib/logstash/outputs/elasticsearch/common.rb
teebu: is join_field an option in the output? I don't see it here https://www.elastic.co/guide/en/logstash/current/plugins-outputs-elasticsearch.html#plugins-outputs-elasticsearch-parent

PR #640, opened 2017-08-16
Extended Extend Retry HTTP Codes

### LINKED ISSUE #626: Logstash loses records on 500 status code

andrewvc:
This adds a commit on top of #628 . 

UPDATE:
This PR now retries ALL non-200 response codes indefinitely. 

Fixes #626

### DIFF
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -55,17 +55,18 @@ the new template is installed.
 
 ==== Retry Policy
 
-The retry policy has changed significantly in the 2.2.0 release.
+The retry policy has changed significantly in the 8.1.1 release.
 This plugin uses the Elasticsearch bulk API to optimize its imports into Elasticsearch. These requests may experience
-either partial or total failures.
+either partial or total failures. The bulk API sends batches of requests to an HTTP endpoint. Error codes for the HTTP
+request are handled differently than error codes for individual documents.
 
-The following errors are retried infinitely:
+HTTP requests to the bulk API are expected to return a 200 response code. All other response codes are retried indefinitely.
 
-- Network errors (inability to connect)
-- 429 (Too many requests) and
-- 503 (Service unavailable) errors
+The following document errors are handled as follows:
+- 400 and 404 errors are sent to the DLQ if enabled. If a DLQ is not enabled a log message will be emitted and the event will be dropped.
+- 409 errors (conflict) are logged as a warning and dropped. 
 
-NOTE: 409 exceptions are no longer retried. Please set a higher `retry_on_conflict` value if you experience 409 exceptions.
+Note that 409 exceptions are no longer retried. Please set a higher `retry_on_conflict` value if you experience 409 exceptions.
 It is more performant for Elasticsearch to retry these exceptions than this plugin.
 
 ==== Batch Sizes ====
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
+## 8.1.1
+  - Retry all non-200 responses of the bulk API indefinitely
+  - Improve documentation on retry codes
+
 ## 8.1.0
   - Support Elasticsearch 6.x join field type
+## 8.0.2
   - Fix bug where logging errors for bad response codes would raise an unhandled exception
 
 ## 8.0.1
--- a/lib/logstash/outputs/elasticsearch.rb
+++ b/lib/logstash/outputs/elasticsearch.rb
@@ -51,9 +51,13 @@ require "forwardable"
 #
 # The following errors are retried infinitely:
 #
-# - Network errors (inability to connect)
-# - 429 (Too many requests) and
-# - 503 (Service unavailable) errors
+# - Network errors (Inability to connect)
+# - 404 (Not found) Indicating an upstream proxy has no matching route
+# - 429 (Too many requests) Graceful backoff indication by Elastic
+# - 500 (Internal Server Error) when an ES node disconnect occurs this is temporarily returned
+# - 502 (Bad gateway) During shutdown a proxy might return this
+# - 503 (Service unavailable) Proxy returns an unavailable service
+# - 504 (Gateway timed out) Proxy backend timed out
 #
 # NOTE: 409 exceptions are no longer retried. Please set a higher `retry_on_conflict` value if you experience 409 exceptions.
 # It is more performant for Elasticsearch to retry these exceptions than this plugin.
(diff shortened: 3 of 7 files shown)


docs/index.asciidoc
```diff
@@ -55,17 +55,26 @@ the new template is installed.
...
+- 429 (Too many requests) Graceful backoff indication by Elastic
+- 500 (Internal Server Error) when an ES node disconnect occurs this is temporarily returned
+- 502 (Bad gateway) During shutdown a proxy might return this
+- 503 (Service unavailable) Proxy returns an unavailable service
+- 504 (Gateway timed out) Proxy backend timed out
 
-NOTE: 409 exceptions are no longer retried. Please set a higher `retry_on_conflict` value if you experience 409 exceptions.
+The following document errors are handled as follows:
+- 400 and 404 errors are sent to the DLQ (if enabled)
+- 409 errors (conflict) are logged as a warn and dropped. If you'd like these retried
```
jordansissel: incomplete sentence "If you'd like these retried <...missing...>"

docs/index.asciidoc
```diff
@@ -55,17 +55,26 @@ the new template is installed.
...
-The following errors are retried infinitely:
+The following HTTP request errors to the bulk API are retried infinitely:
 
-- Network errors (inability to connect)
-- 429 (Too many requests) and
-- 503 (Service unavailable) errors
+- Network errors (Inability to connect)
+- 404 (Not found) Indicating an upstream proxy has no matching route
+- 429 (Too many requests) Graceful backoff indication by Elastic
+- 500 (Internal Server Error) when an ES node disconnect occurs this is temporarily returned
```
untergeek: There should be a comma after "occurs"

docs/index.asciidoc
```diff
@@ -55,17 +55,26 @@ the new template is installed.
...
+batch as a whole and for individual documents
 
-The following errors are retried infinitely:
+The following HTTP request errors to the bulk API are retried infinitely:
 
-- Network errors (inability to connect)
-- 429 (Too many requests) and
-- 503 (Service unavailable) errors
+- Network errors (Inability to connect)
+- 404 (Not found) Indicating an upstream proxy has no matching route
```
untergeek: "Indicates an upstream..."

docs/index.asciidoc
```diff
@@ -55,17 +55,26 @@ the new template is installed.
...
+- 429 (Too many requests) Graceful backoff indication by Elastic
```
untergeek: "...from Elasticsearch"

docs/index.asciidoc
```diff
@@ -55,17 +55,26 @@ the new template is installed.
...
+- 500 (Internal Server Error) when an ES node disconnect occurs this is temporarily returned
+- 502 (Bad gateway) During shutdown a proxy might return this
```
untergeek: Comma after "shutdown"

docs/index.asciidoc
untergeek: Also, "logged as a warning"

untergeek:
@andrewvc I like that you've differentiated between DOC, DLQ, and REQUEST codes.  Much easier to understand when reading the code.

docs/index.asciidoc
andrewvc: @jordansissel apparently my brain drops sentences after a certain point, I need to work on its retry logic too! :)

docs/index.asciidoc
andrewvc: Why? Are you sure? I don't think that's grammatical.

jordansissel:
@andrewvc I'd like this to be backported to the 7.x branch also (so we can include with LS 5.6)

andrewvc:
@untergeek / @jordansissel this is ready for another pass.

andrewvc:
The more I think about this, the more I think we should retry *all* non-2xx responses from the `/_bulk/` endpoint. It should always return a 200.

andrewvc:
I've updated this PR to treat all non-200 response codes as errors.

docs/index.asciidoc
```diff
@@ -55,17 +55,18 @@ the new template is installed.
 
 ==== Retry Policy
 
-The retry policy has changed significantly in the 2.2.0 release.
+The retry policy has changed significantly in the 8.0.3 release.
```
jsvd: s/8.0.3/8.1.1

docs/index.asciidoc
```diff
@@ -55,17 +55,18 @@ the new template is installed.
...
+request are handled differently than error codes for individual documents.
 
-The following errors are retried infinitely:
+HTTP requests to the bulk API are expected to return a 200 response code. All other response codes are retried indefinitely.
 
-- Network errors (inability to connect)
-- 429 (Too many requests) and
-- 503 (Service unavailable) errors
+The following document errors are handled as follows:
+- 400 and 404 errors are sent to the DLQ (if enabled)
```
jsvd: and what if DLQ is not enabled?

andrewvc:
@jsvd I just pushed fixes for those. Am I good to merge? (I'm not sure if that was a final LGTM)

Issue #433, opened 2016-05-31
`ssl_certificate_verification` will not work without `ssl` being explicitly set


dandrestor:
This configuration:

```
elasticsearch {
  ...
  hosts => [ "https://something" ]
  ssl_certificate_verification => false
}
```

does not work. You need to explicitly add the parameter `ssl => true` for this to work as intended. However, the documentation [states](https://www.elastic.co/guide/en/logstash/current/plugins-outputs-elasticsearch.html#plugins-outputs-elasticsearch-ssl) that if the parameter `ssl` is not set, then whatever protocol is specified in the `hosts` parameter will be used.

I believe the reason is that, in `http_client_builder.rb`, the function `self.setup_ssl` [returns](https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/e18800936a913801a75fc564d4c18b930b15ba86/lib/logstash/outputs/elasticsearch/http_client_builder.rb#L66) if the parameter `ssl` is not set explicitly in the config file.

dandrestor:
Another symptom is the error `PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target` when running with a self-signed certificate without specifying `ssl => true` explicitly.

Petrox:
:+1: this issue is still open.

whm0627:
the issue is in the core at https://github.com/logstash-plugins/logstash-mixin-http_client
I made a pull request, but not sure when that would get merged, cuz the latest tests failed before I made my push like 2 months ago.

for now, if you want a dirty fix, go to your local file: logstash/vendor/bundle/jruby/1.9/gems/logstash-mixin-http_client-5.2.1/lib/logstash/plugin_mixins/http_client.rb
and change it to the updated one:
https://github.com/whm0627/logstash-mixin-http_client/blob/master/lib/logstash/plugin_mixins/http_client.rb

after that, set `ssl_certificate_verification` to false

jonassteinberg1:
currently have same error:

[Manticore::ClientProtocolException] PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target"}

Eschon:
I have a similar issue. I have both `ssl => true` and `ssl_certificate_verification => false` setin my configuration but I still get the error:

```
[...]

PR #641, opened 2017-08-16
#433 Update documentation to reflect on the not working option.

Petrox:
https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/433


andrewvc:
I'm thinking we should, rather, remove this option in the 8.x / 6.0 release since it doesn't work. @jordansissel what are your thoughts on the matter?

I'm not really keen on this option existing, since there's no good reason to enable SSL if you're going to disable cert verification anyway.

Petrox:
There is at least one reason to keep: development and testing.
On localhost one can't have fair signed certs, just selfsigned ones. Still the port and protocol should stay the same as to mimic the prod env in dev.

PR #642, opened 2017-08-17
Do not send 'lang' to ES 6+ for scripted updates

untergeek:
There are no tests for this PR yet. Do not merge.

### DIFF
--- a/lib/logstash/outputs/elasticsearch/http_client.rb
+++ b/lib/logstash/outputs/elasticsearch/http_client.rb
@@ -61,6 +61,7 @@ module LogStash; module Outputs; class ElasticSearch;
       # mutex to prevent requests and sniffing to access the
       # connection pool at the same time
       @bulk_path = @options[:bulk_path]
+      @es_major_version = get_version["number"][0].to_i
     end
     
     def build_url_template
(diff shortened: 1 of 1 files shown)


suyograo:
LGTM.

lib/logstash/outputs/elasticsearch/http_client.rb
```diff
@@ -61,6 +61,7 @@ def initialize(options={})
       # mutex to prevent requests and sniffing to access the
       # connection pool at the same time
       @bulk_path = @options[:bulk_path]
+      @es_major_version = get_version["number"][0].to_i
```
suyograo: Can we make this a function which can be re-used?

untergeek: I was reticent to do so as I don't think we want to keep calling it. 

I mean, the primary function is already [here](https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/master/lib/logstash/outputs/elasticsearch/http_client.rb#L85-L88) and it's also called [here](https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/master/lib/logstash/outputs/elasticsearch/template_manager.rb#L19-L21) effectively the same way.  I just didn't see a reasonable way to get that value once, and keep it, other than an instance variable.  I just can't justify repeatedly calling it in the `update_action_builder` method every time we want to build an update action.

Since the `http_client.rb` doesn't explicitly `require` the `TemplateManager` class (nor can it, I think, without cyclic dependencies), it made more sense to put this here.  I suppose we could remove it from that class and put it here, but that's splitting hairs to me.  I'll do that if you want, though.

suyograo: thanks for the explanation. I get it now.

suyograo:
@untergeek can you check what the test failures are?

untergeek:
Yeah, that's the next step.

andrewvc:
Closing in favor of #688

PR #645, opened 2017-08-21
test parent join field datatype on ES 5.6.0 snapshot

jsvd:


jordansissel:
LGTM

PR #646, opened 2017-08-21
use different tagging scheme in integration tests for es versions

jsvd:
~~This is a WIP PR~~ This PR replaces all the integration test tags related to the elasticsearch version with a single one, and adds a `ESHelper.es_version_satisfies?` method that be used to wrap tests like:

```ruby
if ESHelper.es_version_satisfies?(">= 2")
  describe "Versioned delete", :integration => true do
    # ...
  end
end
```

`ESHelper.es_version_satisfies?` compares the version restriction against the value of the 
 the ES version returned by `bin/elasticsearch -V`, or `ES_VERSION` environment variable if that's not available.

At this point this PR does not balance the integration tests indentation so it's easier to see the changes


spec/es_spec_helper.rb
```diff
@@ -10,6 +10,19 @@ def get_host_port
   def get_client
     Elasticsearch::Client.new(:hosts => [get_host_port])
   end
+
+  def self.es_version_satisfies?(requirement)
+    es_version = ENV['ES_VERSION']
+    result = true
+    if es_version == "master"
```
jsvd: TODO

spec/integration/outputs/parent_spec.rb
```diff
@@ -1,7 +1,8 @@
 require_relative "../../../spec/es_spec_helper"
 require "logstash/outputs/elasticsearch"
 
-context "when using elasticsearch 5.x and before", :integration => true, :version_less_than_equal_to_5x => true do
+if ESHelper.es_version_satisfies?("<= 5.x")
+context "when using elasticsearch 5.x and before", :integration => true do
```
andrewvc: Can we indent this and similar lines?

spec/es_spec_helper.rb
andrewvc: My worry with this strategy is that if there's a bug in this code we might accidentally not run tests that should run and get green builds. 

I'm not sure what the solution here is however.  We should make absolutely sure that the tests we think are running actually are running before we merge.

Once this PR is out of WIP and all tests are passing we should both check the CI logs and manually confirm that.

spec/integration/outputs/parent_spec.rb
jsvd: I just put an explanation on top: I didn't indent these lines on purpose at this WIP stage of the PR, so it's much easier to see the changes without whitespace noise.

ci/run.sh
```diff
@@ -121,32 +126,28 @@ else
...
-          # Still have to support ES versions < 2.x so run tests for those.
-          bundle exec rspec -fd $extra_tag_args --tag ~version_greater_than_equal_to_5x --tag ~version_greater_than_equal_to_2x --tag ~version_greater_than_equal_to_6x --tag ~version_less_than_equal_to_5x $spec_path
+          bundle exec rspec -fd $extra_tag_args --tag es_version:$(es_distribution_version) $spec_path
           ;;
       *)
           download_gradle
           build_es master
           start_es
-          bundle exec rspec -fd $extra_tag_args --tag version_greater_than_equal_to_5x --tag update_tests:painless --tag ~update_tests:groovy --tag ~version_less_than_5x --tag ~version_greater_than_equal_to_6x --tag ~version_less_than_equal_to_5x $spec_path
+          bundle exec rspec -fd $extra_tag_args --tag update_tests:painless --tag ~update_tests:groovy --tag es_version:$(es_distribution_version) $spec_path
```
andrewvc: Very nice change here!

jsvd:
@andrewvc the code has been indented, now we wait for ci 🕦

jsvd:
@andrewvc I'm really confused with the two timeout test failures of ES 5.5.0, I tried running that test like travis does both on my mac and on a new vagrant ubuntu vm , and both executed correctly.

jsvd:
@andrewvc this is ready for a final review!

PR #647, opened 2017-08-21
Retries cherry pick 7.x

andrewvc:
Backport of #640 to 7.x

Thanks for contributing to Logstash! If you haven't already signed our CLA, here's a handy link: https://www.elastic.co/contributor-agreement/

PR #648, opened 2017-08-21
Retries 7x

andrewvc:
Moved #647 to here since I accidentally closed the branch . which is apparently fatal.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ env:
  - INTEGRATION=true ES_VERSION=1.7.6 TEST_DEBUG=true
  - INTEGRATION=true ES_VERSION=2.4.4 TEST_DEBUG=true
  - INTEGRATION=true ES_VERSION=5.5.0 TEST_DEBUG=true
- - INTEGRATION=true ES_VERSION=master TEST_DEBUG=true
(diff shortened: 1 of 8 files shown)


.travis.yml
```diff
@@ -7,7 +7,7 @@ env:
  - INTEGRATION=true ES_VERSION=1.7.6 TEST_DEBUG=true
  - INTEGRATION=true ES_VERSION=2.4.4 TEST_DEBUG=true
  - INTEGRATION=true ES_VERSION=5.5.0 TEST_DEBUG=true
- - INTEGRATION=true ES_VERSION=master TEST_DEBUG=true
+ - INTEGRATION=true ES_VERSION=6.x TEST_DEBUG=true
```
suyograo: @andrewvc is this change intentional?

andrewvc: Yes. I don't think the 7.x branch of this plugin is intended to be tested against Logstash 7.0, which I assume will only work with the 8.x branch.

andrewvc: There's no specific breakage now, but I think this still makes sense

suyograo: @andrewvc this will still fail because there is no `6.x` version. It needs to be `6.0.0-beta1` I think

PR #649, opened 2017-08-23
fix a few bugs in ci/run.sh

jsvd:
fixes a few problems with the `ci/run.sh` script:

- running the script locally breaks because $TRAVIS_BUILD_DIR is null
   * Fix: use only $BUILD_DIR
- we append the local gradle path to the $PATH, but that means the existing gradle always wins
  * Fix: prepend the gradle path to $PATH instead of append

[Tests against elasticsearch master were not running for a long time](https://travis-ci.org/logstash-plugins/logstash-output-elasticsearch/jobs/266824924#L992), this PR will allow the tests to run.

### DIFF
--- a/ci/run.sh
+++ b/ci/run.sh
@@ -1,13 +1,13 @@
 #!/bin/bash
 set -ex
 
-export PATH=$PATH:$TRAVIS_BUILD_DIR/gradle/bin/
-
 # Set the build dir to ./ if not set by travis
 BUILD_DIR=$PWD
 if [[ -z "$TRAVIS_BUILD_DIR" && "$TRAVIS_BUILD_DIR" -ne "" ]]; then
-	BUILD_DIR=$TRAVIS_BUILD_DIR
+  BUILD_DIR=$TRAVIS_BUILD_DIR
 fi
+
+export PATH=$BUILD_DIR/gradle/bin:$PATH
 
 function finish {
   last_result=$?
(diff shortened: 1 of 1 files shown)


suyograo:
LGTM

jordansissel:
I tried to test this, but my VM froze ☹️ 

To test this, I have to set ES_VERSION to something unreleased, right? Like `ES_VERSION=6.0` or `ES_VERSION=master` ?

ci/run.sh
```diff
@@ -1,14 +1,14 @@
 #!/bin/bash
 set -ex
 
-export PATH=$PATH:$TRAVIS_BUILD_DIR/gradle/bin/
```
jordansissel: We need this, kinda, otherwise `gradle` is not guaranteed to be in the path.

It should probably be set after BUILD_DIR is set, and then using this:

```
export PATH="${PATH}:${BUILD_DIR}/gradle/bin"
```

```
+ git clone https://github.com/elastic/elasticsearch.git es_src
Cloning into 'es_src'...
remote: Counting objects: 843097, done.
remote: Compressing objects: 100% (241/241), done.
remote: Total 843097 (delta 155), reused 104 (delta 37), pack-reused 842789
Receiving objects: 100% (843097/843097), 356.54 MiB | 13.60 MiB/s, done.
Resolving deltas: 100% (464622/464622), done.
Checking connectivity... done.
+ cd es_src
+ gradle :distribution:zip:assemble
ci/run.sh: line 92: gradle: command not found
+ finish
```

jsvd: it's still there, I just moved it past the declaration of the BUILD_DIR variable (line 10)

jsvd:
> To test this, I have to set ES_VERSION to something unreleased, right? Like ES_VERSION=6.0 or ES_VERSION=master ?

correct

`INTEGRATION=true ES_VERSION=master TEST_DEBUG=true ci/run.sh` should do it.

you can remove gradle from your path and see that without this patch the command above fails.

ci/run.sh
jordansissel: *facepalm* The failure I saw was because I was testing master, not this PR.

PR #650, opened 2017-08-23
dont allow failures in es 6.0.0-beta1 integration tests

jsvd:
with the upcoming release of 6.0.0, having failing tests is an important concern, so builds should fail if the integration tests fail against es 6.0.0-beta1

this is also related to https://github.com/elastic/logstash/issues/8003


jordansissel:
Double LGTM... I clicked the wrong button when reviewing.

PR #653, opened 2017-08-24
properly detect if DLQ is supported and enabled

### LINKED ISSUE #652: incorrect check for dlq_writer leads to not logging DOC_DLQ_CODES failures
Currently the `@dlq_writer` [is declared in the register method as](https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/f3c4895897cb93303fe58132ff186ea8dbb0c465/lib/logstash/outputs/elasticsearch/common.rb#L21):

```ruby
@dlq_writer = supports_dlq? ? execution_context.dlq_writer : nil
```

In releases of Logstash that support DLQ, the `@dlq_writer` will never be nil, even if DLQ is not enabled, as [a dummy DLQ is set instead of the actual DLQ](https://github.com/elastic/logstash/blob/master/logstash-core/lib/logstash/pipeline.rb#L88-L94).

This means that individual documents in the bulk that returned one of the DOC_DLQ_CODES will reach:

```ruby
          if @dlq_writer
            @dlq_writer.write(action_event, "Could not index event to Elasticsearch. status: #{status}, action: #{action}, response: #{response}")
          else
            @logger.warn "Could not index event to Elasticsearch.", status: status, action: action, response: response
          end
```

And take the first branch of the conditional. If the DLQ is disabled, the event is written to a dummy DLQ, and no warning is logged. This means that 404 and 400 failures are silent: we neither log nor send them to a DLQ.

jsvd:
fixes #652

### DIFF
--- a/lib/logstash/outputs/elasticsearch/common.rb
+++ b/lib/logstash/outputs/elasticsearch/common.rb
@@ -18,7 +18,7 @@ module LogStash; module Outputs; class ElasticSearch;
     def register
       @stopping = Concurrent::AtomicBoolean.new(false)
       # To support BWC, we check if DLQ exists in core (< 5.4). If it doesn't, we use nil to resort to previous behavior.
-      @dlq_writer = supports_dlq? ? execution_context.dlq_writer : nil
+      @dlq_writer = dlq_enabled? ? execution_context.dlq_writer : nil
 
       setup_hosts # properly sets @hosts
       build_client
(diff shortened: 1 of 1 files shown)


lib/logstash/outputs/elasticsearch/common.rb
```diff
@@ -280,8 +280,11 @@ def safe_bulk(actions)
       end
     end
 
-    def supports_dlq?
-      respond_to?(:execution_context) && execution_context.respond_to?(:dlq_writer)
+    def dlq_enabled?
+      # TODO there should be a better way to query if DLQ is enabled
+      # See more in: https://github.com/elastic/logstash/issues/8064
+      respond_to?(:execution_context) && execution_context.respond_to?(:dlq_writer) &&
+        !execution_context.dlq_writer.inner_writer.is_a?(::LogStash::Util::PluginDeadLetterQueueWriter)
```
robbavey: Shouldn't this be `DummyDeadLetterQueueWriter`?

jsvd: 🤦‍♂️

jsvd: changed, good catch

PR #655, opened 2017-08-25
new elasticsearch version handling

jsvd:
* have the connection pool remember the elasticsearch version for each connection
* the http_client can get a list of all these versions
* the sniffing now knows the version of elasticsearch the _nodes/http came from
* the template manager installs the template version of the biggest major version of any connected client

This PR is still not production ready but seems to work: it makes the pool track the info about the es versions of each node.
Then the HttpClient, TemplateManager and others can pull that list and decide what to do.

### DIFF
--- a/lib/logstash/outputs/elasticsearch/template_manager.rb
+++ b/lib/logstash/outputs/elasticsearch/template_manager.rb
@@ -28,7 +28,7 @@ module LogStash; module Outputs; class ElasticSearch
     end
 
     def self.default_template_path(es_major_version)
-      template_version = es_major_version == "1" ? "2" : es_major_version
+      template_version = es_major_version == 1 ? 2 : es_major_version
       default_template_name = "elasticsearch-template-es#{template_version}x.json"
       ::File.expand_path(default_template_name, ::File.dirname(__FILE__))
     end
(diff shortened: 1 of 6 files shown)


jsvd:
@untergeek this is as far as I got with the es version handling.

jsvd:
@untergeek: the tests are now passing (failures are related only to es 5.6/6 compatibility), so feel free to review and merge.

/cc @andrewvc this is the implementation that we talked about

I appreciate if y'all can get this PR into master and 7.x so scripting and other ES6 features can be properly supported, relying on the new availability of node es versions information.

ph:
@untergeek This look what you need for your feature. Thanks @jsvd

lib/logstash/outputs/elasticsearch/template_manager.rb
```diff
@@ -12,12 +12,10 @@ def self.install_template(plugin)
...
     private
-    def self.get_es_version(client)
-      client.get_version
-    end
-
     def self.get_es_major_version(client)
-      get_es_version(client)["number"][0]
+      # get the elasticsearch version of each node in the pool and
+      # pick the biggest major version
+      client.connected_es_versions.uniq.map {|version| version.split(".").first.to_i}.sort.last
```
andrewvc: Ideally this would use `max` instead of `sort` but its really not a big deal.

andrewvc:
LGTM

PR #665, opened 2017-08-31
Selectively disable and amend scripting tests to fix es6 integration tests

robbavey:
Fix integration tests to handle scripting changes in Elasticsearch 6.0:

- Do not run groovy scripting tests unless version is < 6
- Do not run 'file' tests unless version is < 6
- Only specify 'lang:painless' if version is < 6

### DIFF
--- a/spec/integration/outputs/parent_spec.rb
+++ b/spec/integration/outputs/parent_spec.rb
@@ -104,6 +104,12 @@ if ESHelper.es_version_satisfies?(">= 5.6")
             }
           }
         }
+        if ESHelper.es_version_satisfies?('<6')
+          mapping.merge!({
+                 "settings" => {
+                   "mapping.single_type" => true
+                 }})
+        end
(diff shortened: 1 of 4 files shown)


spec/integration/outputs/parent_spec.rb
```diff
@@ -104,6 +104,12 @@
             }
           }
         }
+        if ESHelper.es_version_satisfies?('<6')
+          mapping.merge!({
+                 "settings" => {
+                   "mapping.single_type" => true
```
jordansissel: This setting is only valid in ES 5.6 and above, (it will fail on 5.5 and below, right?)

robbavey: @jordansissel This clause is within an outer satisfies?('>=5.6'), so it won't run on 5.5 and below

jordansissel: Ahh yes. Github's diff wasn't showing that until I expanded. Thanks!

PR #667, opened 2017-09-01
Merge pull request #655 from jsvd/elasticsearch_version

robbavey:
new elasticsearch version handling

### DIFF
--- a/lib/logstash/outputs/elasticsearch/template_manager.rb
+++ b/lib/logstash/outputs/elasticsearch/template_manager.rb
@@ -12,12 +12,10 @@ module LogStash; module Outputs; class ElasticSearch
     end
 
     private
-    def self.get_es_version(client)
-      client.get_version
-    end
-
     def self.get_es_major_version(client)
-      get_es_version(client)["number"][0]
+      # get the elasticsearch version of each node in the pool and
+      # pick the biggest major version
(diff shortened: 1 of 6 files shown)


jordansissel:
Waiting on travis, but LGTM if it is green.

robbavey:
@jordansissel This has the same failures that the original commit had on master. I can either commit this as is and apply the fixes in a subsequent commit/backport (as per master), or I can fold those fixes into this commit to get the tests to pass straight away. WDYT?

jsvd:
@robbavey I we can merge this as is, the failures are related to the lack of compatibility of the plugin with the es 6 wrt scripted updates, so we can work on those fixes in a separate issue/pr

lib/logstash/outputs/elasticsearch/template_manager.rb
```diff
@@ -12,12 +12,10 @@ def self.install_template(plugin)
...
     private
-    def self.get_es_version(client)
-      client.get_version
-    end
-
     def self.get_es_major_version(client)
-      get_es_version(client)["number"][0]
+      # get the elasticsearch version of each node in the pool and
+      # pick the biggest major version
+      client.connected_es_versions.uniq.map {|version| version.split(".").first.to_i}.sort.last
```
jsvd: s/sort.last/max

there was already a PR in master to perform this change, I suggest doing it here too

Issue #617, opened 2017-07-09
Document DLQ behavior in ES output


acchen97:
We should document the scenarios where events will be dead lettered if the DLQ is enabled. Events will be sent to DLQ when they encounter permanent error responses from ES:

- 400 - mapping exception
- 404 - not found

https://www.elastic.co/guide/en/logstash/current/plugins-outputs-elasticsearch.html#_retry_policy

/cc @dedemorton @suyograo

dedemorton:
Yup. I'm planning to add a DLQ policy section, but missed the cutoff for getting it into the gem for 5.5. I also need to add something like this blurb (which Suyog provided):

Mapping errors from Elasticsearch can lead to data loss. Unfortunately mapping errors cannot be handled without human intervention and without looking at the field that caused the mapping mismatch. If DLQ is enabled, the original events causing the mapping errors are stored on a file which can be processed at a later time. Often times, the offending field can be removed and re-indexed to ES. If the DLQ is not enabled, and a mapping error happens, this is logged with a warning and the event dropped

dedemorton:
Fixed via https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/670

PR #670, opened 2017-09-08
[Docs] Document DLQ policy

dedemorton:
Fixes https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/617.

Also fixed some minor formatting issues.

Note that the retry policy already covers how the errors are handled. So I only added the section that suyog and I discussed.


ph:
I've fixed the conflict, just waiting the build to be green.

PR #671, opened 2017-09-11
[Docs] Fix link syntax

dedemorton:
Our convention for forming links changed when we went to a shared file for attributes. The old syntax will result in doc build errors.


dedemorton:
Closing because I'm adding this change to a different PR

PR #674, opened 2017-09-19
7 x/es handling

robbavey:
Merges ES version handling test fixes


robbavey:
Bad merge. Abandoning

PR #637, opened 2017-08-15
Pass a safeuri to BadURIError again

andrewvc:
Partially fixes #630 by reverting behavior where a URI was passed to `BadResponseCodeError` errors.

This bug was introduced in https://github.com/logstash-plugins/logstash-output-elasticsearch/commit/9711709d3b6a258f43b9c9005cc9485bf297d8e7

PR #685, opened 2017-09-22
Fix error in instantiating bad response code exceptions

jsvd:
backport of #637 to 7.x


jsvd:
failure is due to https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/686 not yet being in 7.x (support for join data type)

Issue #638, opened 2017-08-16
Parameters for stored scripts not yet compatible with _update API of ES 6


thehgregor:
It seems that the `_update` API of ES 6.0.0.-beta1 does not accept parameter `lang` anymore:
```
POST _scripts/procdata-test
{
  "script": {
    "lang": "painless",
    "source": "ctx._source.xxx = params.event.get('process_id')"
  }
}
```
```
POST staging-procdata/procdata-order/42/_update
{
  "script": {
    "id": "procdata-test",
    "lang": "painless"
  }
}
```
results in:
`HTTP 400 - illegal_argument_exception: illegally specified <lang> for a stored script`

Parameter `lang` is not accepted anymore and needs to be omitted.

Logstash's elasticsearch-output-plugin does not adhere to this yet, and any attempt to use a scripted update results in the same error as above. Parameter `lang` may not be passed to the ES `_update` API anymore in the case of stored scripts (`script_type => indexed`).

(Note: setting `script_lang` to `null` explicitly in the Logstash configuration does not work, so this must be fixed in the code of the output plugin)

- Version: 6.0.0-beta1
- Operating System: CentOS 7
- Config File (if you have sensitive info, please remove it):
```
output {
  elasticsearch {
    "action" => "update"
    "hosts" => ["elasticsearch:9200"]
    "index" => "staging-procdata"
    "document_id" => "%{process_id}"
    "doc_as_upsert" => true
    "script_var_name" => "event"
    "script" => "procdata-test"
    "script_type" => "indexed"
  }
}
```
- Sample Data: any
- Steps to Reproduce: Create a script in ES and try to use it in a scripted update of a document with above configuration of the logstash-output-elasticsearch plugin.

jordansissel:
Thank you for reporting. We will investigate as time permits.

untergeek:
Hmm. No pioneer tag in this repo, but I think this is worthy

andrewvc:
I'm thinking that the solution here is going to involve actual version sniffing on the live cluster. We don't want to be requiring people to explicitly declare a null option, or declare their ES version etc.

jsvd:
we already [fetch the version of the node we hit first when in the template manager](https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/e7259de52f88643cc3311ce73d4dfa4663041a87/lib/logstash/outputs/elasticsearch/template_manager.rb#L15), this should probably be pulled out of that class

[...]

PR #688, opened 2017-09-26
fix stored painless script integration testing for es 6 and above

jsvd:
Also adds documentation explaining that for ES 6 it's necessary to set the `script_lang` parameter to empty string (`""`).

fixes https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/638

replaces #642

### DIFF
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -477,7 +477,8 @@ Set script name for scripted update mode
   * Value type is <<string,string>>
   * Default value is `"painless"`
 
-Set the language of the used script. If not set, this defaults to painless in ES 5.0
+Set the language of the used script. If not set, this defaults to painless in ES 5.0.
+When using indexed (stored) scripts on Elasticsearch 6 and higher, you must set this parameter to `""` (empty string).
 
 [id="plugins-{type}s-{plugin}-script_type"]
 ===== `script_type` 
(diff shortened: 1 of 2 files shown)


ph:
I was looking at the related issues, would it be hard to do that change automatically **or** yield an appropriate log message when you are using `script_lang` set to `painless` when targetting a ES 6.0 cluster.

jsvd:
there are many nuances to calling a cluster "an ES 6.0 cluster". for example, a 5.6 cluster upgrading live to 6.0 will not enable specific 6.0 features and break compatibility until all nodes are 6.0, so we must get the versions from all nodes (not just the list of hosts in the es output, and not just the sniffed hosts).

this is such a specific feature and since we can document it clearly, I'm tempted to avoid adding magic to it.

ph:
@jsvd valid point about the rolling upgrades, I am just concerned that people wont see this change, we need to bring it in the breaking change documentation?

jsvd:
Interesting question..I'm not sure if this is considered a breaking change. A breaking change in ES 5->6 means that users upgrading to es 6 must tweak the config to make it work.

In the same way, if you're using `script_type => file` that will not work in ES 6.0, so you'll see the error that elasticsearch returns and change things accordingly

ph:
@jsvd Right version changes aren't easy. Oh, we will need a version bump to make sure the doc generator pickup the docs changes.

jsvd:
merged also to 7.x -> 3ecb7b1

PR #691, opened 2017-09-29
Doctype deprecation

### LINKED ISSUE #669: Document Type handling with ES6.0
Elasticsearch 6.0 is [removing document types](https://www.elastic.co/guide/en/elasticsearch/reference/6.0/removal-of-types.html)

Beats 6.0 will begin setting `type` to `doc` in the future.

We haven't changed Logstash behavior at all yet.

My thoughts:

1. Currently this plugin will set the `document_type` to the value of the `type` event field. This behavior should be disabled for ES6+ clusters as well as support for sprintf interpolation.
2. We should deprecate `document_type`
3. We should switch our default `logs` type for that setting to `doc` for consistency with beats.

Thoughts?

andrewvc:
Resolves #669 by deprecating the `doc_type` option and moving the default type name from `logs` to `doc`.

We want to get this in before 6.0.0 GA otherwise it will have to wait for LS 7.0.0

The really nice thing about moving from `logs` to `doc`, aside from beats integration, is that it better reflects the generic nature of Logstash.

We will also need to add this to breaking changes in Logstash core.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+### 9.0.0
+  - Change default document type to 'doc' from 'logs' to align with beats and reflect the generic nature of logstash.
+  - Deprecate 'document_type' option
+
 ### 8.2.2
   - Use `#response_body` instead of `#body` when debugging response from the server #679
 
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -236,8 +236,11 @@ Elasticsearch with the same ID.
 
   * Value type is <<string,string>>
   * There is no default value for this setting.
+  * This option is deprecated
 
-The document type to write events to. Generally you should try to write only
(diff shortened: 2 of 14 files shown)


andrewvc:
@dedemorton this potentially has a lot of implications for docs on the site. Would you be able to investigate the impact there?

andrewvc:
After this is merged we'll need to make sure to update the 6.0.0-rc2 lockfile with the latest version.

jordansissel:
This is kind of tricky because `document_type` is not deprecated in Elasticsearch 5.x and Logstash 6.0 will talk to ES 5.x clusters.

I know there is urgency to land this, but please pause and evaluate how we can make this work *and* not annoy users who are using LS 6 to talk to ES 5 clusters.

andrewvc:
@jordansissel that's a good point. I'm open to not deprecating document type given that point.

@ph WDYT?

jordansissel:
LGTM with your revised `:deprecated` warning.

Issue #612, opened 2017-06-16
Bulk Request Fails in Logstash


tamiliniyaan:
I am getting the following message while start a logstash .

Version ES and Logstash - 5.1.1 
OS : CentOS

output plugin as ,
elasticsearch {
                hosts =>["10.4.1.70:9299"]
                index => "%{[@metadata][INDEX_NAME]}"
                document_type => "logs"
        }

[2017-06-16T05:31:44,494][ERROR][logstash.outputs.elasticsearch] Got a bad response code from server, but this code is not considered retryable. Request will be dropped {:code=>400, :response_body=>"No handler found for uri [/_template/logstash/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk/_bulk] and method [POST]"}

It will try for  the logstash template.

nanjj:
similar problem here:
```
[...]

PR #692, opened 2017-10-04
Don't try to sanitize the URL as there is no such method

amitavmohanty01:
The rescue block for `::LogStash::Outputs::ElasticSearch::HttpClient::Pool::BadResponseCodeError` was throwing an exception while logging and the original exception which should be logged is being lost.

https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/612 is caused because of this.


amitavmohanty01:
I have signed the CLA and I got a copy in my email. Let me know if I need to do anything else for the system to pick up that.

amitavmohanty01:
The CI failures were with build and it seems those builds have failed with a requirement of a gem. I would like to hear the suggestions of maintainers on the issue.

Issue #303, opened 2015-11-11
Remove the master only nodes from the hosts list auomaticly when using sniffing 


ESamir:
In 2.0 the sniffing  setting asks Elasticsearch for the list of all cluster nodes and adds them to the hosts list. This will return ALL nodes with HTTP enabled (including master nodes!). 
it will be safer if we remove the master only nodes automatically.

andrewvc:
You're 100% that we should be doing this :) . We plan on doing this very soon.

astefan:
+1 on this. Sending indexing requests to the master node can cause serious stability issues. Especially given that, usually, a master only node doesn't need as many hardware resources as a data node and it is usually a low powered node.

blavoie:
Is disabling http port (9200) on master nodes a good workaround, and also good practice?
Let them only communicate for cluster stuff...
https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-http.html#_disable_http

tomsommer:
+1

jrbeilke:
@andrewvc would you mind taking a look at my open PR to implement improved sniffing of elasticsearch data nodes? I believe that should resolve this issue and #599

andrewvc:
Fixed in #694

Issue #599, opened 2017-05-05
Dealing with proposed http.enabled changes on ES's side


ppf2:
This is to track the outcome of an active [ES ticket](https://github.com/elastic/elasticsearch/issues/12792
) intended to remove http.enabled from the product. 

This will directly affect how we do [sniffing](https://www.elastic.co/guide/en/logstash/current/plugins-outputs-elasticsearch.html#plugins-outputs-elasticsearch-sniffing) today in the LS ES output.  

We already have users today using a combination of sniffing and http.enabled: false (as recommended in the LS documentation) on the ES side to influence the ES output so that it doesn't send the events to the master nodes or other specific nodes of interest (eg. warm nodes).

If ES removes the ability to disable http, it will be breaking for LS ES output's sniffing feature.

jordansissel:
> it will be breaking for LS ES output's sniffing feature

Sounds like we need to update the documentation to indicate that the `http.enabled` recommendation only works with Elasticsearch versions 5.x and older, not 6.0 or beyond.

jasontedor:
Disabling HTTP on master-eligible nodes is not something I would recommend to begin with, sometimes the cluster is misbehaving and you need to send requests directly to a master. If the master is not enabled for HTTP then it can not receive said requests, this is undesirable. I think instead the sniffing should be enhanced to exclude master nodes if needed. 

An option for now is to bind HTTP on master-eligible nodes to localhost only, but this has problems too:
 - without looking at the code, I would guess this means the output would turn around and try to send a request its localhost (because it will see localhost addresses in the sniffing)
 - it doesn't prevent requests from hitting a master-eligible if it's running on the same box (not a configuration I would recommend)
 - it means a user has to SSH to a master-eligible node to send requests directly to a master

jordansissel:
@jasontedor I agree with your concerns about disabling http. 

I tried to find some history. Sniffing was added to our http client in 6e95c72946c8002400bc17a212cec2df74a1de77 added the following text to the existing `sniffing` feature:

> Will return ALL nodes with HTTP enabled (including master nodes!). If you use this with master nodes, you probably want to disable HTTP on them by setting `http.enabled` to false in their elasticsearch.yml.

Relevant tickets: #240, #216, #150.

I was not able to find history behind the decision to recommend `http.enabled: false` on masters.

----

I am +1 on removing the `http.enabled: false` recommendation from our docs, and we can also start work on improving sniffing for users who want such feature.

[...]

Issue #596, opened 2017-05-02
nodes HTTP bound to localhost should not be returned by the sniffer


zumo64:
- Version: 5.x

If you have a coordinating node bound to 127.0.0.1 in a cluster (ex: responding to Kibana queries) , that node will be included in the list of nodes sniffed by the  ES output even if 127.0.0.1 is unreachable.

andrewvc:
@zumo64 this is a tricky thing. A user could have that node running on the same box as Logstash, there's really no way to know that.

That said, this should work today. The sniffer will find the node, but will never connect to it (though it may connect to some other ES node on localhost, that hopefully is from the same cluster).

This is making me think that we should have an invariant when sniffing that makes sure the sniffed node is from the proper cluster.

jordansissel:
What's the negative impact? Wouldn't an unreachable address be detected as
unreachable and would always be marked as down in our connection pool?

andrewvc:
One negative might be false alarms in the logs from the always down host. I
think at some point it would make sense to allow for more refined criteria.
You can set custom metadata for rack awareness, it'd be great to add that
to logstash.

jordansissel:
@andrewvc fair point, so maybe we should do this:

Proposal: 

1) Expose metrics on the health of all known Elasticsearch nodes
2) Stop logging intermittent connection problems (reduce noise)

```
{
  "nodes": [
    {
      "address": "1.2.3.4:9200",
      "nodename": "whatever", // The elasticsearch node name or id?
      "state": "good|bad|ignored"
      "notes": "<reason for state>"
      "metrics": {
        // Metrics for bytes, events, requests, and connection counts.
        "sent_bytes": 12345,
        "sent_events": 1000,
        "sent_requests": 20,
        "connection_failures": 1,
        ...
      }
    }
  }
}
```

The "notes" could show reason for the given state:

* bad: network error (connection refused, connect timeout, request timeout, etc)
* ignored: found by sniffing, but excluded because of a specific sniffing configuration such as "only send data to nodes with tag Foo"

Issue #438, opened 2016-06-14
Allow sniffing to only return nodes with specific ES node attributes


ppf2:
Hot/cold architecture is common in ES/LS deployments where hot nodes will be tagged with a specific attribute.  It will be helpful to allow the end user to specify a node attribute tag to match as part of sniffing, eg. configure sniffing but tell it to only use nodes with node_type:ssd for example.

gmoskovicz:
@acchen97 i noticed the P3 tag. I guess that the output plugin will need to ask for the cluster state to sniff specific nodes. Is it expected to do this anytime soon?

jordansissel:
What is the expected result of this feature? Do we have evidence the
feature will have the desired result?

jordansissel:
Specifically, a tag isn't the same as a routing id, so the node receiving
the request may not even have the index or the right shard(s).

I am not against this, but it's abnormal (Elasticsearch already supports
this feature, so why would a client need it?), so I want to make sure we
know what result is expected and if we even think the proposed feature can
achieve the desired result.

ppf2:
This is more of a request on the config management side of things, so users can simply specify the same tags they set up on the ES side and it will automatically set the hosts array to those nodes instead of having to manage a list of host names (that can change over time) directly in the LS config.  Current workaround is probably to handle this outside of the Elastic Stack via some external config management tool :)

jordansissel:
> This is more of a request on the config management side of things

I agree on the alignment, but I am missing what is the proposed benefit of this. Let's get closer to the problem. Why would a user want to send _bulk requests only to a certain set of data nodes? What is the expected benefit?

allenmchan:
Yes a tag is not a routing ID. So in theory the request will need to be re-routed to another node that has the active shard.

In a hot / warm / cold architecture, you do want all indexing requests to hit the hot tier (SSDs or some form of faster storage). 
So this enhancement is asking that the sniffing return all the nodes that are in the hot tier (which elasticsearch documentation says we can setup tiers using tags). Without this enhancement, i would not be able to leverage the wonderful feature of sniffing. The sniffing would return all nodes by default and bulk requests would hit all nodes therefore violate the teaching that i have to not send bulk indexing to warm or cold nodes.

Hope that clarifies my ask at least.

jordansissel:
> violate the teaching that i have to not send bulk indexing to warm or cold nodes.

What teaching is this?

[...]

PR #694, opened 2017-10-16
feature-improvedsniffing - When sniffing exclude master-only nodes

jrbeilke:
Takes care of #303 and #599 by filtering out nodes that only have the 'master' role enabled when sniffing (ie. master-only nodes).

This is similar to what the other Elasticsearch client libraries do by default.


andrewvc:
@jrbeilke can you please add tests for this?

docs/index.asciidoc
```diff
@@ -518,11 +518,8 @@ if enabled, script is in charge of creating non-existent document (scripted upda
   * Value type is <<boolean,boolean>>
   * Default value is `false`
 
-This setting asks Elasticsearch for the list of all cluster nodes and adds them to the hosts list.
-Note: This will return ALL nodes with HTTP enabled (including master nodes!). If you use
-this with master nodes, you probably want to disable HTTP on them by setting
-`http.enabled` to false in their elasticsearch.yml. You can either use the `sniffing` option or
-manually enter multiple Elasticsearch hosts using the `hosts` parameter.
+This setting asks Elasticsearch for the list of all cluster nodes and adds the data serving nodes to the hosts list.
```
andrewvc: This documentation should state the difference between 5.x and 2.x, where no such sniffing will occur.

jrbeilke: sniff_2x_1x() already filters for nodes with the data attribute so there should be no difference after this change with 5.x vs 2.x now as far as filtering out nodes.

         # Only connect to nodes that serve data
         # this will skip connecting to client, tribe, and master only nodes
         # Note that if 'attributes' is NOT set, then that's just a regular node
         # with master + data + client enabled, so we allow that
         attributes = info['attributes']
         next if attributes && attributes['data'] == 'false'

jrbeilke: Hmm scratch that. If the nodes have the 'data' attribute set then sniffing will only include those nodes, but by default ES 2.x does not include that attribute so sniffing would fallback to checking for http.

    {
      "cluster_name" : "elasticsearch",
      "nodes" : {
        "JS_X6zdyQ2-K9AWBGbdbhw" : {
          "name" : "Mockingbird",
          "transport_address" : "127.0.0.1:9300",
          "host" : "127.0.0.1",
          "ip" : "127.0.0.1",
          "version" : "2.4.6",
          "build" : "5376dca",
          "http_address" : "127.0.0.1:9200",
          "http" : {
            "bound_address" : [ "[::1]:9200", "127.0.0.1:9200" ],
            "publish_address" : "127.0.0.1:9200",
            "max_content_length_in_bytes" : 104857600
          }
        }
      }
    }

jrbeilke:
Tests should be all set now, although it looks like one of the Travis CI jobs timed out in the last build.

Successfully built and validated everything with Travis CI against my fork here - https://travis-ci.org/jrbeilke/logstash-output-elasticsearch

andrewvc:
Great work @jrbeilke . The only question for me is whether we consider this a breaking change, which determines whether the released plugin gets bundled with a 6.x Logstash.

I'm inclined to say that this is not a breaking change, since there's no reason to be querying master only nodes in the first place.

WDYT @jsvd / @jordansissel ? I could use a second opinion on this

jrbeilke:
Any updates @jsvd @jordansissel ?

andrewvc:
Sorry for the delay @jrbeilke , we'll get back to you on this soon.

jordansissel:
I'm ok considering it a bug/mistake that we send _bulk requests to master nodes and that the correct behavior is to remove master nodes from the list of sniffed results.

That said, this PR is a whitelist allowing only `data` nodes, not a blacklist rejecting `master-eligible`-only roles. This means, I think, that coordinating nodes (formerly known as "client nodes") would be rejected from this list as well. Is this intended?

andrewvc:
That is a great point @jordansissel , we definitely don't want to exclude coordinating nodes with no way to disable that behavior. 

Apologies @jrbeilke , but I think the correct behavior we need is what Jordan described, a blacklist, not a whitelist.

jordansissel:
I'm ok with whatever (blacklist/whitelist), but I think the set of single-role nodes allowed are `data` and `coordinating` .

Based on the ES docs, a node is a coordinating node if all roles are disabled:

from: https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-node.html
> Every node is implicitly a coordinating node. This means that a node that has all three node.master, node.data and node.ingest set to false will only act as a coordinating node, which cannot be disabled

So accept all data nodes and all nodes with no roles (aka "coordinating" node).. I think once done this PR should be good to go. @jrbeilke thoughts?

jordansissel:
To stress, it is mainly intuition, not data, that makes me wonder about coordinating nodes. If you think that this PR can merge as-is without coordinating node support, I think this is probably ok. I just want some agreement (as said above) on excluding coordinating nodes so that I have a good story to tell if/when someone asks why coordinating nodes are excluded. I don't want to scope-creep this PR, so pushing back on coordinating node support is an OK choice -- let me know what you think :)

jordansissel:
@andrewvc I'm good to merge this once @jrbeilke gives the thumbs-up on skipping-or-not coordinating nodes.

jrbeilke:
Sigh... my goal when I opened this PR was to get logstash-output-elasticsearch up to parity with the ES transport client's sniffing (only returns nodes with data role), but with the upcoming deprecation of that client maybe that's not the best idea.

I did a quick comparison of the different ES clients, but this is quite a sticky wicket as other clients handle sniffing differently and there doesn't seem to be one clear standard:

| ES client | Default Sniffing | Configurable Sniffing? |
| ---------- | ------------------ | ------------------------- |
| elasticesearch-py | filter out master-only nodes | yes |
| net-api | filter out master-only nodes | yes |
| elasticsearch-js | all http-enabled nodes | yes |
| elasticsearch-perl | all http-enabled nodes | yes |
| elasticsearch-ruby | all http-enabled nodes | no |
| java-rest | all http-enabled nodes | no |
| logstash-output-elasticsearch | all http-enabled nodes | no |
| java transport | only nodes with data role | no |

Seems like what we're after is a sane default for logstash sniffing elasticsearch, which should probably just filter out the master-only nodes for now and include a config for custom sniffing in a future PR.

Should I refactor my code in this PR to filter out master-only nodes or should we close this and open a fresh PR instead?

andrewvc:
Thanks for the patience and the research. , it's up to you on a new PR or
not.

jrbeilke:
Alrighty, refactored my code and this PR to now exclude master-only nodes instead of including data nodes. Looks like there were two test failures in Travis (for ES 6.2) but it appears to be unrelated to this change (404 when fetching elasticsearch-6.2.0-SNAPSHOT.tar.gz).

This PR should still resolve #303 and #599 

There are also #438 and #596 which would need to be covered by a future PR to support custom configs for sniffing.

spec/fixtures/_nodes/5x_and_above.json
```diff
@@ -0,0 +1,62 @@
...
+  "nodes" : {
+    "Ur_68iBvTlm7Xr1HgSsh6w" : {
+      "name" : "dev-es-master01",
+      "transport_address" : "http://localhost:9200",
+      "host" : "localhost",
+      "ip" : "127.0.0.1",
+      "version" : "5.5.1",
+      "build_hash" : "19c13d0",
+      "roles" : [
+        "master"
```
andrewvc: Can we add an additional test case here for a node with `['master', 'data']`? That tests sniffing a default elasticsearch node with no master only stuff.

spec/integration/outputs/sniffer_spec.rb
```diff
@@ -10,38 +10,96 @@
...
+        allow(subject).to receive(:perform_request).and_return([nil, { version: "5.0" }, response_double])
+        subject.start
+      end
+
+      context "with three nodes with one master-only node" do
+        it "should execute a sniff without error" do
+          expect { subject.check_sniff }.not_to raise_error
+        end
+
+        it "should return two sniff URLs" do
```
andrewvc: If we check the exact URLs later this test doesn't really do much.

I don't mind leaving it here, but it doesn't add much either.

spec/fixtures/_nodes/5x_and_above.json
jrbeilke: Updated spec to include testing master-only vs master + data node and also fixed the travis config for ES 6.2

andrewvc:
@jrbeilke beautiful! I'm going to merge this and release a new version later today :).

Thanks for the excellent work here @jrbeilke

Issue #708, opened 2017-11-21
Document type issue with out of the box experience with FB


ph:
Currently  if you are using FB or any other beat the client will send a custom `type` field with every events, when you are using the ES output with the default settings, the behavior is to take the value of the event "type" field as the document type instead of the default "doc".

The problem with the type changes in es 6.0 this will make the logstash output throw theses errors and not persist any events to ES.

```
[2017-11-21T14:26:01,991][WARN ][logstash.outputs.elasticsearch] Could not index event to Elasticsearch. {:status=>400, :action=>["index", {:_id=>nil, :_index=>"packetbeat-6.0.0-2017.11.21", :_type=>"flow", :_routing=>nil}, #<LogStash::Event:0x4af7bf6>], :response=>{"index"=>{"_index"=>"packetbeat-6.0.0-2017.11.21", "_type"=>"flow", "_id"=>"jEL43l8Bp0AXkoCMsC5q", "status"=>400, "error"=>{"type"=>"illegal_argument_exception", "reason"=>"Rejecting mapping update to [packetbeat-6.0.0-2017.11.21] as the final mapping would have more than 1 type: [icmp, flow]"}}}}
```

**Configuration used:** 

```
input {
  beats {
    port => 5044
  }
}
filter {
}
output {
  elasticsearch {
    hosts => ["http://localhost:9200"]
    index => "%{[@metadata][beat]}-%{[@metadata][version]}-%{+YYYY.MM.dd}"
    user => "elastic"
    password => "changeme"
  }
}
```

Not sure what would be the best way to handle this, seems like this behavior could break users configuration if they rely on type. Maybe we should only rely on either _fieldref_ or the "doc" default value and not use the _event.get("type")_

Any ideas @andrewvc @jsvd ?

gingerwizard:
or winlogbeat or metricbeat i suspect - they all produce multiple values for `type`

ph:
@gingerwizard yes, I've changed the wording to be more generic.

jordansissel:
I think for now users will have to set `document_type => "doc"` in their config. Going forward, maybe detecting ES version for a connection and ignoring the `document_type` setting for these connections if ES 6.0+ is detected? It's tricky because we can't just default to "log" if the user already has "foo" as the one-and-only-type in ES (I haven't looked at the code if we try this already, or how complex it would be)

Thoughts on a migration strategy for users?

ph:
[...]

PR #709, opened 2017-11-21
WIP Only use the type field for the event for ES versions < 6.0.

andrewvc:
Since 6.0 only supports a single type this behavior can be problematic as
multiple types will break ES.

Fixes https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/708

This is a WIP, it still needs tests

### DIFF
--- a/lib/logstash/outputs/elasticsearch/common.rb
+++ b/lib/logstash/outputs/elasticsearch/common.rb
@@ -22,6 +22,14 @@ module LogStash; module Outputs; class ElasticSearch;
 
       setup_hosts # properly sets @hosts
       build_client
+
+      # Wait until at least one host connects and we have a version
+      until maximum_seen_major_version
+        @logger.info("ES Output starting, determining ES version...")
+        sleep 1
+      end
+      @logger.info("ES Output version determined", :es_version => maximum_seen_major_version)
+
       install_template
       check_action_validity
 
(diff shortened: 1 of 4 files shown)


lib/logstash/outputs/elasticsearch/common.rb
```diff
@@ -199,12 +211,17 @@ def event_action_params(event)
...
     # Determine the correct value for the 'type' field for the given event
+    DEFAULT_EVENT_TYPE="doc".freeze
     def get_event_type(event)
       # Set the 'type' value for the index.
       type = if @document_type
                event.sprintf(@document_type)
              else
-               event.get("type") || "doc"
+               if maximum_seen_major_version > 6
+                 event.get("type") || DEFAULT_EVENT_TYPE
```
acchen97: I'm trying to understand this code, and for ES 6.0 it seems like it to still try to set the type based on the event type? Can't each event still have multiple different types?

andrewvc: Ah, good call, the logic is wrong, that should be `<= 6`

andrewvc: Errr, `< 6` I mean

andrewvc:
Superceeded by #712

PR #711, opened 2017-11-23
allow failures in es master

jsvd:
due to elasticsearch master containing breaking changes as it moves to 7.0.0,
it's not required that the tests pass for this particular scenario until
it reaches closer to 7.0.0-alpha1


original-brownbear:
@jsvd LGTM :)

PR #712, opened 2017-11-29
Only use the type field for the event for ES versions < 6.0

jsvd:
aims to replace https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/709

This PR ensures the event field "type" isn't used as a document_type if elasticsearch cluster is on 6.0

### DIFF
--- a/lib/logstash/outputs/elasticsearch/http_client/pool.rb
+++ b/lib/logstash/outputs/elasticsearch/http_client/pool.rb
@@ -113,6 +113,12 @@ module LogStash; module Outputs; class ElasticSearch; class HttpClient;
       end
     end
 
+    def maximum_seen_major_version
+      @state_mutex.synchronize do
+        @maximum_seen_major_version
+      end
+    end
+
     def urls
       url_info.keys
     end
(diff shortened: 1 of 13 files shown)


lib/logstash/outputs/elasticsearch/http_client/pool.rb
```diff
@@ -245,6 +245,14 @@ def healthcheck!
           es_version = get_es_version(url)
           @state_mutex.synchronize do
             meta[:version] = es_version
+            major = major_version(es_version)
+            if !@maximum_seen_major_version
+              @logger.info("ES Output version determined", :es_version => @maximum_seen_major_version)
+              set_new_major_version(major)
+            elsif major > @maximum_seen_major_version
+              @logger.warn("Detected a node with a higher major version than previously observed. This could be the result of an elasticsearch cluster upgrade.", :previous_major => @maximum_seen_major_version, :new_major => major, :node_url => url)
```
andrewvc: <3 the extra detail here

spec/unit/outputs/elasticsearch_spec.rb
```diff
@@ -2,7 +2,31 @@
...
+  let(:options) { {} }
+  let(:maximum_seen_major_version) { rand(100) }
+
+  let(:do_register) { true }
+
+  before(:each) do
+    if do_register
+      subject.register
+
+      # Rspec mocks can't handle background threads, so... we can't use any
```
andrewvc: I know! One more reason to refactor into java.

jsvd:
I've also added a tiny documentation change to explain this difference of behaviour in es 6.x and above.
I'm merging and publishing since review as positive pending on version bump (which has been done).

Issue #701, opened 2017-11-02
add support for elasticsearch 7.x


jsvd:
- [ ] template for 7.x
- [ ] ???

PR #713, opened 2017-11-30
add template for elasticsearch 7.x

jsvd:
improves support for elasticsearch 7.x, see more in https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/701

### DIFF
--- a/lib/logstash/outputs/elasticsearch/elasticsearch-template-es7x.json
+++ b/lib/logstash/outputs/elasticsearch/elasticsearch-template-es7x.json
@@ -1,0 +1,45 @@
+{
+  "template" : "logstash-*",
+  "version" : 60001,
+  "settings" : {
+    "index.refresh_interval" : "5s"
+  },
+  "mappings" : {
+    "doc" : {
+      "dynamic_templates" : [ {
+        "message_field" : {
+          "path_match" : "message",
+          "match_mapping_type" : "string",
+          "mapping" : {
+            "type" : "text",
+            "norms" : false
+          }
+        }
+      }, {
+        "string_fields" : {
+          "match" : "*",
+          "match_mapping_type" : "string",
+          "mapping" : {
+            "type" : "text", "norms" : false,
+            "fields" : {
+              "keyword" : { "type": "keyword", "ignore_above": 256 }
+            }
+          }
+        }
+      } ],
+      "properties" : {
+        "@timestamp": { "type": "date"},
+        "@version": { "type": "keyword"},
+        "geoip"  : {
+          "dynamic": true,
+          "properties" : {
+            "ip": { "type": "ip" },
+            "location" : { "type" : "geo_point" },
+            "latitude" : { "type" : "half_float" },
+            "longitude" : { "type" : "half_float" }
+          }
+        }
+      }
+    }
+  }
+}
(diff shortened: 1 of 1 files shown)


lib/logstash/outputs/elasticsearch/elasticsearch-template-es7x.json
```diff
@@ -0,0 +1,45 @@
+{
+  "template" : "logstash-*",
+  "version" : 60001,
+  "settings" : {
+    "index.refresh_interval" : "5s"
+  },
+  "mappings" : {
+    "doc" : {
```
jordansissel: Maybe we should use `_doc` ? Per [elasticsearch master docs](https://www.elastic.co/guide/en/elasticsearch/reference/master/removal-of-types.html): 

> 7.x indices which don’t have an explicit type will use the dummy type name _doc. 

Also, regarding 7.x:

> The GET|PUT _mapping APIs support a query string parameter (include_type_name) which indicates whether the body should include a layer for the type name. It defaults to true

Should we use `include_type_name=false` and omit the doc type entirely?

jsvd: > Maybe we should use _doc ? Per elasticsearch master docs:

indeed, done 👍 

> Should we use include_type_name=false and omit the doc type entirely?

Good question. it will require a deeper change to the TemplateManager to send the `include_type_name`, so I'd consider out of scope for this PR.

yaauie:
There are a bunch of docs (both in source and in asciidoc) that call out changes to the 5.x template; these should likely get additional and/or updated callouts.

lib/logstash/outputs/elasticsearch/elasticsearch-template-es7x.json
jpountz: Actually I think we would reject `_doc` because it starts with an underscore, which could collide with endpoints. (think eg. of a type that would be called `_search`)

jsvd: @jpountz does this mean that the reference to `_doc` in the [schedule for removing mapping types](https://www.elastic.co/guide/en/elasticsearch/reference/master/removal-of-types.html#_schedule_for_removal_of_mapping_types) isn't correct and should be just `doc`?

jpountz: We had not considered that some users would want to start opting in for this behaviour prior to 7.0. We discussed it yesterday and agreed to allow `_doc` as a type name in 6.x in order to allow for a smoother transition to 7.0. I will ping you when this change is in so that this PR can be merged. https://github.com/elastic/elasticsearch/issues/27750

jsvd: @jpountz we'll only be pushing this template for elasticsearch 7.x (not 6.x or before) so we're not blocked on https://github.com/elastic/elasticsearch/issues/27750, I was just wondering what should the es 7.x template for logstash should look like according to the best practices of elasticsearch 7.x

jpountz: We will probably support both `_doc` and no type at all eventually, but master is not ready for it yet. If you are fine with changing it again in a couple weeks/months, we'd ideally set the type to `doc` now and then switch to `_doc` or even remove the type entirely from the mapping when master is ready?

jpountz:
FYI we now accept `_doc` as a type so this PR can be merged: https://github.com/elastic/elasticsearch/pull/27816.

PR #717, opened 2017-12-26
use gradle 4.3 because es 7 needs it

jsvd:
see more in https://github.com/elastic/elasticsearch/pull/27885

an example of failure: https://travis-ci.org/logstash-plugins/logstash-output-elasticsearch/jobs/321736196#L1605


original-brownbear:
@jsvd LGTM :)

PR #727, opened 2018-01-24
ssl_certificate_verification description udated

eedugon:
Just if you consider a useful comment... feel free to discard the PR if you prefer not showing this here.

Added a note/warning about not all use cases working fine when using this feature, to avoid people losing the time troubleshooting. There are reported bugs about this, and it looks there's no plan to fix them as it's actually not recommended to use this feature.

### DIFF
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -563,6 +563,8 @@ Option to validate the server's certificate. Disabling this severely compromises
 For more information on disabling certificate verification please read
 https://www.cs.utexas.edu/~shmat/shmat_ccs12.pdf
 
+Note: Not all use cases of skipping certificate verification are supported. There are known issues in some cases. For example, if the certificate of the server is a wildcard + self-signed certificate, skipping the CA verification might not work and the cacert option in the config is needed.
+
 [id="plugins-{type}s-{plugin}-template"]
 ===== `template` 
 


jakelandis:
@eedugon - would you mind logging a bug for this , or adding your specific use case to an existing bug.  Apologies this isn't working as intended, but I would prefer to fix the bug rather then add it to the documentation.

eedugon:
Hi @jakelandis ,

I would say this issue reflects the problem:
https://github.com/logstash-plugins/logstash-output-http/issues/21
(it's an old issue but still applicable in logstash 5.x).

This one is probably also related:
https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/433

Do you prefer if I write a new issue?

I don't feel there's a plan to fix this, that's the reason I was proposing to update the documentation so certificate verification skipping is used with caution, because there are some scenarios where this does not work fully.

Regards!
Edu

jakelandis:
@eedugon - thanks for the follow up. There are some inconsistencies across plugin TLS support ranging from naming to configuration to functionality to bugs like this.  The Logstash team is aware of this and have internally discussed ways to address. 

I believe that you and many other members of our community have had a bad time with TLS support, and that is a bug we need to fix. Thanks for re-raising our attention to this, and I continue internal discussions for how we can do better.

PR #728, opened 2018-01-24
convert data type links in docs

jsvd:


dedemorton:
LGTM, but I would like to do a full doc build before we mass publish.

PR #730, opened 2018-02-01
Avoid getting stuck on infinite loop

hasnat:
If we get 400 which could be part of DLQ, plugin here gets stuck in loop indefinitely on retry
e.g. https://discuss.elastic.co/t/logstash-5-6-1-inifinite-loop-encountered-a-retryable-error/101753

### DIFF
--- a/lib/logstash/outputs/elasticsearch/common.rb
+++ b/lib/logstash/outputs/elasticsearch/common.rb
@@ -272,7 +272,7 @@ module LogStash; module Outputs; class ElasticSearch;
         end
 
         sleep_interval = sleep_for_interval(sleep_interval)
-        retry
+        retry unless @stopping.true?
       rescue => e
         # Stuff that should never happen
         # For all other errors print out full connection issues


jordansissel:
This will cause data loss when `queue.type: memory`. For these scenarios, you can circumvent the safety here by interrupting logstash two times: the first time it will try to safely shutdown, and a second time should cause it to abandon safety and exit.

The reported error "[2017-09-25T19:54:44,850][ERROR][logstash.outputs.elasticsearch] Encountered a retryable error. Will Retry with exponential backoff  {:code=>400, :url=>"http://es-data-int.ex.com:9200/_bulk", :body=>"{\"error\":{\"root_cause\":[{\"type\":\"parse_exception\",\"reason\":\"request body is required\"}],\"type\":\"parse_exception\",\"reason\":\"request body is required\"},\"status\":400}"}" in the discuss forum, "request body is required" is a bug, not a feature, and we should fix that bug.

jordansissel:
For reasons stated above, Logstash has a default mechanism of safe shutdown. If you wish to abandon this safety, that is a choice you can make, but we will not make this unsafe (abandon the retry) the default behavior.

hasnat:
thanks for the explaining @jordansissel 
I never knew those 400s were due to a different bug, my goal is to pass down all the 400s to dead letter queue. Would there be a way to do this via config instead of sending interrupt signals to logstash?

PR #385, opened 2016-03-03
Fix #239 - interpolated action "update" don't get params "upsert" and…

dchauviere:
This PR fixes a bug (issue #239) in case of interpolated action "update" where specific parameters 'upsert' and 'script' was not set !

PR #732, opened 2018-02-08
Pass through formatted action to event_action_params

### LINKED ISSUE #731: Using sprintf strings for `action` breaks `update` params
I ran into an issue where `_retry_on_conflict` wasn’t being included in my requests, I initially posted about it [on the forums](https://discuss.elastic.co/t/elasticsearch-output-not-respecting-retry-on-conflict/118964), however additional debugging has lead me to believe this is a bug:

```ruby
output {
	stdout { codec => json_lines }
	elasticsearch {
		hosts => "search:9200"
		index => "posts"
		template => "/indexer/templates/posts.json"
		action => "%{[@metadata][index_action]}"
		document_id => "%{id}"
		retry_on_conflict => 5
	}
}
```

I added some debug code to [common.rb](https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/master/lib/logstash/outputs/elasticsearch/common.rb#L189) around line 189:

```ruby
      @logger.warn "About to set update params", action: @action

      if @action == 'update'
        params[:_upsert] = LogStash::Json.load(event.sprintf(@upsert)) if @upsert != ""
        params[:_script] = event.sprintf(@script) if @script != ""
        params[:_retry_on_conflict] = @retry_on_conflict
        @logger.warn "Set update params"
      end
```

The output is as follows:

```
[2018-02-08T16:17:06,751][WARN ][logstash.outputs.elasticsearch] About to set update params {:action=>"%{[@metadata][index_action]}"}
```

Since `action` isn’t evaluated at this point, the if conditional fails and the check is ignored.

shnhrrsn:
This fixes #731 by swapping two lines in common.rb `event_action_tuple` to format the action before calling `event_action_params` and then passing the formatted `action` through so the `update` conditional will work.

### DIFF
--- a/lib/logstash/outputs/elasticsearch/common.rb
+++ b/lib/logstash/outputs/elasticsearch/common.rb
@@ -36,8 +36,8 @@ module LogStash; module Outputs; class ElasticSearch;
 
     # Convert the event into a 3-tuple of action, params, and event
     def event_action_tuple(event)
-      params = event_action_params(event)
       action = event.sprintf(@action)
+      params = event_action_params(event, action)
       [action, params, event]
     end
 
@@ -161,7 +161,7 @@ module LogStash; module Outputs; class ElasticSearch;
     end
 
     # get the action parameters for the given event
-    def event_action_params(event)
+    def event_action_params(event, action)
       type = get_event_type(event)
 
       params = {
@@ -186,7 +186,7 @@ module LogStash; module Outputs; class ElasticSearch;
         end
       end
 
-      if @action == 'update'
+      if action == 'update'
         params[:_upsert] = LogStash::Json.load(event.sprintf(@upsert)) if @upsert != ""
         params[:_script] = event.sprintf(@script) if @script != ""
         params[:_retry_on_conflict] = @retry_on_conflict


guyboertje:
Nice find and thanks for your contribution. The travis build failure is unrelated to changes in this PR.

IMO the tests should have caught this, but the tests don't cover this case.
You should take the opportunity to fix that.

This `action` test [tests  string interpolation of the action setting](https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/master/spec/unit/outputs/elasticsearch_spec.rb#L282).

This `retry_on_conflict` test [claims that interpolation on the action setting should take place](https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/master/spec/unit/outputs/elasticsearch_spec.rb#L352) but does not.

A third test that actually tests `retry_on_conflict` with an interpolated action would have caught this bug. Please switch out lines 338 to 360, the whole describe block, with the following. 

```
  describe "retry_on_conflict" do
    let(:num_retries) { 123 }
    let(:event) { LogStash::Event.new("myactionfield" => "update", "message" => "blah") }
    let(:options) { { 'retry_on_conflict' => num_retries } }

    context "with a regular index" do
      let(:options) { super.merge("action" => "index") }

      it "should not set the retry_on_conflict parameter when creating an event_action_tuple" do
        action, params, event_data = subject.event_action_tuple(event)
        expect(params).not_to include({:_retry_on_conflict => num_retries})
      end
    end

    context "using a plain update" do
      let(:options) { super.merge("action" => "update", "retry_on_conflict" => num_retries, "document_id" => 1) }

      it "should set the retry_on_conflict parameter when creating an event_action_tuple" do
        action, params, event_data = subject.event_action_tuple(event)
        expect(params).to include({:_retry_on_conflict => num_retries})
      end
    end

    context "with a sprintf action that resolves to update" do
      let(:options) { super.merge("action" => "%{myactionfield}", "retry_on_conflict" => num_retries, "document_id" => 1) }

      it "should set the retry_on_conflict parameter when creating an event_action_tuple" do
        action, params, event_data = subject.event_action_tuple(event)
        expect(params).to include({:_retry_on_conflict => num_retries})
      end
    end
  end
```
If you run this ⬆️ test with the old code it fails (I did and it did).

It makes me 😢  that `event_action_tuple` and `event_action_params` still exist separately.

In my local copy after the test fail, to make it pass, I renamed `event_action_params` to `event_action_tuple`, moved the action interpolation down, removed the `@`, returned the tuple and deleted the old `event_action_tuple` method. `event_action_params` is not called in other code or tests BTW.

You are encouraged to do the same but it is not essential.

guyboertje:
@shnhrrsn ⬆️

Da-Wei:
Hello @shnhrrsn, @guyboertje, does this PR address the same issue as #385 ?

dchauviere:
Hello @shnhrrsn, @guyboertje, does this PR address the same issue as #385 ?

shnhrrsn:
@dchauviere yup looks like it's the same thing, should we close this out in favor of #385?

guyboertje:
@dchauviere and @shnhrrsn
Either way, I would like to see, at least, the tests fixed as per my comment above and, at best, the collapsing for the `event_action_tuple` and `event_action_params` methods into one method having the call signature + return value of `event_action_tuple` with the method body of `event_action_params`.
@dchauviere - are you up for that?

shnhrrsn:
@guyboertje definitely agreed! I was planning to incorporate your suggestions, just haven't gotten a chance yet.  Happy to defer to @dchauviere

guyboertje:
#385 needs a rebase though.

dchauviere:
Thanks @shnhrrsn to let me close an old bug that I introduce 😄 

@guyboertje, changes are on the road ...

guyboertje:
Closing this in favour of #385.

@dchauviere @shnhrrsn 
Thanks to you both.

Issue #736, opened 2018-02-14
Zero body size bulk call results in a 400 and infinite loop in 7.x


cjhodges77:
We are running Logstash 5.6.5. We see repeated 400 errors with the message 'Encountered a retryable error. Will Retry with exponential backoff ' that accumulate until the Logstash instance fails and no longer processes logs. We are unsure why Logstash stops processing, possibly file handlers, as in IOException: Too many open files - TCP input. There are no OOM errors.

Enabling debug on the elasticsearch output we get:
{"level":"ERROR","loggerName":"logstash.outputs.elasticsearch","timeMillis":1517840567573,"thread":"[main]>worker2","logEvent":{"message":"Encountered a retryable error. Will Retry with exponential backoff ","code":400,"url":"http://127.0.0.1:9200/_bulk","body":"{"error":{"root_cause":[{"type":"parse_exception","reason":"request body is required"}],"type":"parse_exception","reason":"request body is required"},"status":400}"}}

The errors pile up over time and are never dropped. We have DLQ enabled and do see some messages in it but none we believe are related to this issue.

The only way we have found to prevent Logstash from failing is to intermittently restart the process.

https://discuss.elastic.co/t/logstash-5-6-1-inifinite-loop-encountered-a-retryable-error/101753 and https://discuss.elastic.co/t/inifinite-loop-encountered-a-retryable-error-400-error-from-elasticsearch/119290 describe the problem.

We believe we have identified the problem and will raise a PR.

andrewvc:
Apologies @cjhodges77 we're looking int othis.

jsvd:
I should note here that his issue is solved for versions 8.x and 9.x of this plugin. Like the title says, this issue is restricted to 7.x
So, as a workaround, please contemplate the option of updating the es output.

jsvd:
version 7.4.3 has been released with a fix to this issue

jsvd:
(and thanks to @cjhodges77 for the PR)

PR #737, opened 2018-02-14
Avoid infinite loop by checking for zero body size #736

cjhodges77:
Implement https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/master/lib/logstash/outputs/elasticsearch/http_client.rb#L118 from master in 7.x.

Avoid an infinite loop caused by sending a bulk request to Elasticsearch with a zero body size.
Issue #736 
https://discuss.elastic.co/t/inifinite-loop-encountered-a-retryable-error-400-error-from-elasticsearch/119290
https://discuss.elastic.co/t/logstash-5-6-1-inifinite-loop-encountered-a-retryable-error/101753

### DIFF
--- a/lib/logstash/outputs/elasticsearch/http_client.rb
+++ b/lib/logstash/outputs/elasticsearch/http_client.rb
@@ -50,11 +50,11 @@ module LogStash; module Outputs; class ElasticSearch;
     #   through a special http path, such as using mod_rewrite.
     def initialize(options={})
       @logger = options[:logger]
-      
+
       # Again, in case we use DEFAULT_OPTIONS in the future, uncomment this.
       # @options = DEFAULT_OPTIONS.merge(options)
       @options = options
-      
+
       @url_template = build_url_template
 
       @pool = build_pool(@options)
@@ -62,7 +62,7 @@ module LogStash; module Outputs; class ElasticSearch;
       # connection pool at the same time
       @bulk_path = @options[:bulk_path]
     end
-    
+
     def build_url_template
       {
         :scheme => self.scheme,
@@ -106,7 +106,7 @@ module LogStash; module Outputs; class ElasticSearch;
       if http_compression
         body_stream.set_encoding "BINARY"
         stream_writer = Zlib::GzipWriter.new(body_stream, Zlib::DEFAULT_COMPRESSION, Zlib::DEFAULT_STRATEGY)
-      else 
+      else
         stream_writer = body_stream
       end
       bulk_responses = []
@@ -115,9 +115,9 @@ module LogStash; module Outputs; class ElasticSearch;
                     action.map {|line| LogStash::Json.dump(line)}.join("\n") :
                     LogStash::Json.dump(action)
         as_json << "\n"
-	if (body_stream.size + as_json.bytesize) > TARGET_BULK_BYTES
-          bulk_responses << bulk_send(body_stream) unless body_stream.size == 0
-	end
+        if body_stream.size > 0 && (body_stream.size + as_json.bytesize) > TARGET_BULK_BYTES
+          bulk_responses << bulk_send(body_stream)
+        end
         stream_writer.write(as_json)
       end
       stream_writer.close if http_compression
@@ -135,7 +135,7 @@ module LogStash; module Outputs; class ElasticSearch;
 
     def bulk_send(body_stream)
       params = http_compression ? {:headers => {"Content-Encoding" => "gzip"}} : {}
-      # Discard the URL 
+      # Discard the URL
       response = @pool.post(@bulk_path, params, body_stream.string)
       if !body_stream.closed?
         body_stream.truncate(0)
@@ -208,7 +208,7 @@ module LogStash; module Outputs; class ElasticSearch;
                         else
                           nil
                         end
-      
+
       calculated_scheme = calculate_property(uris, :scheme, explicit_scheme, sniffing)
 
       if calculated_scheme && calculated_scheme !~ /https?/
@@ -228,7 +228,7 @@ module LogStash; module Outputs; class ElasticSearch;
       # Enter things like foo:123, bar and wind up with foo:123, bar:9200
       calculate_property(uris, :port, nil, sniffing) || 9200
     end
-    
+
     def uris
       @options[:hosts]
     end
@@ -247,7 +247,7 @@ module LogStash; module Outputs; class ElasticSearch;
 
     def build_adapter(options)
       timeout = options[:timeout] || 0
-      
+
       adapter_options = {
         :socket_timeout => timeout,
         :request_timeout => timeout,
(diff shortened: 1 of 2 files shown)


jordansissel:
This seems like a workaround for a more fundamental bug: Why is this plugin generating a request with a zero-length body?

cjhodges77:
Good point. I'll look into it.

cjhodges77:
After looking into it pretty thoroughly our conclusion is that a document > TARGET_BULK_BYTES is moving through the pipeline. When the large doc hits this point the condition becomes TRUE and a bulk_send() is called - but before the doc has been written to body_stream - so a 400 is returned because the request body is empty, and the batch of events is then re-run indefinitly.

If a doc > TARGET_BULK_BYTES enters the pipeline it always results in an infinite loop and any other documents in the batch after the large doc are never processed.

This PR would prevent bulk_send() from being called when the body size is zero and would mean that the large doc is processed (although, as the doc is larger than TARGET_BULK_BYTES it would result in a break of the threshold).

The change is already in master and we think it's correct.

lib/logstash/outputs/elasticsearch/http_client.rb
```diff
@@ -115,7 +115,9 @@ def bulk(actions)
                     action.map {|line| LogStash::Json.dump(line)}.join("\n") :
                     LogStash::Json.dump(action)
         as_json << "\n"
-        bulk_responses << bulk_send(body_stream) if (body_stream.size + as_json.bytesize) > TARGET_BULK_BYTES
+	if (body_stream.size + as_json.bytesize) > TARGET_BULK_BYTES
```
yaauie: nitpick: since the only statement in this `if`...`end` block has a tail-conditional, it's easy to lose track of what we're doing; perhaps the following will read more clearly, and as a bonus will short-circuit out of doing the will-we-overflow calculation if our `body_stream` is empty:

~~~ ruby
        if body_stream.size > 0 && (body_stream.size + as_json.bytesize) > TARGET_BULK_BYTES
          bulk_responses << bulk_send(body_stream)
        end
~~~

spec/unit/outputs/elasticsearch/http_client_spec.rb
```diff
@@ -180,6 +180,53 @@
...
+    let(:message) { "hey" }
+    let(:actions) { [
+      ["index", {:_id=>nil, :_index=>"logstash"}, {"message"=> message}],
+    ]}
+
+    context "if a message is over TARGET_BULK_BYTES" do
+      let(:target_bulk_bytes) { LogStash::Outputs::ElasticSearch::TARGET_BULK_BYTES }
+      let(:message) { "a" * (target_bulk_bytes + 1) }
+
+      it "should be handled properly" do
```
yaauie: Phrases like "should be handled properly" require that the reader have an implicit understanding of what "handled" and "properly" mean in this particular context, which doesn't effectively summarise the example in question; although I see that the existing examples aren't particularly consistent about doing so, it's also nice to use imperative language instead of "should"-type statements.

perhaps this would be better?

~~~ ruby
      it "sends the message as its own bulk payload" do
~~~

spec/unit/outputs/elasticsearch/http_client_spec.rb
```diff
@@ -180,6 +180,53 @@
     end
   end
 
+  describe "#bulk" do
+    subject { described_class.new(base_options) }
+
+    require "json"
```
yaauie: 🤔 I don't see any of the lines in the new spec using the top-level `::JSON` constant, so either we don't need this `require "json"` statement or something is broken further down.

As a side note, because of how rspec loads things, when we _do_ have specs that require external additional dependencies, it's better to load them in a `before` block so that they're only loaded if we're _executing_ the example group:

~~~ ruby
    before do
      require "json"
    end
~~~

spec/unit/outputs/elasticsearch/http_client_spec.rb
```diff
@@ -180,6 +180,53 @@
...
+    end
+
+    context "with two messages" do
+      let(:message1) { "hey" }
+      let(:message2) { "you" }
+      let(:actions) { [
+        ["index", {:_id=>nil, :_index=>"logstash"}, {"message"=> message1}],
+        ["index", {:_id=>nil, :_index=>"logstash"}, {"message"=> message2}],
+      ]}
+      it "executes one bulk_send operation" do
```
yaauie: nitpick: adding a line of whitespace between the setup `let` statements and the specs themselves would be consistent with our other example groups, and add a bit of visual separation making it easier to skim.

yaauie:
CI is failing because of https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/740

mkubenka:
Any update on this?

jsvd:
@mkubenka any progress is (and will be) visible through this issue.

jsvd:
@mkubenka I agree that this is generating frustration..The PR has been reviewed an changes have been asked for.

yaauie:
It looks like we've got a couple failing tests in CI -- can you either resolve or assert that they're not your fault? :) https://travis-ci.org/logstash-plugins/logstash-output-elasticsearch/jobs/364678364#L1091

cjhodges77:
Explicitly setting compression to false rather than {} fixes the related broken tests as {} is truthy and causes bulk_send to always set {"Content-Encoding" => "gzip"}. 

The remaining broken test "Host/URL Parsing an ipv6 host should properly transform a host:port string to a URL" isn't me.

cjhodges77:
Hi, is there any update on this please?

jordansissel:
I'll defer to @yaauie on the review. @cjhodges77, can you update the PR subject to reflect your new approach to fixing this?

cjhodges77:
I don't think the approach has changed. My last commit fixed an additional issue highlighted by the change to fix the initial bug, in that content-encoding was always set to gzip regardless of whether compression was true or false.

I'm happy to make changes as asked if necessary to progress. Thanks

cjhodges77:
As stated above, this change already exists on the other branches. 7.x is missing this fix.

jordansissel:
Ahh, I misread the PR (there's a ton of unrelated whitespace changes, so the noise level is high).

This PR's `if body_stream.size > 0` check is still working around a bug instead of fixing the bug? The bug should be fixed, not worked around.

At a minimum, if the workaround-for-the-bug is to be merged, please put a descriptive comment explaining why the code checks for nonzero body size.

jsvd:
I added this to my ToDo list, I'll try to check it tomorrow.

I think the naming of some variables in this code is unfortunate, this
body_stream is an accumulator, so in some scenarios it can be empty as
we're about to put an event in it, or we want to put something there but
that something needs the accumulator for itself and we must flush its
contents.

cjhodges77:
Tests passed.

PR #754, opened 2018-03-28
Add an example for "script" option in documentation

fbaligand:
Add this example : 
`script => "ctx._source.message = event.get('message')"`


fbaligand:
Thanks for the merge @jsvd  !

PR #758, opened 2018-04-23
Improve plugin behavior when Elasticsearch is down on startup

### LINKED ISSUE #715: Sending events when ES is down causes LS to crash
If you send LS an event, by invoking it thusly:

` bin/logstash -e 'input { stdin { type => foo } } output { elasticsearch {} }'`

and do **not** have ES up and running, LS will crash once an event is sent due to an error like:

```
NoMethodError: undefined method `<' for nil:NilClass
          get_event_type at /Users/andrewcholakian/Downloads/logstash-6.1.0/vendor/bundle/jruby/2.3.0/gems/logstash-output-elasticsearch-9.0.2-java/lib/logstash/outputs/elasticsearch/common.rb:213
```

We should make the connection pool block on returning the version until we've seen at least one host.

### LINKED ISSUE #750: Unexpected error when plugin tries to resurrect connection to Elasticsearch
I had a case where Logstash loose connection to Elasticsearch, then tries to resurrect it, and raises an unexpected error, which brings to Logstash brutal shutdown.
So it is a critical bug.

Here's the logs : 

```
[2018-03-15T09:46:37,812][INFO ][logstash.outputs.elasticsearch] Running health check to see if an Elasticsearch connection is working {:healthcheck_url=>http://logstash:xxxxxx@localhost:10400/, :path=>"/"}
[2018-03-15T09:46:37,815][WARN ][logstash.outputs.elasticsearch] Attempted to resurrect connection to dead ES instance, but got an error. {:url=>"http://logstash:xxxxxx@localhost:10400/", :error_type=>LogStash::Outputs::ElasticSearch::HttpClient::Pool::BadResponseCodeError, :error=>"Got response code '403' contacting Elasticsearch at URL 'http://localhost:10400/'"}
[2018-03-15T09:46:41,692][FATAL][logstash.runner          ] An unexpected error occurred! {:error=>#<NoMethodError: undefined method `<' for nil:NilClass>, :backtrace=>["/usr/product/logstash/logstash-6.1.3/vendor/bundle/jruby/2.3.0/gems/logstash-output-elasticsearch-9.0.2-java/lib/logstash/outputs/elasticsearch/common.rb:213:in `get_event_type'", "/usr/product/logstash/logstash-6.1.3/vendor/bundle/jruby/2.3.0/gems/logstash-output-elasticsearch-9.0.2-java/lib/logstash/outputs/elasticsearch/common.rb:165:in `event_action_params'", "/usr/product/logstash/logstash-6.1.3/vendor/bundle/jruby/2.3.0/gems/logstash-output-elasticsearch-9.0.2-java/lib/logstash/outputs/elasticsearch/common.rb:39:in `event_action_tuple'", "/usr/product/logstash/logstash-6.1.3/vendor/bundle/jruby/2.3.0/gems/logstash-output-elasticsearch-9.0.2-java/lib/logstash/outputs/elasticsearch/common.rb:34:in `block in multi_receive'", "org/jruby/RubyArray.java:2486:in `map'", "/usr/product/logstash/logstash-6.1.3/vendor/bundle/jruby/2.3.0/gems/logstash-output-elasticsearch-9.0.2-java/lib/logstash/outputs/elasticsearch/common.rb:34:in `multi_receive'", "/usr/product/logstash/logstash-6.1.3/logstash-core/lib/logstash/output_delegator_strategies/shared.rb:13:in `multi_receive'", "/usr/product/logstash/logstash-6.1.3/logstash-core/lib/logstash/output_delegator.rb:50:in `multi_receive'", "/usr/product/logstash/logstash-6.1.3/logstash-core/lib/logstash/pipeline.rb:487:in `block in output_batch'", "org/jruby/RubyHash.java:1343:in `each'", "/usr/product/logstash/logstash-6.1.3/logstash-core/lib/logstash/pipeline.rb:486:in `output_batch'", "/usr/product/logstash/logstash-6.1.3/logstash-core/lib/logstash/pipeline.rb:438:in `worker_loop'", "/usr/product/logstash/logstash-6.1.3/logstash-core/lib/logstash/pipeline.rb:393:in `block in start_workers'"]}
```

robbavey:
Prior to this commit, if Elasticsearch was down during Logstash startup,
  one of two outcomes would occur:

  If Elasticsearch was still down when the first events arrived, Logstash
    would immediately exit with an error
  If Elasticsearch was up when the first events arrived, Logstash would
    accept the messages, but would the template for the index would not
    be installed.

This commit will install the template only after a successful call to
  Elasticsearch has been made. New messages will block until connectivity
  to Elasticsearch is re-established and the template has been installed

fixes #715 
fixes #750

### DIFF
--- a/lib/logstash/outputs/elasticsearch/common.rb
+++ b/lib/logstash/outputs/elasticsearch/common.rb
@@ -23,9 +23,8 @@ module LogStash; module Outputs; class ElasticSearch;
 
       setup_hosts # properly sets @hosts
       build_client
-      after_successful_connection { install_template }
+      install_template_after_successful_connection
       check_action_validity
-      #after_successful_connection { install_template }
       @bulk_request_metrics = metric.namespace(:bulk_requests)
       @document_level_metrics = metric.namespace(:documents)
 
@@ -40,25 +39,24 @@ module LogStash; module Outputs; class ElasticSearch;
       retrying_submit(events.map {|e| event_action_tuple(e)})
     end
 
-    def after_successful_connection
-      @template_installer = Thread.new do
+    def install_template_after_successful_connection
+      @template_installer ||= Thread.new do
           sleep_interval = @retry_initial_interval
           until successful_connection? || @stopping.true?
             @logger.debug("Waiting for connectivity to Elasticsearch cluster")
             Stud.stoppable_sleep(sleep_interval) { @stopping.true? }
             sleep_interval = next_sleep_interval(sleep_interval)
           end
-          yield if successful_connection?
+            install_template if successful_connection?
         end
       end
 
     def stop_template_installer
-      @template_installer.join if @template_installer
+      @template_installer.join unless @template_installer.nil?
     end
 
     def successful_connection?
-      # !!maximum_seen_major_version
-      maximum_seen_major_version
+      !!maximum_seen_major_version
     end
 
     # Convert the event into a 3-tuple of action, params, and event
--- a/lib/logstash/outputs/elasticsearch/common.rb
+++ b/lib/logstash/outputs/elasticsearch/common.rb
@@ -23,11 +23,10 @@ module LogStash; module Outputs; class ElasticSearch;
 
       setup_hosts # properly sets @hosts
       build_client
-      install_template_after_successful_connection
       check_action_validity
       @bulk_request_metrics = metric.namespace(:bulk_requests)
       @document_level_metrics = metric.namespace(:documents)
-
+      install_template_after_successful_connection
       @logger.info("New Elasticsearch output", :class => self.class.name, :hosts => @hosts.map(&:sanitized).map(&:to_s))
     end
 
@@ -41,15 +40,15 @@ module LogStash; module Outputs; class ElasticSearch;
 
     def install_template_after_successful_connection
       @template_installer ||= Thread.new do
-          sleep_interval = @retry_initial_interval
-          until successful_connection? || @stopping.true?
-            @logger.debug("Waiting for connectivity to Elasticsearch cluster. Retrying in #{sleep_interval}s")
-            Stud.stoppable_sleep(sleep_interval) { @stopping.true? }
-            sleep_interval = next_sleep_interval(sleep_interval)
-          end
-            install_template if successful_connection?
+        sleep_interval = @retry_initial_interval
+        until successful_connection? || @stopping.true?
+          @logger.debug("Waiting for connectivity to Elasticsearch cluster. Retrying in #{sleep_interval}s")
+          Stud.stoppable_sleep(sleep_interval) { @stopping.true? }
+          sleep_interval = next_sleep_interval(sleep_interval)
         end
+        install_template if successful_connection?
       end
+    end
 
     def stop_template_installer
       @template_installer.join unless @template_installer.nil?
--- a/lib/logstash/outputs/elasticsearch/common.rb
+++ b/lib/logstash/outputs/elasticsearch/common.rb
@@ -43,7 +43,7 @@ module LogStash; module Outputs; class ElasticSearch;
       @template_installer ||= Thread.new do
           sleep_interval = @retry_initial_interval
           until successful_connection? || @stopping.true?
-            @logger.debug("Waiting for connectivity to Elasticsearch cluster")
+            @logger.debug("Waiting for connectivity to Elasticsearch cluster. Retrying in #{sleep_interval}s")
             Stud.stoppable_sleep(sleep_interval) { @stopping.true? }
             sleep_interval = next_sleep_interval(sleep_interval)
           end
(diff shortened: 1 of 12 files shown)


jsvd:
@robbavey it seems some tests are still failing

lib/logstash/outputs/elasticsearch/common.rb
```diff
@@ -16,15 +16,16 @@ module Common
...
       # To support BWC, we check if DLQ exists in core (< 5.4). If it doesn't, we use nil to resort to previous behavior.
       @dlq_writer = dlq_enabled? ? execution_context.dlq_writer : nil
 
       setup_hosts # properly sets @hosts
       build_client
-
-      install_template
+      after_successful_connection { install_template }
       check_action_validity
+      #after_successful_connection { install_template }
```
jsvd: this leftover commented code can be removed

lib/logstash/outputs/elasticsearch/common.rb
```diff
@@ -33,9 +34,33 @@ def register
...
+          yield if successful_connection?
+        end
+      end
+
+    def stop_template_installer
+      @template_installer.join if @template_installer
+    end
+
+    def successful_connection?
+      # !!maximum_seen_major_version
```
jsvd: this leftover commented code can be removed
also, any particular reasoning to (not) use !! here?

lib/logstash/outputs/elasticsearch/common.rb
```diff
@@ -33,9 +34,33 @@ def register
...
     # Receive an array of events and immediately attempt to index them (no buffering)
     def multi_receive(events)
+      until @template_installed.true?
+        sleep 1
+      end
       retrying_submit(events.map {|e| event_action_tuple(e)})
     end
 
+    def after_successful_connection
+      @template_installer = Thread.new do
```
jsvd: I'm a bit concerned here that a pipeline could mistakenly call register twice, causing after_successful_connection to also be called twice, therefore leaking threads as `@template_installer` will be rewritten.

lib/logstash/outputs/elasticsearch/common.rb
```diff
@@ -16,15 +16,16 @@ module Common
...
+      @template_installed = Concurrent::AtomicBoolean.new(false)
       @stopping = Concurrent::AtomicBoolean.new(false)
       # To support BWC, we check if DLQ exists in core (< 5.4). If it doesn't, we use nil to resort to previous behavior.
       @dlq_writer = dlq_enabled? ? execution_context.dlq_writer : nil
 
       setup_hosts # properly sets @hosts
       build_client
-
-      install_template
+      after_successful_connection { install_template }
```
jsvd: since we use `after_successful_connection` only once, could we avoid threading here and just have a `wait_for_successful_connection` method that performs the waiting for a connection stuff in the thread calling register?:

```ruby
def wait_for_successful_connection
  until successful_connection? || @stopping.true?
    @logger.debug("Waiting for connectivity to Elasticsearch cluster")
    Stud.stoppable_sleep(sleep_interval) { @stopping.true? }
    sleep_interval = next_sleep_interval(sleep_interval)
  end
end
```

```ruby
def register
  #...
  setup_hosts
  build_client
  wait_for_succesful_connection
  install_template
  # ..
end
```

lib/logstash/outputs/elasticsearch/common.rb
robbavey: I thought I'd already pushed the commit that did this...

lib/logstash/outputs/elasticsearch/common.rb
robbavey: Will fix

lib/logstash/outputs/elasticsearch/common.rb
robbavey: I started here, but the problem with this approach is that it stops a clean shutdown if Elasticsearch never becomes available.

lib/logstash/outputs/elasticsearch/common.rb
robbavey: !! makes more sense in the context of a `?` method, so I'll use that

lib/logstash/outputs/elasticsearch/common.rb
jsvd: why would it prevent a clean shutdown? the `wait_for_successful_connection` method would loop with the `@stopping.true?` guard + stoppable_sleep

robbavey: The `close` method of the plugin (which calls `@stopping.make_true`) never gets called if the `register` method is blocked.

jsvd:
@robbavey failures seem to be real, specially [this one](https://travis-ci.org/logstash-plugins/logstash-output-elasticsearch/jobs/376105533#L1073)

robbavey:
Thanks for merging @jsvd - missed the LGTM...

jsvd:
no worries, with the LGTM I shouldn't have to wait anyway, either of us can merge

PR #761, opened 2018-04-27
cleanup travis ci file for 7.x branch

jsvd:
no point in testing against logstash 6.x as 6.0.0 shipped with es output 9.0.0

remove references to 6.0.0-beta2, update version of logstash 5.6 from 5.6.0 to 5.6.9


robbavey:
@jsvd Builds failing on old Elasticsearch versions in Travis, although they look like they may be transient. Will LGTM once builds go green

jsvd:
💚

PR #763, opened 2018-04-27
Remove unused gzip dep

andrewvc:
This dep wasn't being used, and had no explicit license either. (and also monkey patched string for bad reasons).


andrewvc:
Will do once tests pass 🙏

PR #771, opened 2018-05-17
Raise logging level for 'invalid_index_name_exception' ES exceptions to ERROR

### LINKED ISSUE #759: WARN message in logstash.log
Operations use beat to send data  to logstash and then to Elasticsearch. They add the following filters 
in logstash pipeline:

```
filter {
   mutate {
      add_field => { "[beat][beat]" => "%{[@metadata][beat]}"}
      add_field => { "[beat][version]" => "%{[@metadata][version]}"}
 }
}
```

And here is the logstash output plugin :

```
output {
  stdout { codec => rubydebug }
  elasticsearch {
    hosts => ["localhost:9200"]
    id => "beats_to_elk_test"
    codec => "json"
    manage_template => false
    index => "%{[beat][beat]}-%{[beat][version]}-%{+YYYY.MM.dd}"
  }
}
```

Index didn't got created because the beat.version is an array now "beat.version => 5.6.4, %{[@metadata][version]}", There is WARN message in logstash log:

[2018-04-16T17:03:25,774][WARN ][logstash.outputs.elasticsearch] Could not index event to Elasticsearch. {:status=>400, :action=>["index", {:_id=>nil, :_index=>"filebeat-5.6.4,%{[@metadata][version]}-2018.04.17", :_type=>"log", :_routing=>nil}, 2018-04-17T00:03:20.605Z local 111.111.111.111 - - [25/Sep/2014:12:00:00 +0000] "GET /scripts/netcat-webserver HTTP/1.1" 200 182 "-" "Mozilla/5.0 (compatible; EasouSpider; +http://www.easou.com/search/spider.html)"], :response=>{"index"=>{"_index"=>"filebeat-5.6.4,%{[@metadata][version]}-2018.04.17", "_type"=>"log", "_id"=>nil, "status"=>400, "error"=>{"type"=>"invalid_index_name_exception", "reason"=>"Invalid index name [filebeat-5.6.4,%{[@metadata][version]}-2018.04.17], must not contain the following characters [ , \", *, \\, <, |, ,, >, /, ?]", "index_uuid"=>"_na_", "index"=>"filebeat-5.6.4,%{[@metadata][version]}-2018.04.17"}}}}

Since it is a WARN message, customer probably miss that one and don't know why index not created in Elasticsearch.

Could we make this as ERROR message instead?

- Version: logstash 5.6.4

webmat:
As discussed in #759, an invalid index exception should log at the ERROR level.
Presumably, most of these will happen just after an update to a user's Logstash
configuration.

Still missing in this PR

- [x] Add tests
- [x] Version bump

Questions

- Does `dlq_writer.write` already support `(event, reason, params_hash)`?
- When we get an ES exception, are `response['index']` and `response['index']['error']`
  guaranteed to be set, or should I guard against anything in there being nil?
  - Got nil exceptions in the integration tests. Now guarding.

I would posit that making the list of exceptions that are elevated to ERROR is
out of scope for this. We could release this as is, and if there's interest we
make it configurable later on.

Generate an `invalid_index_name_exception` with

```
bin/logstash -e "input { generator { count => 1 } }
  output { elasticsearch { user => 'elastic' password => 'elastic'
    index => 'bad,index'
  }}"
```

Generate another ES exception (mapping) with

```
bin/logstash -e "input { generator { count => 2 } }
  filter { ruby { code => \" event.set('mapping_fail', (event.get('sequence') == 0 ? {'foo' => 'bar'} : 'foobar' ) ) \"} }
  output { elasticsearch { user => 'elastic' password => 'elastic' }}"
```

closes #759

### DIFF
--- a/lib/logstash/outputs/elasticsearch/common.rb
+++ b/lib/logstash/outputs/elasticsearch/common.rb
@@ -213,14 +213,7 @@ module LogStash; module Outputs; class ElasticSearch;
           @logger.warn "Failed action.", status: status, action: action, response: response if !failure_type_logging_whitelist.include?(failure["type"])
           next
         elsif DOC_DLQ_CODES.include?(status)
-          action_event = action[2]
-          # To support bwc, we check if DLQ exists. otherwise we log and drop event (previous behavior)
-          if @dlq_writer
-            # TODO: Change this to send a map with { :status => status, :action => action } in the future
-            @dlq_writer.write(action_event, "Could not index event to Elasticsearch. status: #{status}, action: #{action}, response: #{response}")
-          else
-            @logger.warn "Could not index event to Elasticsearch.", status: status, action: action, response: response
-          end
+          handle_dlq_status("Could not index event to Elasticsearch.", action, status, response)
           @document_level_metrics.increment(:non_retryable_failures)
           next
         else
@@ -232,6 +225,22 @@ module LogStash; module Outputs; class ElasticSearch;
       end
 
       actions_to_retry
+    end
+
+    def handle_dlq_status(message, action, status, response)
+      # To support bwc, we check if DLQ exists. otherwise we log and drop event (previous behavior)
+      if @dlq_writer
+        # TODO: Change this to send a map with { :status => status, :action => action } in the future
+        @dlq_writer.write(action[2], "#{message} status: #{status}, action: #{action}, response: #{response}")
+      else
+        error_type = response.fetch('index', {}).fetch('error', {})['type']
+        if 'invalid_index_name_exception' == error_type
+          level = :error
+        else
+          level = :warn
+        end
+        @logger.send level, message, status: status, action: action, response: response
+      end
     end
 
     # Determine the correct value for the 'type' field for the given event
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 9.1.4
+  - Log an error -- not a warning -- when ES raises an invalid\_index\_name\_exception.
+
 ## 9.1.3
   - Improve plugin behavior when Elasticsearch is down on startup #758
 
--- a/logstash-output-elasticsearch.gemspec
+++ b/logstash-output-elasticsearch.gemspec
@@ -1,6 +1,6 @@
 Gem::Specification.new do |s|
   s.name            = 'logstash-output-elasticsearch'
-  s.version         = '9.1.3'
+  s.version         = '9.1.4'
   s.licenses        = ['apache-2.0']
   s.summary         = "Stores logs in Elasticsearch"
   s.description     = "This gem is a Logstash plugin required to be installed on top of the Logstash core pipeline using $LS_HOME/bin/logstash-plugin install gemname. This gem is not a stand-alone program"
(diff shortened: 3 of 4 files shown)


webmat:
This new `if` is deep in the `submit` method. This will not be easy to get to. Perhaps I should extract the error logging to a small helper method?

webmat:
Test failures on 3a848c8 with older versions of ES answered my question about guarding for `response['index']['error']`. Got nil exceptions there. I'll guard against this and move this code to a small method that'll be easier to test.

lib/logstash/outputs/elasticsearch/common.rb
```diff
@@ -234,6 +227,22 @@ def submit(actions)
       actions_to_retry
     end
 
+    def report_doc_level_error(message, action, status, response)
+      # To support bwc, we check if DLQ exists. otherwise we log and drop event (previous behavior)
+      if @dlq_writer
+        # TODO: Change this to send a map with { :status => status, :action => action } in the future
+        @dlq_writer.write(action[2], "#{message} status: #{status}, action: #{action}, response: #{response}")
```
jsvd: I'm not a huge fan of this "report_doc_level_error" to be writing data to the DLQ. Maybe we can have this method only take care of the `else` case in this if, where the DLQ isn't  available.
Another alternative is to call this method something like `handle_doc_level_error`. the word "handle" works for everything, yay

webmat: Yes, `handle_` is a better name

webmat:
Ah, I had missed the [PluginDeadLetterQueueWriter class](https://github.com/elastic/logstash/blob/master/logstash-core/lib/logstash/util/dead_letter_queue_manager.rb) sitting in the middle. Now it makes sense 😂

lib/logstash/outputs/elasticsearch/common.rb
webmat: Actually this method is specifically for statuses that are candidates for the DLQ. I'll go even further and rename it to `handle_dlq_status`. It logs if DLQ is not enabled, and it sends to DLQ if it's enabled.

webmat:
We should be good to go for PR, as of commit 5c074c1. Version bump is done, `invalid_index_name_exception` logs an error instead of a warning, and I've started adding significant tests around handling of events getting statuses that are candidates for the DLQ (whether or not it's enabled/available).

I've noticed a few things while working on this, and I've created follow-up issue #772 to handle them.

lib/logstash/outputs/elasticsearch/common.rb
```diff
@@ -234,6 +227,22 @@ def submit(actions)
       actions_to_retry
     end
 
+    def handle_dlq_status(message, action, status, response)
+      # To support bwc, we check if DLQ exists. otherwise we log and drop event (previous behavior)
+      if @dlq_writer
+        # TODO: Change this to send a map with { :status => status, :action => action } in the future
+        @dlq_writer.write(action[2], "#{message} status: #{status}, action: #{action}, response: #{response}")
+      else
+        error_type = response['index'] && response['index']['error'] && response['index']['error']['type']
```
jsvd: We can do `response.fetch('index', {}).fetch('error', {}).get('type')` instead.

webmat:
Ok @jsvd All tests are passing now. I'd like to merge tomorrow if there's nothing else.

Thanks for the terser syntax for digging into nested hashes :-)

Issue #772, opened 2018-05-18
Tweak the handling of the retry-able & DLQ errors


webmat:
In working on -- and looking at -- simple enough issues (#771 and #697), I've noticed a few problems on how the plugin handles the DLQ.

Here are a few things I think we could tweak:

* Add a new attribute to configure additional statuses to DLQ (currently only on 400 and 404)
  * This would solve situations like #697, #134 (See also customer issues linked to #697)
* Deprecate & rename confusingly named `failure_type_logging_whitelist`. This is a list of error statuses we don't want to log. Perhaps an attribute named `silence_`...
* Encapsulate the DLQing in a small helper class
  * Avoid the nil setting and checking business
  * Messy logic in detecting status of DLQ would be encapsulated there, making https://github.com/elastic/logstash/issues/8064 less of an issue
  * Sending to DLQ should also log (unless it's a status configured to be silenced), it shouldn't be DLQ vs log.
  * This encapsulation would help handle potential big changes to how the DLQ work, e.g. https://github.com/elastic/logstash/issues/9502

chandlermelton:
I have noticed that Logstash pipeline have become less resilient to this problem since my upgrade from 5.1.2 to 5.6.9. I am looking into implementing DLQ for this. (My issue is that a server will come out of a day or weeks old maintenance and filebeat will attempt to ship any unsent logs near the time of the original crash into an index which has since been closed.)


Perhaps an option could also be adding a tag or flag on an index which could instruct Logstash's response. ( index_when_closed_repsonse: "DLQ/ drop / new_target_index" )

webmat:
@chandlermelton Another thing you can do, when your machines can send old logs, is to have Logstash automatically drop any messages older than your index closing/pruning policy.

chandlermelton:
I am definitely interested in this. I condsidered it, but wasn't sure how I could go about this without hardcoding index names in a drop filter. Is there a way to say:
if dest_index = today-7days, then {drop} or something similar?

An example would be logtype-2018.06.04-7days meaning 2018.05.28 and older would be dropped.

webmat:
One way to do it is with the `age` filter. It's specifically the example given in the [description of the plugin](https://www.elastic.co/guide/en/logstash-versioned-plugins/current/v1.0.2-plugins-filters-age.html). Note that you may have to add more checks if some indices have longer retentions than others. 

[...]

Issue #718, opened 2017-12-28
DLQ when sprintf references a missing field


jordansissel:
Proposal: When sprintf on the index name would fail because one or more fields do not exist, Logstash should DLQ the event instead of sending it to Elasticsearch.

---

Problem: 

Logstash's event sprintf has a behavior that when a field does not exist, the formatting string is left unaltered.

For example, if a user has this:

```
index => "%{[my][field]}-%{+YYYY}"
````

and supposing that `[my][field]` does not exist in the event, an index will be created in Elasticsearch literally as `%{[my][field]}-2017`. 

This is confusing and also tricky to recover from. It's tricky partly because of the `{}` and `[]` format and field-reference characters which `curl` uses (by default) with special meaning: globbing. So for a user to delete this index, they will experience an error:

```
% curl 'https://es.foo.com:9200/%{[my][field]]}-2017'
curl: (3) [globbing] nested brace in column 27
```

Or worse, if no nested fields are being used:

```
% curl -v 'http://localhost:9200/%{myfield}-2017.12.28'
> GET /%myfield-2017.12.28 HTTP/1.1
```

As above, curl interprets the `{}` to have special meaning, and this will certainly confuse users.

This is the beginning of a [yak shave](http://whatis.techtarget.com/definition/yak-shaving).

PR #777, opened 2018-05-25
Survive attempts at indexing events that don't have  a @timestamp

### LINKED ISSUE #739: Unexpected error when @timestamp field missing
- Version: Logstash 5.x and Logstash 6.x

- Operating System: Any

- Config File (if you have sensitive info, please remove it):

```
input {
  generator {}
}
filter {
  mutate {
    rename => {
      "@timestamp" => "timestamp"
    }
  }
}

output {
  elasticsearch {
    user => elastic
    password => xxxxx
  }
}
```

- Sample Data:

Any data

- Steps to Reproduce:

Run logstash with that configuration.

Output:

5.x

```
[2018-02-22T13:49:42,425][FATAL][logstash.runner          ] An unexpected error occurred! {:error=>#<LogStash::Error: timestamp field is missing>, :backtrace=>["org/logstash/ext/JrubyEventExtLibrary.java:205:in `sprintf'", "/Users/Gabriel/Documents/ElasticSearch/logstash-5.6.2/vendor/bundle/jruby/1.9/gems/logstash-output-elasticsearch-7.4.0-java/lib/logstash/outputs/elasticsearch/common.rb:169:in `event_action_params'", "/Users/Gabriel/Documents/ElasticSearch/logstash-5.6.2/vendor/bundle/jruby/1.9/gems/logstash-output-elasticsearch-7.4.0-java/lib/logstash/outputs/elasticsearch/common.rb:44:in `event_action_tuple'", "/Users/Gabriel/Documents/ElasticSearch/logstash-5.6.2/vendor/bundle/jruby/1.9/gems/logstash-output-elasticsearch-7.4.0-java/lib/logstash/outputs/elasticsearch/common.rb:38:in `multi_receive'", "org/jruby/RubyArray.java:2414:in `map'", "/Users/Gabriel/Documents/ElasticSearch/logstash-5.6.2/vendor/bundle/jruby/1.9/gems/logstash-output-elasticsearch-7.4.0-java/lib/logstash/outputs/elasticsearch/common.rb:38:in `multi_receive'", "/Users/Gabriel/Documents/ElasticSearch/logstash-5.6.2/logstash-core/lib/logstash/output_delegator_strategies/shared.rb:13:in `multi_receive'", "/Users/Gabriel/Documents/ElasticSearch/logstash-5.6.2/logstash-core/lib/logstash/output_delegator.rb:49:in `multi_receive'", "/Users/Gabriel/Documents/ElasticSearch/logstash-5.6.2/logstash-core/lib/logstash/pipeline.rb:436:in `output_batch'", "org/jruby/RubyHash.java:1342:in `each'", "/Users/Gabriel/Documents/ElasticSearch/logstash-5.6.2/logstash-core/lib/logstash/pipeline.rb:435:in `output_batch'", "/Users/Gabriel/Documents/ElasticSearch/logstash-5.6.2/logstash-core/lib/logstash/pipeline.rb:381:in `worker_loop'", "/Users/Gabriel/Documents/ElasticSearch/logstash-5.6.2/logstash-core/lib/logstash/pipeline.rb:342:in `start_workers'"]}
```

6.x

```
[2018-02-22T13:50:52,960][FATAL][logstash.runner          ] An unexpected error occurred! {:error=>#<LogStash::Error: timestamp field is missing>, :backtrace=>["org/logstash/ext/JrubyEventExtLibrary.java:168:in `sprintf'", "/Users/Gabriel/Documents/ElasticSearch/logstash-6.1.0/vendor/bundle/jruby/2.3.0/gems/logstash-output-elasticsearch-9.0.2-java/lib/logstash/outputs/elasticsearch/common.rb:169:in `event_action_params'", "/Users/Gabriel/Documents/ElasticSearch/logstash-6.1.0/vendor/bundle/jruby/2.3.0/gems/logstash-output-elasticsearch-9.0.2-java/lib/logstash/outputs/elasticsearch/common.rb:39:in `event_action_tuple'", "/Users/Gabriel/Documents/ElasticSearch/logstash-6.1.0/vendor/bundle/jruby/2.3.0/gems/logstash-output-elasticsearch-9.0.2-java/lib/logstash/outputs/elasticsearch/common.rb:34:in `block in multi_receive'", "org/jruby/RubyArray.java:2486:in `map'", "/Users/Gabriel/Documents/ElasticSearch/logstash-6.1.0/vendor/bundle/jruby/2.3.0/gems/logstash-output-elasticsearch-9.0.2-java/lib/logstash/outputs/elasticsearch/common.rb:34:in `multi_receive'", "/Users/Gabriel/Documents/ElasticSearch/logstash-6.1.0/logstash-core/lib/logstash/output_delegator_strategies/shared.rb:13:in `multi_receive'", "/Users/Gabriel/Documents/ElasticSearch/logstash-6.1.0/logstash-core/lib/logstash/output_delegator.rb:50:in `multi_receive'", "/Users/Gabriel/Documents/ElasticSearch/logstash-6.1.0/logstash-core/lib/logstash/pipeline.rb:487:in `block in output_batch'", "org/jruby/RubyHash.java:1343:in `each'", "/Users/Gabriel/Documents/ElasticSearch/logstash-6.1.0/logstash-core/lib/logstash/pipeline.rb:486:in `output_batch'", "/Users/Gabriel/Documents/ElasticSearch/logstash-6.1.0/logstash-core/lib/logstash/pipeline.rb:438:in `worker_loop'", "/Users/Gabriel/Documents/ElasticSearch/logstash-6.1.0/logstash-core/lib/logstash/pipeline.rb:393:in `block in start_workers'"]}
LogStash::Error: timestamp field is missing
                 sprintf at org/logstash/ext/JrubyEventExtLibrary.java:168
     event_action_params at /Users/Gabriel/Documents/ElasticSearch/logstash-6.1.0/vendor/bundle/jruby/2.3.0/gems/logstash-output-elasticsearch-9.0.2-java/lib/logstash/outputs/elasticsearch/common.rb:169
      event_action_tuple at /Users/Gabriel/Documents/ElasticSearch/logstash-6.1.0/vendor/bundle/jruby/2.3.0/gems/logstash-output-elasticsearch-9.0.2-java/lib/logstash/outputs/elasticsearch/common.rb:39
  block in multi_receive at /Users/Gabriel/Documents/ElasticSearch/logstash-6.1.0/vendor/bundle/jruby/2.3.0/gems/logstash-output-elasticsearch-9.0.2-java/lib/logstash/outputs/elasticsearch/common.rb:34
                     map at org/jruby/RubyArray.java:2486
           multi_receive at /Users/Gabriel/Documents/ElasticSearch/logstash-6.1.0/vendor/bundle/jruby/2.3.0/gems/logstash-output-elasticsearch-9.0.2-java/lib/logstash/outputs/elasticsearch/common.rb:34
           multi_receive at /Users/Gabriel/Documents/ElasticSearch/logstash-6.1.0/logstash-core/lib/logstash/output_delegator_strategies/shared.rb:13
           multi_receive at /Users/Gabriel/Documents/ElasticSearch/logstash-6.1.0/logstash-core/lib/logstash/output_delegator.rb:50
   block in output_batch at /Users/Gabriel/Documents/ElasticSearch/logstash-6.1.0/logstash-core/lib/logstash/pipeline.rb:487
                    each at org/jruby/RubyHash.java:1343
            output_batch at /Users/Gabriel/Documents/ElasticSearch/logstash-6.1.0/logstash-core/lib/logstash/pipeline.rb:486
             worker_loop at /Users/Gabriel/Documents/ElasticSearch/logstash-6.1.0/logstash-core/lib/logstash/pipeline.rb:438
  block in start_workers at /Users/Gabriel/Documents/ElasticSearch/logstash-6.1.0/logstash-core/lib/logstash/pipeline.rb:393
```

----

Should this be a better error rather than FATAL? Given that this could happen just with a specific document (if you have conditionals), best is to fail that event rather than a `FATAL` error?

webmat:
- Events being indexed to an index that doesn't require time-based interpolation
  are unaffected and are still indexed, whether or not they have a `@timestamp`.
- When events are missing `@timestamp` are destined to a timestamped index name,
  we don't send them for indexing, and we log an error.
- `#event_action_tuple` now returns nil when an event is discarded this way.
- Created a helper class to be a home for trivial functions that don't
  require the full plugin to be tested.
- I am perhaps over-testing this.

This prevents a full Logstash crash, and closes #739

**Update**

After multiple discussions on the subject elsewhere, let’s remind ourselves that this PR is about fixing a crash today for all users of 5.x and 6.x.

Whether or not we eventually prevent the removal of `@timestamp` in a future breaking change (in LS 7.0 or later) is orthogonal to this.

Let’s try to discuss only the merits of this PR in whether it solves the crash in a good way or if needs improvements.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## 9.2.0
+  - Fix a crash when the plugin tries to index an event missing a timestamp
+    to an index that needs time-based interpolation
+    (like the default `logstash-%{+YYYY.MM.dd}`).
+    When this happens, an error is logged instead.
+    ([PR #777](https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/777))
+
 ## 9.1.4
   - Log an error -- not a warning -- when ES raises an invalid\_index\_name\_exception.
 
--- a/lib/logstash/outputs/elasticsearch/common.rb
+++ b/lib/logstash/outputs/elasticsearch/common.rb
@@ -1,4 +1,5 @@
 require "logstash/outputs/elasticsearch/template_manager"
+require "logstash/outputs/elasticsearch/helpers"
 
 module LogStash; module Outputs; class ElasticSearch;
   module Common
@@ -35,7 +36,7 @@ module LogStash; module Outputs; class ElasticSearch;
       until @template_installed.true?
         sleep 1
       end
-      retrying_submit(events.map {|e| event_action_tuple(e)})
+      retrying_submit(events.map {|e| event_action_tuple(e)}.compact)
     end
 
     def install_template_after_successful_connection
@@ -58,11 +59,16 @@ module LogStash; module Outputs; class ElasticSearch;
       !!maximum_seen_major_version
     end
 
-    # Convert the event into a 3-tuple of action, params, and event
+    # Convert the event into a 3-tuple of [action, params, and event]
+    # Returns nil if the event cannot be indexed
     def event_action_tuple(event)
 
       action = event.sprintf(@action)
 
+      if Helpers.predict_timestamp_issue_for(@index, event)
+        @logger.error "Cannot index event missing @timestamp to a timestamped index", index: @index, event: event.to_json
+        return nil
+      end
       params = {
         :_id => @document_id ? event.sprintf(@document_id) : nil,
         :_index => event.sprintf(@index),
--- a/lib/logstash/outputs/elasticsearch/helpers.rb
+++ b/lib/logstash/outputs/elasticsearch/helpers.rb
@@ -1,0 +1,12 @@
+# Small functions that don't require a full instance of the plugin should go here.
+# This will promote lightning fast tests.
+module LogStash; module Outputs; class ElasticSearch;
+  module Helpers
+    INDEX_REQUIRING_TIMESTAMP = /%{\+.+\}/
+    def self.predict_timestamp_issue_for(index, event)
+      return false  if event.include?('@timestamp')
+      return true   if index =~ INDEX_REQUIRING_TIMESTAMP
+      false
+    end
+  end
+end ; end ; end
--- a/lib/logstash/outputs/elasticsearch/helpers.rb
+++ b/lib/logstash/outputs/elasticsearch/helpers.rb
@@ -1,5 +1,7 @@
 # Small functions that don't require a full instance of the plugin should go here.
 # This will promote lightning fast tests.
+#
+# Methods are defined right on the module to ensure no context can be shared with the plugin instance :-)
 module LogStash; module Outputs; class ElasticSearch;
   module Helpers
     INDEX_REQUIRING_TIMESTAMP = /%{\+.+\}/
--- a/logstash-output-elasticsearch.gemspec
+++ b/logstash-output-elasticsearch.gemspec
@@ -1,6 +1,6 @@
 Gem::Specification.new do |s|
   s.name            = 'logstash-output-elasticsearch'
-  s.version         = '9.1.4'
+  s.version         = '9.2.0'
   s.licenses        = ['apache-2.0']
   s.summary         = "Stores logs in Elasticsearch"
   s.description     = "This gem is a Logstash plugin required to be installed on top of the Logstash core pipeline using $LS_HOME/bin/logstash-plugin install gemname. This gem is not a stand-alone program"
--- a/spec/unit/outputs/elasticsearch_helpers.rb
+++ b/spec/unit/outputs/elasticsearch_helpers.rb
@@ -1,0 +1,39 @@
+require_relative "../../../spec/es_spec_helper"
+require "logstash/outputs/elasticsearch"
+require "logstash/outputs/elasticsearch/helpers"
+
+describe LogStash::Outputs::ElasticSearch::Helpers do
+  context "detecting if an index name can be determined for an event" do
+    let(:event_with_ts) { LogStash::Event.new() }
+    let(:event_with_no_ts) { LogStash::Event.new().tap { |e| e.remove('@timestamp') } }
+
+    context "when the index pattern doesn't include a timestamp" do
+      ['my-index', 'logstash-%{normal_field_interpolation}'].each do |index|
+        context "where index = #{index.inspect}" do
+          it "should detect no problems for events that have a timestamp" do
+            expect(subject.predict_timestamp_issue_for(index, event_with_ts)).to eq(false)
+          end
+
+          it "should detect no problems for events that do not have a timestamp" do
+            expect(subject.predict_timestamp_issue_for(index, event_with_no_ts)).to eq(false)
+          end
+        end
+      end
+    end
+
+    context "when the index pattern includes a timestamp" do
+      ['logstash-%{+YYYY.MM.dd}', 'logstash-%{+YYYY}', '%{+YYYY.MM.dd}-what'].each do |index|
+        context "where index = #{index.inspect}" do
+          it "should detect no problems for events that have a timestamp" do
+            expect(subject.predict_timestamp_issue_for(index, event_with_ts)).to eq(false)
+          end
+
+          it "should detect a problem for events that do not have a timestamp" do
+            expect(subject.predict_timestamp_issue_for(index, event_with_no_ts)).to eq(true)
+          end
+        end
+      end
+    end
+
+  end
+end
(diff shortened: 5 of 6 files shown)


webmat:
Prior to this PR, the following crashes Logstash with a `FATAL` error:

```
bin/logstash -e "input { stdin {} generator { count => 1 } }
  filter { mutate { remove_field => ['@timestamp'] } }
  output { elasticsearch {}}"
```

After this PR, Logstash simply skips the event, and logs this instead:

```
[2018-05-25T15:10:43,326][ERROR][logstash.outputs.elasticsearch] Cannot index event missing @timestamp to a timestamped index {:index=>"logstash-%{+YYYY.MM.dd}", :event=>"{\"host\":\"matbook-pro.lan\",\"message\":\"Hello world!\",\"sequence\":0,\"@version\":\"1\"}"}
```

webmat:
Oh I had not noticed this suggestion #718, which is an even better behaviour to what this PR puts in place.

Although I'd say since this PR fixes a full-on Logstash crash, we should release +/- as is (following review suggestions), and I'll add #718 to my shortlist. I may even include it in #772.

ph:
Thanks for working on this, I agree this is a bad experience +1 to improving it!

But I think this could be part of a broader discussion:

1. Do we want to allow users to remove `@timestamp`, some use case breaks if the field is removed.  (Index, serialization, or some filter)
2. This look like a problem with the string interpolation feature of the event, maybe the string interpolation should return an error or allow a default when the field is missing.

Concerning 1, I think you are right, so many use cases don't need it.

webmat:
I agree that Logstash's string interpolation for timestamp could be improved. I think there are already tickets for this.

This PR fixes the problem today for LS 5.x and 6.x users, though.

webmat:
Looking at the code around DLQ, I see metrics being tracked for each type of outcome for an event.

My PR here does not do that. This potentially needs to be discussed. Or perhaps we can simply introduce such a metric in #772 and fix this crash today.

webmat:
After multiple discussions on the subject elsewhere, let’s remind ourselves that this PR is about fixing a crash today for all users of 5.x and 6.x.

Whether or not we eventually prevent the removal of @timestamp in a future breaking change (in LS 7.0 or later) is orthogonal to this.

Let’s try to discuss only the merits of this PR in whether it solves the crash in a good way or if needs improvements.

jsvd:
@webmat I think I'd prefer this to remain a situation where logstash crashes fatally. Logstash expects that any event will have `@timestamp` and should operate only under this assumption. With this in mind, we should assume that if `@timestamp` is removed then something went wrong and we should prevent as much data loss as possible. With this PR the es output would massively start dropping data if for some reason the field gets removed.

webmat:
@jsvd What about people importing semi-static CSVs (e.g. zip codes) with Logstash? Should these have a `@timestamp` field?

Is this requirement documented anywhere? Finally, is this a hard requirement for the Logstash => ES scenario only, or should any Logstash configuration (even when it doesn't talk to ElasticSearch) have this hard requirement?

jsvd:
> @jsvd What about people importing semi-static CSVs (e.g. zip codes) with Logstash? Should these have a @timestamp field?
> Is this requirement documented anywhere? Finally, is this a hard requirement for the Logstash => ES scenario only, or should any Logstash configuration (even when it doesn't talk to ElasticSearch) have this hard requirement?

A long time ago `@version` 1 of the LogStash::Event was created and stated that it would always have an `@timestamp`: https://github.com/elastic/logstash/commit/95a5bcb9620785f3abfa427a044b97c017ac3d93#diff-b09e11a609fa0d943a749a72972396ccR14

About documentation, we didn't do a good job there at all. 

That said, I'm up for discussing if we should have a version 2 where we drop this requirement for `@timestamp`, but falls out of scope for this PR/issue (and should include core folks)

webmat:
Gotcha. Should I close this PR, or repurpose it to output a slightly more explicit message when either are missing? The following has the merit of complaining about the missing timestamp:

`[2018-02-22T13:49:42,425][FATAL][logstash.runner ] An unexpected error occurred! {:error=>#<LogStash::Error: timestamp field is missing>`

But this message will not stop people from contacting support and saying Logstash is crashing. It doesn't make clear that this is crashing because missing `@timestamp` is not supported.

So perhaps we could literally output a message that says both `@version` and `@timestamp` are mandatory fields that should never be removed, when sending to ElasticSearch? Perhaps we could even have the error contain a link to documentation about this.

The point of this would be to prevent needless support and SDH activity.

Note that with the discussions I had last week, I was under the impression that requiring `@timestamp` was a potential upcoming requirement, not a requirement that had been in place a long time.

Here's another idea, though (based on #718). We could also consider sending events missing a `@timestamp` to the DLQ, when it's enabled? It's not as hard as a crash, though, so it may fly under the radar and lead to data loss...

jsvd:
> Gotcha. Should I close this PR, or repurpose it to output a slightly more explicit message when either are missing? The following has the merit of complaining about the missing timestamp:

maybe we can create a new PR and refer to this one as the discussion that led to it.

webmat:
New issue created: #779. Let's move the discussion there.

Issue #723, opened 2018-01-10
Custom HTTP headers


andrewherron:
I saw a PR (https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/418) for this that got closed due to not being reviewed. Is there a possibility of getting this reviewed? Currently I need to go through the AWS API Gateway and I can't assign the API Key to the request. I'd prefer this since I can batch records with the bulk indexing and can't bulk index with the http library, though it offers custom HTTP headers.

jordansissel:
> Currently I need to go through the AWS API Gateway 

I am not familiar with this use case. In many cases where there are interfering proxies in front of Elasticsearch, these proxies do not act the same way as Elasticsearch, and as a result, can cause incorrect behavior by a client (Logstash) as the responses and behaviors are different between the proxy and Elasticsearch.

That said, can you detail your use case more?

andrewherron:
Sure, the architecture we'd like is Logstash -> AWS API Gateway -> AWS Lambda -> AWS ES (Behind a VPC).
Because Logstash will be on premises and the ES cluster in a VPC, we need to proxy requests from Logstash through Lambda, and we have a requirement to use the API Gateway in between them. The API Gateway requires an X-Api-Key header in the HTTP request so it can authenticate and have the Gateway pass the request to Lambda. I'd go the route of using the HTTP output plugin, which supports custom headers, but from what I can tell, it won't batch the documents for bulk inserting.

dylanmei:
We do this today, because [Filebeat supports setting the http headers](https://www.elastic.co/guide/en/beats/filebeat/current/elasticsearch-output.html). It would be incredibly useful if Logstash could also support this, as some shippers wish to use Logstash's powerful filter plugins before shipping.

axrayn:
At an organisational level, we'd like to use our existing F5 infrastructure to perform the actual load balancing to the various ECE proxy nodes we have configured. We're currently using HAProxy, but when we're trying to ingest at ~200k e/s from upwards of 30k nodes, HAProxy doesn't cope when things get busy.

Being able to specify the Host header in the Elasticsearch output plugin would mean that we can set the cluster destination on the LS output and not have to rely on setting up an iRule on the F5 and changing it every time we add/remove/modify our clusters.

jsvd:
This was implemented and released in version 9.2.0

PR #782, opened 2018-06-06
Allow configuration of custom headers

jsvd:
fixes https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/723


colinsurprenant:
LGTM - not tested.

Issue #785, opened 2018-06-14
Output fails with 413 Payload Too Large


msvticket:
Output is failing occasionally since ElasticSearch is giving a http status 413 response with the message {"Message":"Request size exceeded 10485760 bytes"}

The actual error in the log is 

` [logstash.outputs.elasticsearch] Encountered a retryable error. Will Retry with exponential backof`

The ElasticSearch is an AWS Elasticsearch service. AWS is apparently restricting the setting of http.max_content_length 10MB for smaller instance types according to this table:
https://docs.aws.amazon.com/elasticsearch-service/latest/developerguide/aes-limits.html#network-limits

Oktaz:
Hitting this issue as well. I think by making the following variable configurable could resolve the issue.

https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/443b07b5e8847f30d24400e57e27fb4a67633798/lib/logstash/outputs/elasticsearch/http_client.rb#L24

RishikeshDarandale:
Any update on these changes?

lifeofguenter:
Same issue with https://github.com/awslabs/logstash-output-amazon_es - ideally this should be configurable.

PR #786, opened 2018-06-14
Resolve 413 Payload Too Large

msvticket:
If we get a http status 413 on a bulk request we try halfing target_bulk_bytes, which is also changed from constant to attribute.

solving issue #785

### DIFF
--- a/lib/logstash/outputs/elasticsearch/common.rb
+++ b/lib/logstash/outputs/elasticsearch/common.rb
@@ -313,6 +313,11 @@ module LogStash; module Outputs; class ElasticSearch;
         # Even though we retry the user should be made aware of these
         if e.response_code == 429
           logger.debug(message, log_hash)
+        # 413 means that the request is to large. So trying to decrease the request size if there
+        # actually where multiple actions in a bulk request.
+        elsif e.response_code == 413 && e.num_actions_in_request > 1
+          @client.target_bulk_bytes /= 2;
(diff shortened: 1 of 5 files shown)


msvticket:
I have submitted the CLA (twice) but I have still not got the mail I'm supposed to get.

As for the integration tests I don't think the remaining failures are related to my changes. If they are I need some pointers as to how they could be.

If any additional tests are needed for the change I've made I'll try to fix that, but I'd like my suggested fix to reviewed first so I don't do it in vain.

untergeek:
Isn't the simplest way forward on this to just change the batch size to be smaller batches instead of changing a bunch of code?  

Batch size is configurable to send smaller batches, which would also effectively address this, would it not?

msvticket:
Yeah, but then everybody encountering this problem needs to realise that this is the problem and add the setting. And the amount of code isn't that different. I wouldn't mind adding a setting for the start value as well though.

msvticket:
Anyway, just lowering the number to something safely lower than 10MB would be fine with me, but preferably somebody else should do it since signing the CLA doesn't work now.

msvticket:
Now the CLA form worked, so CLA is signed.

kares:
Thanks for the PR, this one has fallen a bit into dust.

Since a better functionality is already available (since 10.8.6) where payload is split into batches using the uncompressed size (https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/972).

PR #792, opened 2018-08-11
Add text offering Elasticsearch Service hosted es

karenzone:
Added note offering hosted Elasticsearch Service

### DIFF
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -31,8 +31,15 @@ plugin to version 6.2.5 or higher.
 
 ================================================================================
 
-This plugin is the recommended method of storing logs in Elasticsearch.
-If you plan on using the Kibana web interface, you'll want to use this output.
+If you plan to use the Kibana web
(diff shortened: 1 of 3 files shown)


docs/index.asciidoc
```diff
@@ -31,8 +31,15 @@ plugin to version 6.2.5 or higher.
 
 ================================================================================
 
-This plugin is the recommended method of storing logs in Elasticsearch.
-If you plan on using the Kibana web interface, you'll want to use this output.
+If you are hosting your own Elastic Stack and plan to use the Kibana web
```
dedemorton: I'm not sure about the language you use here because I'm concerned it might make users think the output is only meant for people who are running the stack on their own hardware and using the web interface. I also don't like the word "hosting" here because it conveys meaning beyond the idea that users are running the stack on their own hardware. 

The tip resolves the initial confusion but makes it sound like the Elasticsearch service has its own ingest component built in. Did you take this text from someplace else?

PR #799, opened 2018-10-04
Update build scripts to support downloading ES >= 6.3

robbavey:
Update travis build target
Update artifact downloads to pull oss and standard versions of
 Elasticsearch when version >= 6.3.0
Fixes tests to actually run secure integration tests
Updates version of gradle used to build master version of Elasticsearch
Fixes flaky tests


.travis.yml
```diff
@@ -8,9 +8,13 @@ env:
 - INTEGRATION=true ES_VERSION=2.4.4 TEST_DEBUG=true
 - INTEGRATION=true ES_VERSION=5.6.9 TEST_DEBUG=true
 - INTEGRATION=true ES_VERSION=6.2.4 TEST_DEBUG=true
+- OSS=true INTEGRATION=true ES_VERSION=6.4.2 TEST_DEBUG=true
```
jsvd: what if we instead only tested against the latest 2.x, 5.x, 6.x and master?
This way we only need to install x-pack on 5.x

robbavey: @jsvd Makes sense - I wasn't sure if there was any value in keeping around the last version of 6.x before x-pack was opened up, but I think you're right that there isn't really.

jsvd: also, taking it a bit further, I see no obvious value of testing `OSS=true`, when it is a subset of OSS=false, thoughts?

robbavey: Yeah, I've still got a little clean up to do here - not quite ready for formal review. Early PR was to get tests running on Travis in case that I might have missed locally.

robbavey: But thanks for the early feedback ;)

jsvd: Ahahaha "go away", got it :D

.travis.yml
```diff
@@ -4,13 +4,15 @@ language: ruby
...
-- INTEGRATION=true ES_VERSION=2.4.4 TEST_DEBUG=true
-- INTEGRATION=true ES_VERSION=5.6.9 TEST_DEBUG=true
-- INTEGRATION=true ES_VERSION=6.2.4 TEST_DEBUG=true
-- INTEGRATION=true ES_VERSION=master TEST_DEBUG=true
-- SECURE_INTEGRATION=true INTEGRATION=true ES_VERSION=5.6.9 TEST_DEBUG=true
-- SECURE_INTEGRATION=true INTEGRATION=true ES_VERSION=6.2.4 TEST_DEBUG=true
+- DISTRIBUTION=legacy INTEGRATION=true ES_VERSION=1.7.6 TEST_DEBUG=true
+- DISTRIBUTION=legacy INTEGRATION=true ES_VERSION=2.4.4 TEST_DEBUG=true
+- DISTRIBUTION=legacy INTEGRATION=true ES_VERSION=5.6.9 TEST_DEBUG=true
+- DISTRIBUTION=oss INTEGRATION=true ES_VERSION=6.4.2 TEST_DEBUG=true
```
jsvd: If we removed oss testing (only testing the default package) then we could likely get rid of the whole DISTRIBUTION environment variable, since 
- legacy means < 6.3.0
- legacyxpack means < 6.3.0 + xpack (which we know we need because of SECURE_INTEGRATION=true)
- default >= 6.3.0

robbavey: I wanted to have the `OSS` testing for ILM when it comes - as it is an xpack feature, there will be different tests for open source and xpack flavours of elasticsearch

jsvd: testing the oss won't give us more information as it's a subset of the codebase of the default distro

jsvd: oops, I only now read your comment, got it, makes sense.

robbavey:
Thanks @jsvd. I'll do a follow up PR to fix up the rest to make it look more like the tcp input

PR #800, opened 2018-10-12
Fix: script parameter is lost on retry.

vcostet:
The script parameter was lost on retry.

update_action_builder shouldn't change the args parameter (it deletes _script and _upsert), because it will be reused on next attempts.
This fix will clone 'args' object at the beginning to avoid side effects.


vcostet:
@jsvd Any update on this PR?

jsvd (approved):
LGTM

jsvd:
@vcostet thanks for the contribution. I took your commit, added a version bump to it (retaining your ownership) and I'll publish the change in 10.8.4 today.

PR #805, opened 2018-10-29
Index Lifecycle Management Support

### LINKED ISSUE #798: Index Lifecycle Management  support for Logstash
With Index lifecycle management currently being added to [Elasticsearch](https://github.com/elastic/elasticsearch/issues/29823), support should be added to Logstash along similar lines to the support currently being implemented for [Beats](https://github.com/elastic/beats/issues/7935)

The implementation will require new configuration options, a changed default Elasticsearch template, and a default ILM policy that will be installed if one is not already present.

**Configuration Options**

**New Options** 
`ilm_enabled` [boolean] (optional) - flag for enabling ILM integration. False by default, set to true to turn on ILM integration.
`ilm_write_alias` [string] (required) - write alias used for indexing data. If the write alias doesn’t exist, Logstash will create it and map it to the relevant index/pattern. Consider making this a required field when ilm_enabled is set
`ilm_pattern` [string] (optional) - appends “000001” by default for new index creation, subsequent rollover indices will increment based on this pattern i.e. “000002”
`ilm_policy` [string] (optional) - ILM policy to use, if undefined the default policy will be used.

**Existing Options**
`index` [string] - if ILM is enabled, this config option becomes the new index pattern to be used on new index creation.

Example:
```
output {
  elasticsearch {
    hosts => [...]
    ilm_enabled => true
    ilm_write_alias => “logstash”
    index => “logstash”
    ilm_pattern => “000001”
    ilm_policy => “logstash”
  }
}
```

**Change to Default ES Template**

When ILM is enabled, the default Logstash ES template should contain the ILM integration info which will be used when ILM is enabled.

The ES template should be dynamically configurable based on defined configurations in the ES output - `index` could map to “index_patterns”, `policy` could map to “index.lifecycle.name”, and `write_alias` could map to “index.lifecycle.rollover_alias”.

```
PUT _template/logstash
{
  "index_patterns": ["logstash-*"],
  "settings": {
    "index.lifecycle.name": "logstash",
    "index.lifecycle.rollover_alias": "logstash"
  },
  "mappings": {
    "_doc": {
      ...
    }
  }
}
```

At startup, Logstash will check for the existence of the write alias:

```
HEAD logstash
```

If write alias doesn’t exist yet, then it will create the new index, and then create the respective write alias:
```
PUT logstash-000001
{
  "aliases": {
      "logstash":{
            "is_write_index": true
     }
  }
}
```
From here on all data is sent to the `write_alias` alias, the Elasticsearch ILM implementation will take care of the index rotation behind the scenes.

**New Default ILM Policy**

Logstash will install a new ILM policy if an already existing policy is not explicitly defined in the configuration.

The default policy will have a size limit of `25GB` and a date limit of `30 days`:

```
PUT _ilm/logstash
{
  "policy" : {
    "phases": {
      "hot" : {
        "actions" : {
          "rollover" : {
            "max_size" : "25gb",
            "max_age":"30d"
          }
        }
      }
    }
  }
}
```

Size limit - we generally recommend a 30GB maximum per shard, so 25GB allows us to be close to that recommended limit while building in buffering for when rollovers may be executed late.

Date limit - this is a backstop for users who may have low ingest throughput, which allows them to avoid delete-by-query in the future in the case of misconfiguration. With a default date limit, these users would be able to delete/archive data with monthly indices if desired.

robbavey:
* [X] Add configuration options to plugin
* [X] Implement basic Index Lifecycle Management support
* [X] Create default policy, and add if specified
* [X] Check gracefully handles oss with ILM and pre-ILM versions of Elasticsearch
* [X] Add integration tests
* [x] Add License enforcement
* [ ] Add docs . (will do this in a separate PR)
* [x] Check todos, and clean up code

Resolves #798


lib/logstash/outputs/elasticsearch/default-ilm-policy.json
```diff
@@ -0,0 +1,14 @@
+{
+  "policy" : {
+    "phases": {
+      "hot" : {
+        "actions" : {
+          "rollover" : {
+            "max_size" : "50gb",
+            "max_age":"30d"
```
jsvd: we should also include in the policy that the indices that are created will have the current date - the creation date for the index - to give an idea of what kind of contents exist in the index

LucaWintergerst:
I was wondering if this would also support dynamic index names based on a field of an event, but from what Rob told me this is not possible right now:
```    ilm_write_alias => “%{foobar}” ```

How bad of an idea would it be to do something like the following to support it being dynamic? 
```
my_index_name = event.sprintf(@ilm_index_name_field)
if !@@my_ilm_indices.include?(my_index_name)
  maybe_create_write_alias
  maybe_create_ilm_policy
  @@my_ilm_indices.push(my_index_name)
end
```

The following issues come to mind: 
- logstash is stateless and would have to collect the list of known patterns over and over again after each restart. We could mitigate this by adding an initial list for known indices where we would add aliases and ILMs at startup
- it hurts the performance as we have to check this for every event 
- the array could grow too large and break something. I suppose we could add soft limits to restrict it to a max size
- likely a whole lot more

Just wanted to bring it up as a lot of users rely on the dynamic index creation based on event data

File: lib/logstash/outputs/elasticsearch/default-ilm-policy.json (hunk 1 above)
Comment (robbavey) [reply lib/logstash/outputs/elasticsearch/default-ilm-policy.json]: @jsvd The default index pattern now includes the date - is this what you were referring to?

Comment (jsvd) [reply lib/logstash/outputs/elasticsearch/default-ilm-policy.json]: right, all good 👍

File: lib/logstash/outputs/elasticsearch/common.rb [hunk 2]
```diff
@@ -281,6 +290,7 @@ def safe_bulk(actions)
       sleep_interval = @retry_initial_interval
       begin
         es_actions = actions.map {|action_type, params, event| [action_type, params, event.to_hash]}
+        @logger.error("There are #{actions.count} actions to do")
```
guyboertje: Is this an error situation?

robbavey: @guyboertje I think you are looking at an outdated commit

lib/logstash/outputs/elasticsearch/http_client.rb
```diff
@@ -348,6 +348,37 @@ def template_put(name, template)
       @pool.put(path, nil, LogStash::Json.dump(template))
     end
 
+    def write_alias_exists?(name)
```
guyboertje: Consider refactoring this to `es_resource_exists?` and calling it from `template _exists?` as well as `maybe_create_ilm_policy` and `maybe_create_write_alias ` so that these last two read
`plugin.client.es_resource_exists?(ilm_write_alias)` as it is quite clear what you are checking the existence of.

Also, do you need interpolation and the creation of a new string object in `@pool.head("#{name}")`?

lib/logstash/outputs/elasticsearch/http_client.rb
```diff
@@ -348,6 +348,37 @@ def template_put(name, template)
...
+    end
+
+
+    def ilm_policy_put(name, policy)
+      path = "_ilm/#{name}"
+      logger.info("Creating policy #{policy} at #{path}")
+      begin
+        response = @pool.put(path, nil, LogStash::Json.dump(policy))
+      rescue => e
+        puts e.inspect
```
guyboertje: log these instead?

lib/logstash/outputs/elasticsearch/ilm_manager.rb
```diff
@@ -0,0 +1,64 @@
...
+    end
+
+    def self.maybe_create_ilm_policy(plugin, ilm_policy)
+      plugin.client.ilm_policy_put(ilm_policy, default_policy_payload) if ilm_policy == DEFAULT_POLICY && !plugin.client.ilm_policy_exists?(ilm_policy)
+    end
+
+    def self.write_alias_payload(plugin)
+      {
+        "aliases" => {
+            plugin.ilm_write_alias =>{
```
guyboertje: nitpick: A space after the hash rocket would be nice

PR #809, opened 2018-11-05
reduced logging verbosity and unecessary upper case error message

jsvd:
There are multiple logging statements in this plugin that are either superfluous or could be moved to a more verbose level. This PR changes a couple of them to reduce overall logging fatigue when using the ES output.


jsvd:
Thanks @robbavey

PR #812, opened 2018-11-08
Fix support for Elasticsearch 7.x

robbavey:
This commit fixes issues with supporting Elasticsearch 7.x

This handles the removal of the '_routing' field name and
'_retry_on_conflict' update action aliases in place of 'routing' and
'retry_on_conflict' respectively in Elasticsearch 7.x.

Also fixes integration tests to handle these changes and
the renaming of doc type from 'doc' to '_doc' and the mapping
name change from `_default_` to '_doc'.


ci/run.sh
```diff
@@ -132,6 +132,9 @@ else
   fi
 
   case "$ES_VERSION" in
+  # ES_VERSION prefixed with LATEST-SNAPSHOT- will interrogate the artifacts API to find the latest version that matches
```
jsvd: this documentation seems out of place in this PR, was it on purpose?

lib/logstash/outputs/elasticsearch/common.rb
```diff
@@ -114,6 +114,15 @@ def maximum_seen_major_version
       client.maximum_seen_major_version
     end
 
+
+    def routing_field_name
+      maximum_seen_major_version >= 6 ? :routing : :_routing
```
jsvd: since we'll use the non underscored version in 6.x and 7.x here, can we confirm that all versions of 6.x support `routing` too?

ci/run.sh
jsvd: this seems out of place in this PR, is it an intentional change?

robbavey:
@jsvd Thanks for the LGTM - do you think it makes sense to bump the version as a patch release or a minor release?

jsvd:
I think patch is ok, the actual fix for es7.x compat is rather small, the rest are testing changes and ci a travis yml tweak

PR #820, opened 2018-12-14
Documentation for ILM feature

robbavey:

### DIFF
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -93,6 +93,56 @@ re-indexed to Elasticsearch. If the DLQ is not enabled, and a mapping error
 happens, the problem is logged as a warning, and the event is dropped. See
 <<dead-letter-queues>> for more information about processing events in the DLQ.
 
+[[index-lifecycle-management]]
+==== Index Lifecycle Management (Beta)
+
+beta[]
+
+[NOTE]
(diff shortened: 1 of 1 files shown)


docs/index.asciidoc
```diff
@@ -93,6 +93,56 @@ re-indexed to Elasticsearch. If the DLQ is not enabled, and a mapping error
 happens, the problem is logged as a warning, and the event is dropped. See
 <<dead-letter-queues>> for more information about processing events in the DLQ.
 
+[[index-lifecycle-management]]
```
karenzone: For plugins, we need to define anchors using attributes, like this:
`[id="plugins-{type}s-{plugin}-ilm"]`

Then reference them like this:
`<<plugins-{type}s-{plugin}-ilm>>`

This approach makes the Versioned Plugin Reference possible. Otherwise, the doc build fails with `element section: validity error : ID already defined` errors.  
Gotcha: These will work the first time. When we bump the plugin version, the doc build will break. 

Additional comments outside of recent changes:
-  Same note about anchor format applies for [[dlq-policy]] around line 85 
- I'm guessing that <<dead-letter-queue>> references the LS Reference Guide? If so, we'll need to hard code it or use the attribute like this:
`{logstash-ref}/dead-letter-queues.html`

Issue #819, opened 2018-12-14
Wrong documentation regards codec usage in ES output


marekduciuc:
isseu [420](https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/420) is closed with note that codec is nto used for elasticsearch output plugin but documentation is not in line with this: 
https://www.elastic.co/guide/en/logstash/current/plugins-outputs-elasticsearch.html#plugins-outputs-elasticsearch-common-options

PR #822, opened 2018-12-20
docs: flag as `no_codec` to reduce confusion

yaauie:
Resolves: logstash-plugins/logstash-output-elasticsearch#819


karenzone:
@yaauie   I'm glad to see that `no_codec` is getting some use. Thanks!

PR #824, opened 2019-01-10
remove obsolete fields and obsolete document_type

jsvd:
- remove obsolete `flush_size` and `idle_flush_time`
- obsolete deprecated `document_type`
- remove support for parent child (still support join data type) since we don't support multiple document types any more


jsvd:
replaced by https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/828

PR #826, opened 2019-01-22
Fix tests for 7.x after type removal

robbavey:


spec/es_spec_helper.rb
```diff
@@ -32,11 +32,15 @@ def todays_date
 
   def mapping_name
```
jsvd: This method becomes a bit less useful now, could we maybe replace it with just

```ruby
def default_mapping_from_mappings(mappings)
  if ESHelper.es_version_satisfies?(">=7")
    mappings
  else
    mappings["_default_"]  
  end
end
```

and then:

```ruby
  def field_properties_from_template(template_name, field)
    mappings = @es.indices.get_template(name: template_name)[template_name]["mappings"]
    mapping = default_mapping_from_mappings(mappings)
    mapping["properties"][field]["properties"]
  end
```

PR #831, opened 2019-01-25
Handle escaping of index names

robbavey:
Ensure that path entries containing '/' characters, such as those using
date math are handled correctly.

Also adds an integration test target using logstash source


robbavey:
@jsvd - This PR now has integration tests that will test ilm (and other elasticsearch functions) against latest logstash source rather than `5.4.0` that was being pulled in previously

jsvd:
@robbavey sorry, not yet, I'm testing it today

PR #833, opened 2019-01-25
Resolve 413 Payload Too Large

### LINKED ISSUE #785: Output fails with 413 Payload Too Large

msvticket:
In particular AWS ElasticSearch often allows maximum 10MiB HTTP request payloads

Fixes #785

### DIFF
--- a/lib/logstash/outputs/elasticsearch/http_client.rb
+++ b/lib/logstash/outputs/elasticsearch/http_client.rb
@@ -12,16 +12,15 @@ module LogStash; module Outputs; class ElasticSearch;
   # there really isn't a good reason to configure it.
   #
   # The criteria used are:
-  # 1. We need a number that's less than 100MiB because ES
-  #    won't accept bulks larger than that.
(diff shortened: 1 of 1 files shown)


msvticket:
It would be good to have this merged to 7.x as well. But I can do another pull request for this if preferred.

msvticket:
I still like the solution in #786, but if that was deemed to complicated maybe this can be merged instead.

msvticket:
The failed builds doesn't seem to have anything to do with the change.

PR #857, opened 2019-04-02
Look up cluster_uuid

cachedout:
One of two PRs necessary to address https://github.com/elastic/logstash/issues/10602

### DIFF
--- a/lib/logstash/outputs/elasticsearch/common.rb
+++ b/lib/logstash/outputs/elasticsearch/common.rb
@@ -47,6 +47,7 @@ module LogStash; module Outputs; class ElasticSearch;
           sleep_interval = next_sleep_interval(sleep_interval)
         end
         if successful_connection?
+          discover_cluster_uuid
           install_template
           setup_ilm if ilm_in_use?
         end
@@ -128,6 +129,13 @@ module LogStash; module Outputs; class ElasticSearch;
     def install_template
       TemplateManager.install_template(self)
       @template_installed.make_true
+    end
+
+    def discover_cluster_uuid
+      cluster_info = @client.get('/')
+      unless LogStash::SETTINGS.registered?(@id + ".cluster_uuid")
+        LogStash::SETTINGS.register(LogStash::Setting::String.new(@id + ".cluster_uuid", cluster_info["cluster_uuid"]))
+      end
     end
 
     def check_action_validity
--- a/lib/logstash/outputs/elasticsearch/common.rb
+++ b/lib/logstash/outputs/elasticsearch/common.rb
@@ -132,9 +132,9 @@ module LogStash; module Outputs; class ElasticSearch;
     end
 
     def discover_cluster_uuid
-      cluster_info = @client.get('/')
-      unless LogStash::SETTINGS.registered?(@id + ".cluster_uuid")
-        LogStash::SETTINGS.register(LogStash::Setting::String.new(@id + ".cluster_uuid", cluster_info["cluster_uuid"]))
+      if defined?(plugin_metadata)
+        cluster_info = @client.get('/')
+        plugin_metadata.set(:cluster_uuid, cluster_info['cluster_uuid'])
       end
     end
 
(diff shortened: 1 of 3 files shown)


ycombinator:
@cachedout I've never built an LS plugin and tested it in conjunction with another LS core PR. Would you mind posting some instructions on how I can essentially run this PR along with https://github.com/elastic/logstash/pull/10561? Thanks!

cachedout:
@ycombinator I did my development by editing the installed gem directly, the instructions here should allow for testing in your environment after checking out this PR into a local branch: https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/master/README.md#2-running-your-unpublished-plugin-in-logstash

lib/logstash/outputs/elasticsearch/common.rb
```diff
@@ -130,6 +131,13 @@ def install_template
       @template_installed.make_true
     end
 
+    def discover_cluster_uuid
+      cluster_info = @client.get('/')
+      unless LogStash::SETTINGS.registered?(@id + ".cluster_uuid")
```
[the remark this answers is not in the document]
cachedout: @jsvd recommended using the settings object to store the data after we discussed it on a call. We discussed a few other approaches as well. Perhaps he can comment here and we can find another route?

yaauie:
Relevant PR on Logstash core is https://github.com/elastic/logstash/pull/10636

With it, the implementation of `LogStash::Outputs::ElasticSearch::Common#discover_cluster_uuid` becomes something like:

~~~ ruby
    def discover_cluster_uuid
      if defined?(plugin_metadata)
        cluster_info = @client.get('/')
        plugin_metadata.set(:cluster_uuid, cluster_info['cluster_uuid'])
      end
    end
~~~

And extracting this metadata in your other PR becomes something like:

~~~ ruby
  def resolve_cluster_uuids
    outputs.each_with_object(Set.new) do |output, cluster_uuids|
      if LogStash::PluginMetadata.exists?(output.id)
        cluster_uuids << LogStash::PluginMetadata.for_plugin(output.id).get(:cluster_uuid)
      end
    end.to_a.compact
  end
~~~

cachedout:
jenkins, re-test this

lib/logstash/outputs/elasticsearch/common.rb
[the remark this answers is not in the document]
cachedout: This is now resolved using your new metadata approach.

cachedout:
The failing tests also appear to be failing on the master branch for me.

I had to also fix up some other tests which were failing because they depended on a certain sequence of log events instead of being fully isolated.

@yaauie can you re-review this?

lib/logstash/outputs/elasticsearch/common.rb
```diff
@@ -130,6 +131,17 @@ def install_template
       @template_installed.make_true
     end
 
+    def discover_cluster_uuid
```
jsvd: a good way to remove the amount of indentation in a method is to return early for known situations. We can also move the catch into the method definition as it's pretty small. in this case meaning we can have instead:

```ruby
def discover_cluster_uuid
  return unless if defined?(plugin_metadata)
  cluster_info = client.get('/')
  plugin_metadata.set(:cluster_uuid, cluster_info['cluster_uuid'])
rescue => e
  @logger.error("Unable to retrieve elasticsearch cluster uuid", error => e.message)
  nil
end
```

cachedout: The problem with this is that even with the changes I have made to the DLQ tests, emitting *any* logging message at this point will break that entire series of tests, which is why it was `nil` to begin with. Certainly, the DLQ tests should be further addressed to prevent this problem but I'm not sure if we want to do that in this PR or a follow-up. Thoughts?

jsvd: ok let's comment out the logger entry and leave a `# TODO introducing this logging message breaks many tests that need refactoring`

cachedout:
Changes pushed. This is still failing a local test but I believe this is a problem unrelated to this PR. It also fails in the master branch and is due to a problem that occurs when attempting to fetch the ES version with a path specified.

As such, I believe this is ready for merge.

PR #861, opened 2019-04-16
Fix issue with templates being erroneously overwritten

robbavey:
When customers were managing their own templates, but not using ilm,
templates could still be overwritten. This commit fixes the logic
to only overwrite templates when ilm is in use

### DIFF
--- a/lib/logstash/outputs/elasticsearch/ilm.rb
+++ b/lib/logstash/outputs/elasticsearch/ilm.rb
@@ -4,20 +4,22 @@ module LogStash; module Outputs; class ElasticSearch
     ILM_POLICY_PATH = "default-ilm-policy.json"
 
     def setup_ilm
-      return unless ilm_enabled?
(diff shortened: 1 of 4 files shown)


lib/logstash/outputs/elasticsearch/ilm.rb
```diff
@@ -4,20 +4,22 @@ module Ilm
...
-      end
     end
 
     def default_rollover_alias?(rollover_alias)
       rollover_alias == LogStash::Outputs::ElasticSearch::DEFAULT_ROLLOVER_ALIAS
     end
 
-    def ilm_enabled?
+    def ilm_alias_set?
+      default_index?(@index) || !default_rollover_alias?(@ilm_rollover_alias)
```
jsvd: why is this alias predicate negated? isn't the test about alias being set to `@ilm_rollover_alias`?

robbavey: It is about being set to something other than the default rollover alias

Issue #493, opened 2016-10-13
output plugins load balancing via hosts array not aware of other logstash instances.


kobybr:
- logstash-2.4.0
- centos 7.2

Say I have three logstash servers loadbalanced behind an F5. All three servers are identical, and contain an elasticsearch output plugin with a host array of something like: hosts => ["host1","host2",host3","host4"].

What I'm trying to prevent is a scenerio where the three logstash servers would get in a rhythm of rotating through the same hosts at the same time, thus swamping one elasticsearch host, while the others go unused.

From what I understand, an individual instance will round-robin through the array of available elasticsearch hosts.   I believe I have experienced this scenario and was likely the cause of a backlog of unprocessed/dropped udp packets.   I alleviated the issue by having each logstash instance only contain unique sets of hosts.  i.e.   logstash-1 would contain hosts => ["host1","host2"] and logstash-2 would contain hosts => ["host3","host4"].   This would not be the most ideal way of handling. 

Am I missing something?  Thoughts?

Thanks

jsvd:
a connection is picked per bulk request so it's unlikely that three instances an any point will be hammering the same server for a large period of time unless each bulk request has very large documents (in the order of megabytes) or the ids for events are being set explicitly are aren't well balanced causing hotspots due to routing issues.

I'm closing this issue as I don't believe this is an issue, please reopen or open a new one if you're seeing a similar problem

Issue #274, opened 2015-10-16
Verify this plugin round-robins for every bulk request


suyograo:
The elasticsearch-ruby client is supposed to round-robin requests to the nodes specified in the `hosts` setting. From internal testing it seems like one of the node is getting hit more often than the other. The setting used in testing was `[localhost:9200, localhost:9201]`

![round-robin](https://cloud.githubusercontent.com/assets/1595958/10546440/d17e1d6e-73e2-11e5-9a89-21ed6e990bea.png)

y-axis in the graph is the bytes-in to a port. Data in bulk request used for testing was the same size, so we should not see 2:1 requests distribution to hosts, should have almost equal distribution

gingerwizard:
The following test was performed:

1. Two node cluster running on 9202 and 9201 respectively.
2. Logstash configuration indexed data over HTTP. Following output configuration:

output {
    elasticsearch {
    	host => ["localhost:9202","localhost:9201"]
    	cluster => "test-cluster"
    	protocol => "http"
    	workers => 1
    	index => "test"
    	flush_size => 500
	}
}

3. Packetbeat monitoring HTTP ports 9202, 9201. Packetbeat configured to index into separate Elastic cluster on 9200 so as to not skew results.

4. Kibana configured to use 9200 packetbeat index. Only requests with HTTP path path:"/_bulk" considered in the following visuals.

5. 4 visuals as shown below.  
 - Sum of bytes over time per port
- Total requests over time per port
- Total Requests port 9201
- Total Requests port 9202
- Sum of bytes port 9201
- Sum of bytes port 9201
 
6. The same log line was sent to logstash using nc from a file containing 1m duplicate lines.  Ensures packet data is of equal size. No filters in logstash.

![image](https://cloud.githubusercontent.com/assets/12695796/10548156/bef2bb66-7430-11e5-94bb-b083372c9ce2.png)


The above shows a balance of requests. Issue can be closed.

Thanks
Dale

gingerwizard:
Wrong image.

![image](https://cloud.githubusercontent.com/assets/12695796/10548204/e4c9960c-7430-11e5-96a4-ceae35691904.png)

TinLe:
There is a related bug where in the first two hosts listed in hosts => get more connections and consumed lots more filehandles.

It's easily reproduced in logstash versions 2.1.0 - 2.1.3, and logstash versions 2.2.0 - 2.2.2

PR #866, opened 2019-05-03
Clarify load balance behavior if array of hosts is used

jakommo:
I've seen the question, how the output load balances, coming up every now and then and the docs should have the answer.

I'd still like a logstash developer to review and double check that it still is Round-robin per bulk in the current version of the plugin. 

My assumption is based on https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/274 and https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/493#issuecomment-389808368, but the first one is from 2015 and the second onw is from 2018, but with Logstash 2.4.


ripper2hl:
Have there been any updates on that?

PR #867, opened 2019-06-03
[DOC] Adding details for template installation

lucabelluccini:
Clarify Template file is not monitored.
Clarify when the Template is installed and what conditions apply.


lucabelluccini:
This change should be applied to all docs back 6.x and 7.x

karenzone:
Docs build cleanly and look good. Requesting a quick tech review to double-check content.

yaauie:
Instead of bouncing it back to @lucabelluccini for changelog/version bump, I've opened https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/898 which includes this commit.

PR #871, opened 2019-06-27
Create elasticsearch-template-es8x.json

jsvd:
now that es master points to 8.0 we need a template for that version.


jsvd:
@jakelandis are you aware of any specificity we may need to include right away in the 8.x template compared to 7.x?

PR #873, opened 2019-07-14
Update testing to docker

jsvd:
* create a docker image one for logstash
* if it's an integration test create another for Elasticsearch
* reduces test matrix to same version testing
* integration tests run against host 'elasticsearch' so tests were updated
* certificates were recreated to validate the 'elasticsearch' host
* removed any prior dependency to running tests, only docker is needed
* it's possible to pass options to rspec using "SPEC_OPTS=' ...' ./ci/docker-run.sh`

Check the .travis.yml to see how to run the tests


jsvd:
failure is due to lack of 8.x support in sniffing, fixed in #874

.travis.yml
```diff
@@ -1,53 +1,19 @@
...
-script: ci/build.sh
-jdk: openjdk8
-before_install: gem install bundler -v '< 2'
+  - INTEGRATION=false ELASTIC_STACK_VERSION=5.x
+  - INTEGRATION=false ELASTIC_STACK_VERSION=6.x
+  - INTEGRATION=false ELASTIC_STACK_VERSION=7.x
+  - DISTRIBUTION=default INTEGRATION=true ELASTIC_STACK_VERSION=6.x
+  - DISTRIBUTION=default INTEGRATION=true ELASTIC_STACK_VERSION=7.x
+  - DISTRIBUTION=default INTEGRATION=true ELASTIC_STACK_VERSION=7.x SNAPSHOT=true
+  - DISTRIBUTION=default INTEGRATION=true ELASTIC_STACK_VERSION=8.x SNAPSHOT=true
```
robbavey: Just making sure that it is intentional to fail the builds to surface issues ASAP for visibility due to `8.x` incompatibilities given that we are so early in the `8.x` release cycle.

jsvd: It is intentional indeed, allow failures setting in travis can be useful but more often not if I see those fail I'll start caring less and less about them until we're on top of the release, may as well have these failures up front

PR #875, opened 2019-08-06
check if ES is licensed before pushing out events to it

jsvd:
when running a default logstash distribution we should error out if the elasticsearch distribution is oss. this makes logstash behave like beats

### DIFF
--- a/lib/logstash/outputs/elasticsearch/http_client/pool.rb
+++ b/lib/logstash/outputs/elasticsearch/http_client/pool.rb
@@ -31,6 +31,7 @@ module LogStash; module Outputs; class ElasticSearch; class HttpClient;
     attr_reader :logger, :adapter, :sniffing, :sniffer_delay, :resurrect_delay, :healthcheck_path, :sniffing_path, :bulk_path
 
     ROOT_URI_PATH = '/'.freeze
+    LICENSE_PATH = '/_license'.freeze
 
     DEFAULT_OPTIONS = {
       :healthcheck_path => ROOT_URI_PATH,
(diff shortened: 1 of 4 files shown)


spec/unit/outputs/elasticsearch/http_client/pool_spec.rb
```diff
@@ -235,4 +240,31 @@
...
+          subject.update_initial_urls
+        end
+      end
+      context "if ES returns a valid license" do
+        let(:valid_license) { true }
+        it "marks the url as active" do
+          subject.update_initial_urls
+          expect(subject.alive_urls_count).to eq(1)
+        end
+      end
```
robbavey: Does it make sense to have a test proving the license deprecation is not logged in the license is there?

lib/logstash/outputs/elasticsearch/http_client/pool.rb
```diff
@@ -101,7 +110,7 @@ def in_use_connections
     end
 
     def alive_urls_count
-      @state_mutex.synchronize { @url_info.values.select {|v| !v[:state] == :alive }.count }
```
robbavey: oops...

spec/unit/outputs/elasticsearch/http_client/pool_spec.rb
```diff
@@ -235,4 +240,31 @@
...
+          expect(subject).to receive(:log_license_deprecation_warn).once
+          subject.update_initial_urls
+        end
+      end
+      context "if ES returns a valid license" do
+        let(:valid_license) { true }
+        it "marks the url as active" do
+          subject.update_initial_urls
+          expect(subject.alive_urls_count).to eq(1)
+        end
```
jsvd:
```suggestion
        end
        it "logs a warning" do
          expect(subject).to receive(:log_license_deprecation_warn)
          subject.update_initial_urls
        end
```

jsvd:
many thanks @robbavey

PR #880, opened 2019-09-18
Docs: Clarify default 'document_type' for 7.x indices

cbuescher:
I was confused that the default document type was "doc" not "_docs" in 6.x.
From https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/master/lib/logstash/outputs/elasticsearch/common.rb#L269 it appears this has changed from
6.x to 7.x, to clarify this it would be good to add the behaviour that probably started with 7.0?

### DIFF
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -303,6 +303,7 @@ This sets the document type to write events to. Generally you should try to writ
 similar events to the same 'type'. String expansion `%{foo}` works here.
 If you don't set a value for this option:
 
+- for elasticsearch clusters 7.x and above: the value of '_doc' will be used. 
 - for elasticsearch clusters 6.x and above: the value of 'doc' will be used;
 - for elasticsearch clusters 5.x and below: the event's 'type' field will be used, if the field is not present the value of 'doc' will be used.
 


cbuescher:
Looks like this has been taken care of on master already.

cbuescher:
Actually, I was wrong, at least the docs at https://www.elastic.co/guide/en/logstash/7.4/plugins-outputs-elasticsearch.html still don't mention the change in behaviour with version 7. Reopening.

cbuescher:
This is confusing. It looks like https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/892 should have added the right version 7/8 type info. The docs I see online still don't reflect that (as of today). Maybe they don't get built so often? I'll close again because the docs source seems to be what like what was suggested in this PR. Sorry for the noise.

jsvd:
Hi @cbuescher, version 10.2.2 of the plugin ships with this new documentation and we can see it in https://www.elastic.co/guide/en/logstash-versioned-plugins/current/v10.2.2-plugins-outputs-elasticsearch.html#v10.2.2-plugins-outputs-elasticsearch-document_type

These docs will only show up when this version in included in a Logstash release (so if there's a new 7.4.3 or in 7.5.1.), since documentation of plugins is tied to the plugin version itself.

I know this isn't ideal in this scenario, and we can "patch" docs in previous versions directly in https://github.com/elastic/logstash-docs/tree/7.4/docs/plugins/outputs, but it won't likely be done any time soon.

PR #881, opened 2019-09-25
Corrected documentation for `action => update`

robin13:
I think that the "update a document if not already present" only works with the `doc_as_upsert` option.  The `upsert` option adds data to the document if the document does not exist I think.
Some more explicit documentation for both `upsert` and `doc_as_upsert` would also be helpful


karenzone:
@robbavey Please verify the content, and I'll do the rest. Thanks!

karenzone:
Added secondary request for more explicit documentation for both upsert and doc_as_upsert in #885

PR #883, opened 2019-10-07
[Doc]Update wording on ilm_enabled setting and bump to v10.2.1

karenzone:
Wording fix in response to request from Support


karenzone:
Note to self (@karenzone):  Work with @robbavey to verify content in #881 and group it with this version bump

karenzone:
**Published**
Screen Shot 2019-10-09 at 12 16 07 PM

PR #884, opened 2019-10-07
[Doc]Add note about _type setting change from doc to _doc

karenzone:
Call out _type change from `doc` to `_doc` in documentation

### DIFF
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -24,16 +24,16 @@ include::{include_path}/plugin_header.asciidoc[]
 .Compatibility Note
 [NOTE]
 ================================================================================
-The `_type` setting changed from `doc` to `_doc` in Elasticsearch 7.0.
+When connected to Elasticsearch 7.x, modern versions of this plugin
+use the required `_doc` document-type when inserting documents.
 
-If you are using an earlier version of Logstash, upgrade to
-version 6.8 to pick up changes to the Elasticsearch index template before you
-upgrade to version 7.0 (or later). 
+If you are using an earlier version of Logstash and wish to connect to
+Elasticsearch 7.x, first upgrade Logstash to version 6.8 to ensure it
+picks up changes to the Elasticsearch index template.
 
-After upgrading to 6.8, use the {ref}/indices-templates.html#getting[Index
-Template API] to verify that the 6.8 index template has been created in
-Elasticsearch.
-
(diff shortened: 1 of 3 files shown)


karenzone:
@yaauie @jsvd  Am I capturing the right messaging here?  I mentioned the upgrade implications, but I'm wondering if there's more the user would want/need to know.

docs/index.asciidoc
```diff
@@ -21,6 +21,21 @@ include::{include_path}/plugin_header.asciidoc[]
...
+The `_type` setting changed from `doc` to `_doc` in Elasticsearch 7.0.
+
+If you are using an earlier version of Logstash, upgrade to
+version 6.8 to pick up changes to the Elasticsearch index template before you
+upgrade to version 7.0 (or later). 
+
+After upgrading to 6.8, use the {ref}/indices-templates.html#getting[Index
+Template API] to verify that the 6.8 index template has been created in
+Elasticsearch.
+
```
yaauie: This is a bit tricky to communicate, because this Logstash plugin's behaviour depends primarily on the version of _Elasticsearch_ it is connected to, not on the version of Logstash in which it is running. Of course users will need an up-to-date version of this plugin that knows how to talk with Elasticsearch 7.x, which they can get with the latest Logstash 6.8.x, but they need to point it to an Elasticsearch 7.x cluster in order for the change in behaviour to be triggered.

This happens naturally when users follow our [Stack Upgrade Guide](https://www.elastic.co/guide/en/elastic-stack/current/upgrading-elastic-stack.html), which instructs users to upgrade their Elasticsearch before upgrading Logstash, so long as they also follow the guidance to upgrade the _whole stack_ to the last minor on 6.x before upgrading any to 7.x.

```suggestion
When connected to Elasticsearch 7.x, modern versions of this plugin
use the required `_doc` document-type when inserting documents.

If you are using an earlier version of Logstash and wish to connect to
Elasticsearch 7.x, first upgrade Logstash to version 6.8 to ensure it
picks up changes to the Elasticsearch index template.

If you are using a custom <<plugins-{type}s-{plugin}-template>>,
ensure your template uses the `_doc` document-type before
connecting to Elasticsearch 7.x.
```

karenzone:
@yaauie Thanks for the assist.

karenzone:
**PUBLISHED:**
Screen Shot 2020-03-27 at 12 04 51 PM

PR #895, opened 2019-11-13
[DOC] Add best practices on single/multiple Elasticsearch outputs

lucabelluccini:
This PR adds some background on preferring a single Elasticsearch output vs multiple Elasticsearch outputs when events are routed to multiple indices within the same Cluster


karenzone:
Docs build cleanly and look good. Requesting a quick tech review to double-check content.

karenzone:
Thanks for the review, @jsvd.  I have a few unrelated (minor) edits I want to make. I'll do those, and bump the version when I do.

PR #898, opened 2019-11-18
Template installation docs

yaauie:
Includes https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/867 with changelog and version bump.


docs/index.asciidoc
```diff
@@ -707,6 +707,9 @@ https://www.cs.utexas.edu/~shmat/shmat_ccs12.pdf
 You can set the path to your own template here, if you so desire.
 If not set, the included template will be used.
 
+Note that Logstash will not monitor the template file for changes.
+The template file will be read when its installation is triggered.
```
jsvd:
```suggestion
The template file will be read when its installation is triggered, which happens at the startup of a pipeline.
```

Issue #893, opened 2019-10-24
Support Elastic Cloud id


jsvd:
Currently, only core features of logstash like monitoring [support the configuration of cloud id](https://www.elastic.co/guide/en/cloud/current/ec-cloud-id.html).

I'd be nice if elastic cloud users can put the cloud id instead of host/port/user/auth

sgreszcz:
Is there anyway of doing this?  Cloud ID works fine with beats, but not Logstash. Even the documentation is confusing about Logstash and cloud ID: https://www.elastic.co/guide/en/cloud/current/ec-cloud-id.html

PR #906, opened 2019-12-16
Feat: Added support for cloud_id and cloud_auth

kares:
should be fairly straight-forward: 
- **cloud_id** is a substitute for the *hosts* setting (one can not use both)
- **cloud_auth** is an *user/password* alternative

resolves: https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/893

~~TODO documentation for new options~~

### DIFF
--- a/lib/logstash/outputs/elasticsearch.rb
+++ b/lib/logstash/outputs/elasticsearch.rb
@@ -125,7 +125,7 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
   # Cloud authentication string ("<username>:<password>" format) is an alternative for the `user`/`password` configuration.
   #
   # For mode details, check out the https://www.elastic.co/guide/en/logstash/current/connecting-to-cloud.html#_cloud_auth[cloud documentation]
-  config :cloud_auth, :validate => :string
+  config :cloud_auth, :validate => :password
 
   # HTTP Path at which the Elasticsearch server lives. Use this if you must run Elasticsearch behind a proxy that remaps
   # the root path for the Elasticsearch HTTP API lives.
--- a/lib/logstash/outputs/elasticsearch.rb
+++ b/lib/logstash/outputs/elasticsearch.rb
@@ -17,7 +17,7 @@ require "forwardable"
 # called `http.content_type.required`. If this option is set to `true`, and you
 # are using Logstash 2.4 through 5.2, you need to update the Elasticsearch output
 # plugin to version 6.2.5 or higher.
-# 
+#
 # ================================================================================
 #
 # This plugin is the recommended method of storing logs in Elasticsearch.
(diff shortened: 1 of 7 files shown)


lib/logstash/outputs/elasticsearch.rb
```diff
@@ -122,6 +122,11 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
   # Password to authenticate to a secure Elasticsearch cluster
   config :password, :validate => :password
 
+  # Cloud authentication string ("<username>:<password>" format) is an alternative for the `user`/`password` configuration.
+  #
+  # For mode details, check out the https://www.elastic.co/guide/en/logstash/current/connecting-to-cloud.html#_cloud_auth[cloud documentation]
+  config :cloud_auth, :validate => :string
```
robbavey: This should be `:validate => :password` as it contains password information

```suggestion
  config :cloud_auth, :validate => :password
```

lib/logstash/outputs/elasticsearch/common.rb
```diff
@@ -115,6 +117,64 @@ def setup_hosts
...
+      end
+      @user, @password = parse_user_password_from_cloud_auth(@cloud_auth)
+      params['user'], params['password'] = @user, @password
+    end
+
+    def parse_host_uri_from_cloud_id(cloud_id)
+      begin # might not be available on older LS
+        require 'logstash/util/cloud_setting_id'
+      rescue LoadError
+        raise LogStash::ConfigurationError, 'The cloud.id setting is not supported by LogStash, please set hosts instead.'
```
robbavey: Nits: `cloud.id` should read `cloud_id` to reflect configuration property name
          Capitalization of "LogStash" should read Logstash
          Maybe state that the setting is not supported by this version of Logstash, as later versions do support it.

lib/logstash/outputs/elasticsearch/common.rb
```diff
@@ -115,6 +117,64 @@ def setup_hosts
...
+      cloud_uri = "#{cloud_id.elasticsearch_scheme}://#{cloud_id.elasticsearch_host}"
+      LogStash::Util::SafeURI.new(cloud_uri)
+    end
+    private :parse_host_uri_from_cloud_id
+
+    def parse_user_password_from_cloud_auth(cloud_auth)
+      begin # might not be available on older LS
+        require 'logstash/util/cloud_setting_auth'
+      rescue LoadError
+        raise LogStash::ConfigurationError, 'The cloud.auth setting is not supported by LogStash, please set user/password instead.'
```
robbavey: Nits: cloud.id should read cloud_id to reflect configuration property name
Capitalization of "LogStash" should read Logstash
Maybe state that the setting is not supported by this version of Logstash, as later versions do support it.

kares:
circling back here, as this was promised to eventually land in next LS release ...
did try generating docs locally and seemed fine, 
Rob's docs suggestions are still relevant but those won't be addressed here.

PR #911, opened 2020-01-13
Docs: fix typos introduced recently

kares:
:scissors: - align with filter/input ES plugin docs


kares:
Hey @karenzone do we need a release bump to get these into the generated docs?

PR #912, opened 2020-01-16
Fix: handle proxy => '' as if none was set

kares:
a convenient feature when users are using env variables for proxy
e.g. `proxy => ${LS_PROXY:}`. 
empty proxy host settings are common to be ignored by libraries ...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 10.3.1
+ - Fix: handle proxy => '' as if none was set [#912](https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/912)
+
 ## 10.3.0
   - Feat: Added support for cloud_id and cloud_auth [#906](https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/906)
 
--- a/lib/logstash/outputs/elasticsearch.rb
+++ b/lib/logstash/outputs/elasticsearch.rb
@@ -188,7 +188,7 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
   # Set the address of a forward HTTP proxy.
   # This used to accept hashes as arguments but now only accepts
   # arguments of the URI type to prevent leaking credentials.
-  config :proxy, :validate => :uri
+  config :proxy, :validate => :uri # but empty string is allowed
 
   # Set the timeout, in seconds, for network operations and requests sent Elasticsearch. If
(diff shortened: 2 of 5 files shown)


lib/logstash/outputs/elasticsearch.rb
```diff
@@ -188,7 +188,7 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
   # Set the address of a forward HTTP proxy.
   # This used to accept hashes as arguments but now only accepts
   # arguments of the URI type to prevent leaking credentials.
-  config :proxy, :validate => :uri
+  config :proxy, :validate => :uri # but empty string is allowed
```
jsvd: Let's add a note to the `docs/index.asciidoc` entry on the proxy setting in that it's ok to have it empty and what the consequence is (same as no proxy). Even a note on how this is useful when you're using envionment variables with an example would be awesome!

lib/logstash/outputs/elasticsearch.rb
```diff
@@ -238,8 +238,27 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
   # Custom Headers to send on each request to elasticsearch nodes
   config :custom_headers, :validate => :hash, :default => {}
 
+  # @override to handle proxy => '' as if none was set
+  def config_init(params)
+    proxy = params['proxy']
+    if proxy.is_a?(String)
+      # environment variables references aren't yet resolved
+      proxy = deep_replace(proxy)
+      if proxy.eql?('')
```
jsvd: maybe we can support `.empty?` here?

lib/logstash/outputs/elasticsearch.rb
kares: yy, circling back here. left this one unfinished - still appreciate the review.

kares:
should be good to do, :red_circle: ci is existing and will be addressed elsewhere

CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## 10.3.1
+ - Feat: handle proxy => '' as if none was set [#912](https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/912)
```
jsvd: I'd say this is a bug fix instead of feature, as trying to use and empty string as the actual proxy does not make sense. This way it also aligns with the patch level version bump.

PR #922, opened 2020-02-25
Fix integration tests for Elasticsearch 7.6+

robbavey:
Elasticsearch `7.6` introduced an `ilm-history` which caused a number of
tests to fail. Certain ILM tests were failing because they were trying to
delete ILM policies tied to histories, while others were failing due to the
search query being applied across all indices, rather than being targeted.

This commit fixes the search criteria across a number of tests, and avoids
trying to delete any `ilm-history` policies, which is not allowed due to items
existing in the associated history indices.

### DIFF
--- a/spec/es_spec_helper.rb
+++ b/spec/es_spec_helper.rb
@@ -139,7 +139,7 @@ module ESHelper
   end
 
   def clean_ilm(client)
-    client.get_ilm_policy.each_key {|key| client.delete_ilm_policy(name: key)}
+    client.get_ilm_policy.each_key { |key| client.delete_ilm_policy(name: key)  unless key =~ /history-ilm-policy/ }
   end
 
   def supports_ilm?(client)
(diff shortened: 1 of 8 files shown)


robbavey:
Thanks @kares!

spec/es_spec_helper.rb
```diff
@@ -139,7 +139,7 @@ def put_alias(client, the_alias, index)
   end
 
   def clean_ilm(client)
-    client.get_ilm_policy.each_key {|key| client.delete_ilm_policy(name: key)}
+    client.get_ilm_policy.each_key { |key| client.delete_ilm_policy(name: key)  unless key =~ /history-ilm-policy/ }
```
jsvd: @robbavey thanks for sorting out this fix :)

One note, something to keep in mind in the futre, as this is the only place were we whilelist a policy: these clean* would be more resilient if they delete only the indices/aliases/ilm policies that are installed in the tests vs a "*".

For now it's not a big deal as they're only used in the ilm spec, but as ES has more and more content preloaded with it this list will grow.

PR #923, opened 2020-02-26
More fixes to Elasticsearch output build

robbavey:
Later versions of the Elasticsearch mandate a `name` parameter to the
`exists_alias` method. This commit updates the tests to include that
parameter.

Additionally, there are some fixes to move towards an "inclusion" list
of policies to delete, rather than an "exclusion" list.

### DIFF
--- a/spec/es_spec_helper.rb
+++ b/spec/es_spec_helper.rb
@@ -139,7 +139,7 @@ module ESHelper
   end
 
   def clean_ilm(client)
-    client.get_ilm_policy.each_key { |key| client.delete_ilm_policy(name: key)  unless key =~ /history-ilm-policy/ }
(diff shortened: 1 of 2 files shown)


spec/es_spec_helper.rb
```diff
@@ -139,7 +139,7 @@ def put_alias(client, the_alias, index)
   end
 
   def clean_ilm(client)
-    client.get_ilm_policy.each_key { |key| client.delete_ilm_policy(name: key)  unless key =~ /history-ilm-policy/ }
+    client.get_ilm_policy.each_key { |key| client.delete_ilm_policy(name: key)  if key =~ /logstash-policy/ }
```
kares: nice, to have a logstash- prefix convention and only delete what specs create!

robbavey:
Thanks @kares!

PR #927, opened 2020-03-24
[Doc]Fix default value for index and bump to v.10.4.0

### LINKED ISSUE #876: Elasticsearch output plugin index option has invalid example
https://www.elastic.co/guide/en/logstash/current/plugins-outputs-elasticsearch.html#plugins-outputs-elasticsearch-index

The default option listed says "Default value is `"logstash-%{+YYYY.MM.dd}"`. However, that is invalid, the year should be lower case: `"logstash-%{+yyyy.MM.dd}"`.

https://discuss.elastic.co/t/logstash-not-creating-index-daily-based-on-date/189306

karenzone:
Resolves #876 
Default value listed for `index` was listed as `logstash-%{+YYYY.MM.dd}`
This PR corrects the default value to `logstash-%{+yyyy.MM.dd}`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 10.4.0
+ - Fixed default index value [#927](https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/927)
+
 ## 10.3.3
  - [DOC] Replaced link to Elastic Cloud trial with attribute, and fixed a comma splice [#926](https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/926)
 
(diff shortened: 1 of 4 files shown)


kares:
Hey Karen, nice find!
Believe we should also change the [implementation](https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/v10.3.3/lib/logstash/outputs/elasticsearch/common_configs.rb#L6) to use `yyyy` instead of `YYYY` 
and release a new minor version.

`Y` - represents year-of-era
`y` - is 'normal' year we should be using

Happy to do that here if needed.

karenzone:
@kares 

> nice find!
Thanks! The credit goes to the people who created and commented in #876 and in Discuss. I'm just making the fix.  :-)

Thank you for your suggestion. Please take another look, and let me know. Are you set up to test this quickly?

karenzone:
**PUBLISHED**
Screen Shot 2020-03-26 at 9 58 37 AM

PR #930, opened 2020-03-27
Centralize travis testing

jsvd:
same work as done in the https://github.com/logstash-plugins/logstash-filter-elasticsearch/pull/126 and https://github.com/logstash-plugins/logstash-input-elasticsearch/pull/120.

The only difference is that integration tests need to know the exact elasticsearch version being tested (but the input can be "7.x"). This PR [modifies the `.ci/logstash-run`](https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/930/files#diff-1c2a32dd6b80f9ce379979bbdaa3103fR19) to fetch the version from elasticsearch itself.


.ci/logstash-run.sh
```diff
@@ -3,21 +3,20 @@ set -ex
...
+  while ! curl -s $ES_URL >/dev/null && [[ $count -ne 0 ]]; do
     count=$(( $count - 1 ))
-    [[ $count -eq 0 ]] && return 1
+    [[ $count -eq 0 ]] && exit 1
     sleep 1
   done
-  echo "Elasticsearch is Up !"
-
-  return 0
+  echo $(curl -s $ES_URL | python2 -c "import sys, json; print json.load(sys.stdin)['version']['number']")
```
kares:
```suggestion
  echo $(curl -s $ES_URL | python -c "import sys, json; print(json.load(sys.stdin)['version']['number'])")
```
... should be future proof against Python 2/3

PR #934, opened 2020-04-16
add elasticsearch API key support

### LINKED ISSUE #908: Add support authentication with API keys
Elasticsearch has support for authentication with [API keys](https://www.elastic.co/guide/en/elasticsearch/reference/7.5/security-api-create-api-key.html). As a user I would like to be able to configure the elasticsearch output with an API key, as opposed to a username and password, for authenticating.

Logstash would add a header like `Authorization: ApiKey base64(<api_key_id>:<key>)` to HTTP requests to Elasticsearch.

colinsurprenant:
Related to https://github.com/elastic/logstash/issues/11788
Fixes #908

- Add support for [elasticsearch API key](https://www.elastic.co/guide/en/elasticsearch/reference/current/security-api-create-api-key.html) authentication using the `api_key` option
  - SSL must be enabled
  - Only one authentication method must be specified out of (`user`/`password`, `cloud_auth` and `api_key`)
- Related specs
- Manually tested against a cloud cluster

### TODO
- [x] Add docs /cc (@karenzone)
- [x] Bump version

### DIFF
--- a/lib/logstash/outputs/elasticsearch/common.rb
+++ b/lib/logstash/outputs/elasticsearch/common.rb
@@ -20,10 +20,6 @@ module LogStash; module Outputs; class ElasticSearch;
       @stopping = Concurrent::AtomicBoolean.new(false)
       # To support BWC, we check if DLQ exists in core (< 5.4). If it doesn't, we use nil to resort to previous behavior.
       @dlq_writer = dlq_enabled? ? execution_context.dlq_writer : nil
-
-      fill_hosts_from_cloud_id
-      fill_user_password_from_cloud_auth
-      setup_hosts # properly sets @hosts
       build_client
       setup_after_successful_connection
       check_action_validity
@@ -112,6 +108,28 @@ module LogStash; module Outputs; class ElasticSearch;
       [action, params, event]
     end
 
+    def validate_authentication
+      authn_options = 0
+      authn_options += 1 if @cloud_auth
+      authn_options += 1 if (@api_key && @api_key.value)
+      authn_options += 1 if (@user || (@password && @password.value))
+
+      if authn_options > 1
+        raise LogStash::ConfigurationError, 'Multiple authentication options are specified, please only use one of user/password, cloud_auth or api_key'
+      end
+
+      if @api_key && @api_key.value && @ssl   != true
+        raise(LogStash::ConfigurationError, "Using api_key authentication requires SSL/TLS secured communication using the `ssl => true` option")
+      end
+
+      if @cloud_auth
+        @user, @password = parse_user_password_from_cloud_auth(@cloud_auth)
+        # params is the plugin global params hash which will be passed to HttpClientBuilder.build
+        params['user'], params['password'] = @user, @password
+      end
+    end
+    private :validate_authentication
+
     def setup_hosts
       @hosts = Array(@hosts)
       if @hosts.empty?
@@ -133,16 +151,6 @@ module LogStash; module Outputs; class ElasticSearch;
         raise LogStash::ConfigurationError, 'Both cloud_id and hosts specified, please only use one of those.'
       end
       @hosts = parse_host_uri_from_cloud_id(@cloud_id)
-    end
-
-    def fill_user_password_from_cloud_auth
-      return unless @cloud_auth
-
-      if @user || @password
-        raise LogStash::ConfigurationError, 'Both cloud_auth and user/password specified, please only use one.'
-      end
-      @user, @password = parse_user_password_from_cloud_auth(@cloud_auth)
-      params['user'], params['password'] = @user, @password
     end
 
     def parse_host_uri_from_cloud_id(cloud_id)
--- a/lib/logstash/outputs/elasticsearch/http_client_builder.rb
+++ b/lib/logstash/outputs/elasticsearch/http_client_builder.rb
@@ -1,4 +1,5 @@
 require 'cgi'
+require "base64"
 
 module LogStash; module Outputs; class ElasticSearch;
   module HttpClientBuilder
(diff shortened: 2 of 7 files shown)


lib/logstash/outputs/elasticsearch/http_client_builder.rb
```diff
@@ -151,6 +154,18 @@ def self.setup_basic_auth(logger, params)
...
+    def self.setup_api_key(logger, params)
+      api_key = params["api_key"]
+
+      return {} unless (api_key && api_key.value)
+
+      unless params["ssl"] == true
+        raise(LogStash::ConfigurationError, "Using api_key authentication requires SSL/TLS secured communication")
+      end
+
+      { "Authorization" => "ApiKey " + Base64.encode64(api_key.value).chomp }
```
yaauie, line 166: `String#chomp` only chomps of the _last_ trailing newline, and will include any intermediate newlines. Instead we can use the RFC4648-compliant [`Base64#strict_encode64`](https://ruby-doc.org/stdlib-2.5.3/libdoc/base64/rdoc/Base64.html#method-i-strict_encode64):

>   Implementations MUST NOT add line feeds to base-encoded data unless
>   the specification referring to this document explicitly directs base
>   encoders to add line feeds after a specific number of characters.
> -- [RFC4648 3.1: Line Feeds In Encoded Data](https://tools.ietf.org/html/rfc4648)

```suggestion
      { "Authorization" => "ApiKey " + Base64.strict_encode64(api_key.value) }
```

lib/logstash/outputs/elasticsearch/http_client_builder.rb
```diff
@@ -151,6 +154,18 @@ def self.setup_basic_auth(logger, params)
       }
     end
 
+    def self.setup_api_key(logger, params)
+      api_key = params["api_key"]
+
+      return {} unless (api_key && api_key.value)
+
+      unless params["ssl"] == true
```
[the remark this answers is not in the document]
colinsurprenant: Good catch. 
- Note that ES will rejet non-SSL requests using an API key - that, of course, does not solve potentially exposing the plaintext API key.
- I suggest we then rely on the explicit `ssl => true` option for this PR. I think that solving the unreliable SSL auto-detection is out of the scope here and relying on the `ssl => true` option, for now, is totally fine IMO.

lib/logstash/outputs/elasticsearch/http_client_builder.rb
colinsurprenant: TIL 👍

lib/logstash/outputs/elasticsearch/http_client_builder.rb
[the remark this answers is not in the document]
colinsurprenant: I pushed a fix to require explicit `ssl => true`. LMKWYT.

colinsurprenant:
Thanks @karenzone - I just pushed some docs updates, LMKWYT.

lib/logstash/outputs/elasticsearch/common.rb
```diff
@@ -112,6 +111,28 @@ def event_action_tuple(event)
...
+        raise(LogStash::ConfigurationError, "Using api_key authentication requires SSL/TLS secured communication using the `ssl => true` option")
+      end
+
+      if @cloud_auth
+        @user, @password = parse_user_password_from_cloud_auth(@cloud_auth)
+        # params is the plugin global params hash which will be passed to HttpClientBuilder.build
+        params['user'], params['password'] = @user, @password
+      end
+    end
+    private :validate_authentication
```
kares, line 134: Hey Colin, we need to consider a case were the plugin is not `register`-ed properly.
There's a use-case (monitoring) where LS just instantiates and tries to get the client.
Only realized lately (after cloud-id support shipped) and **[LS needed to make changes to get it to work](https://github.com/elastic/logstash/pull/11800/files#diff-dc3cf786acdc338cc714f141bc60c7e4R62-R65)**.
(changeset is ugly, relying even more on internals, but for the time being the only way to get cloud-id working)

We need to think about that as we'll eventually upgrade the plugin internally, ideally we would drop all of these plugin method invocations (if `build_client` would do all of the hosts filling based on *hosts, cloud-auth, api-key*). Alternative is to make `validate_authentication` public API (with a *NOTE: used by LS to build a client without plugin.register*).

colinsurprenant: ouch, this is pretty messy. thanks for the heads-up. 
the problem we have here is with backports on elastic/logstash#11800 - by relying on internals like that the correct version on the plugin has to be installed where these changes have been made.

Instead of directly calling `fill_hosts_from_cloud_id`, `fill_user_password_from_cloud_auth` and `setup_hosts`  maybe we could have moved these into `build_client` and avoid leaking more internals and this PR here would not interfere?

colinsurprenant:
@kares following our conversation about the logstash license checker, I refactored to call the plugin validation methods from within `build_client` which now enables https://github.com/elastic/logstash#11818

LMKWYT

colinsurprenant:
@yaauie let me know if this looks good to you too plus if the requirement on explicit `ssl => true` seems reasonable (and defer solving the unreliable SSL auto-detection in another PR).

colinsurprenant:
v10.5.0 published. Thanks @yaauie @kares @karenzone for the reviews!

AMITKHETAN:
Hi, Its fixed but I still need to pass both API_ID and API_KEY value as "<API_ID>:<API_KEY>

colinsurprenant:
@AMITKHETAN can you please open an issue about this and explain your use-case and how & why you would prefer to pass the API_ID and API_KEY values? Thanks,

Issue #823, opened 2019-01-08
Encountered a retryable error. Will Retry with exponential backoff


13142025891:
My log is very large, a single JSON can reach up to 5M, so I set http_compression => true, but as long as http_compression is turned on, I will report an error, and then repeat the error. After a while, logstash will automatically exit if there is insufficient memory. If I don't set it Http_compression is fine, but the traffic is very large.

- Version: logstash 6.2.3  6.5.4
- Version: logstash-output-elasticsearch 9.0 9.3
- Operating System: windows
- Config File 
       output config
	elasticsearch {
		hosts => ["xxx:9200"] 
		index => "xmllog-%{[@metadata][index]}"
	        document_type => "xmllogdb"
                **http_compression => true**  #I set http_compression => true  so error
	}
       logstash  config
queue.type: persisted
queue.max_bytes: 10gb
queue.page_capacity: 200mb
log.level: info
node.name: logstash4xml
#pipeline.batch.delay: 1000
dead_letter_queue.enable: true
 
**logstash log**
**### logstash-output-elasticsearch 9.0 log**
`[2019-01-08T10:30:31,872][ERROR][logstash.outputs.elasticsearch] Encountered a retryable error. Will Retry with exponential backoff  {:code=>400, :url=>"http://172.18.21.126:9200/_bulk", :body=>"{\"error\":{\"root_cause\":[{\"type\":\"illegal_argument_exception\",\"reason\":\"The bulk request must be terminated by a newline [\\n]\"}],\"type\":\"illegal_argument_exception\",\"reason\":\"The bulk request must be terminated by a newline [\\n]\"},\"status\":400}"}`

**### Update logstash-output-elasticsearch 9.3  log**
`[2019-01-08T16:34:56,213][ERROR][logstash.outputs.elasticsearch] Attempted to send a bulk request to elasticsearch' but Elasticsearch appears to be unreachable or down! {:error_message=>"Elasticsearch Unreachable: [http://172.18.21.126:9200/][Manticore::SocketException] Connection reset by peer: socket write error", :class=>"LogStash::Outputs::ElasticSearch::HttpClient::Pool::HostUnreachableError", :will_retry_in_seconds=>64}
[2019-01-08T16:34:58,052][INFO ][logstash.outputs.elasticsearch] Running health check to see if an Elasticsearch connection is working {:healthcheck_url=>http://172.18.21.126:9200/, :path=>"/"}
[...]

PR #937, opened 2020-04-20
Compression should be done on splitted batches

agix:
Elasticsearch `http.max_content_length` limit (https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-http.html) is applied on the plain content wether it's compressed or not.

In this plugin when compression is enabled, compressed batches are splitted using arbitrary constant `TARGET_BULK_BYTES` of 20mb.

It's possible to find cases where uncompressed data are bigger than `http.max_content_length` (100 mb) but smaller than `TARGET_BULK_BYTES` if compressed resulting on `413 Request Entity Too Large`. (related to https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/823)

(with elasticsearch `http.max_content_length` defined to 1mb):

```
agix@host:/tmp# ls -alh /tmp/aaaa*
-rw-r--r-- 1 agix agix 2.1M Apr 20 18:14 /tmp/aaaa
-rw-r--r-- 1 agix agix 2.1K Apr 20 18:14 /tmp/aaaa.gz
agix@host:/tmp# curl -i 127.0.0.1:9200 --data-binary @/tmp/aaaa -H 'Content-Type: application/json'
HTTP/1.1 413 Request Entity Too Large
content-length: 0

agix@host:/tmp# curl -i 127.0.0.1:9200 --data-binary @/tmp/aaaa.gz -H 'Content-Type: application/json' -H 'Content-Encoding: gzip'
HTTP/1.1 100 Continue

HTTP/1.1 413 Request Entity Too Large
content-length: 0
connection: close
```

Furthermore, as the elasticsearch `http.max_content_length` options exists, `TARGET_BULK_BYTES` should be configurable (default to the elasticsearch default of 100mb) so people can pick their own value.

I think multiple issues/PR are related to this (https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/785, https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/833, https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/786)

### DIFF
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -48,7 +48,7 @@ plugin to version 6.2.5 or higher.
 
 If you plan to use the Kibana web
 interface, use the Elasticsearch output plugin to get your log data into
-Elasticsearch. 
+Elasticsearch.
 
 TIP: You can run Elasticsearch on your own hardware, or use our
 https://www.elastic.co/cloud/elasticsearch-service[hosted {es} Service] on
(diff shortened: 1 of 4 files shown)


kares:
Thanks for the PR, believe similar functionality has been implemented https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/972 (in plugin version 10.8.6)

PR #943, opened 2020-05-18
[Doc] Update http compression in Elasticsearch

Leaf-Lin:
See https://github.com/elastic/elasticsearch/pull/56899, Elasticsearch disables compression if HTTPS is enabled.


andsel:
I've added @karenzone as reviewer to have a quick look

PR #948, opened 2020-06-18
bump rake version

jmlrt:
This commit bump rake version to 12.3.3 to fix severity CVE-2020-8130[0]

[0]: https://github.com/advisories/GHSA-jppv-gw3r-w3q8

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -11,5 +11,5 @@ if Dir.exist?(logstash_path) && use_logstash_source
 end
 
 if RUBY_VERSION == "1.9.3"
-  gem 'rake', '12.2.1'
+  gem 'rake', '12.3.3'
 end


Gemfile
```diff
@@ -11,5 +11,5 @@ if Dir.exist?(logstash_path) && use_logstash_source
 end
 
 if RUBY_VERSION == "1.9.3"
-  gem 'rake', '12.2.1'
-end
\ No newline at end of file
+  gem 'rake', '12.3.3'
```
robbavey, line 14: This change cannot be made as rake `12.3.0` and above requires ruby version > 2.0, which Logstash `6.x` and above requires.

We might want to think about just removing this block entirely (and in other plugins) - this is here to retain `5.x` compatibility, but we haven't tested this plugin (or many others) against `5.x` for some time.

Thoughts @jsvd ?

PR #952, opened 2020-07-13
Implement ECS Compatibility Mode

yaauie:
Implements ECS Compatibility Mode for ECS v1.

 - The templates are imported from elastic/ecs, and targeted to indices matching `ecs-logstash-*`.
 - Default index and rollover alias names are prepended with `ecs-` when ECS is enabled.


yaauie:
Remaining test failure is caused by a spec relying on a `Plugin#register` hook before the plugin is sent `Plugin#register`. I'm moving the hook to be a `Plugin#initialize` hook instead, which should eliminate the issue.

lib/logstash/outputs/elasticsearch/template_manager.rb
```diff
@@ -19,9 +19,11 @@ def self.install_template(plugin)
     end
 
     private
-    def self.get_template(path, es_major_version)
-      template_path = path || default_template_path(es_major_version)
+    def self.load_default_template(es_major_version, ecs_compatibility)
+      template_path = default_template_path(es_major_version, ecs_compatibility)
       read_template_file(template_path)
+    rescue => e
+      fail "Failed to load default template for Elasticsearch v#{es_major_version} with ECS #{ecs_compatibility}; caused by: #{e.message}"
```
kares, line 26: nitpick: not sure what kind of (io) exceptions might go wrong here, we might want `e.inspect` just in case?

lib/logstash/outputs/elasticsearch/template_manager.rb
```diff
@@ -46,9 +48,9 @@ def self.template_name(plugin)
       plugin.ilm_in_use? && !plugin.original_params.key?('template_name') ? plugin.ilm_rollover_alias : plugin.template_name
     end
 
-    def self.default_template_path(es_major_version)
+    def self.default_template_path(es_major_version, ecs_compatibility=:disabled)
       template_version = es_major_version == 1 ? 2 : es_major_version
-      default_template_name = "elasticsearch-template-es#{template_version}x.json"
+      default_template_name = "templates/ecs-#{ecs_compatibility}/elasticsearch-#{template_version}x.json"
```
kares, line 53: :+1: naming convention, me being a user - probably "want this ecs thing enabled"

docs/index.asciidoc
```diff
@@ -500,7 +531,9 @@ NOTE: If this setting is specified, the policy must already exist in Elasticsear
 ===== `ilm_rollover_alias`
 
   * Value type is <<string,string>>
-  * Default value is `logstash`
+  * Default value depends on whether <<plugins-{type}s-{plugin}-ecs_compatibility>> is enabled:
+  ** ECS Compatibility disabled: `logstash`
+  ** ECS Compatibility enabled: `ecs-logstash` (ECS Compatibility is enabled)
```
yaauie, line 536:
```suggestion
  ** ECS Compatibility enabled: `ecs-logstash`
```

docs/index.asciidoc
```diff
@@ -58,6 +58,17 @@ https://www.elastic.co/cloud/elasticsearch-service[hosted {es} Service] on
 Elastic Cloud. The Elasticsearch Service is available on AWS, Google Cloud
 Platform, and Microsoft Azure. {ess-trial}[Try the {es} Service for free].
 
+==== Compatibility with the Elastic Common Schema (ECS)
+
+This plugin will persist events to Elasticsearch in the shape produced by
+your pipeline, and _cannot_ be used to re-shape the event structure into a
+shape that complies with ECS.
+
```
karenzone, line 66: What would a user need to do if they need to reshape?  Or maybe your next sentence is the answer. :-)

If we can offer additional approaches or guidance on this topic, we can add more info to the LS Ref and link to it.

docs/index.asciidoc
```diff
@@ -58,6 +58,17 @@ https://www.elastic.co/cloud/elasticsearch-service[hosted {es} Service] on
...
+However, the Elasticsearch Index Templates it manages can be configured to
+be ECS-compatible by setting <<plugins-{type}s-{plugin}-ecs_compatibility>>.
+By having an ECS-compatible template in-place, we can ensure that Elasticsearch
```
karenzone, line 69:
```suggestion
By having an ECS-compatible template in place, we can ensure that Elasticsearch
```

docs/index.asciidoc
```diff
@@ -393,6 +405,25 @@ If you don't set a value for this option:
...
+
+* Value type is <<string,string>>
+* Supported values are:
+** `disabled`: does not provide ECS-compatible templates
+** `v1`: provides defaults that are compatible with v1 of the Elastic Common Schema
+* Default value depends on which version of Logstash is running:
+** When Logstash provides a `pipeline.ecs_compatibility` setting, its value is used as the default
+** Otherwise, the default value is `disabled`.
+
+Controls this plugin's desired compatibility with the https://www.elastic.co/blog/introducing-the-elastic-common-schema[Elastic Common Schema],
```
karenzone, line 419:
```suggestion
Controls this plugin's compatibility with the https://www.elastic.co/blog/introducing-the-elastic-common-schema[Elastic Common Schema],
```

docs/index.asciidoc
```diff
@@ -393,6 +405,25 @@ If you don't set a value for this option:
...
+including the installation of ECS-Compatible index templates.
```
karenzone, line 420:
```suggestion
including the installation of ECS-compatible index templates.
```

docs/index.asciidoc
karenzone: If you'd like to link to the ECS documentation (in addition or instead of), here's the formatting you need:
`{ecs-ref}[Elastic Common Schema (ECS)]`

docs/index.asciidoc
yaauie: A user would need to write their pipeline in such a way that they use field structure as semantically defined by the ECS schema, which is beyond the scope of the ECS Compatibility Mode of this output plugin.

An event in the Elastic Common Schema relies on the semantic meaning that ECS applies to field _values_ based on the field _names_, and as much as we'd like to magically shove everything into the schema, by the time it gets to this output we typically don't have enough information to do so (e.g., a string that looks like an ip address could be any of `client.ip`, `client.nat.ip`, `destination.ip`, `destination.nat.ip`, `host.ip`, `observer.ip`, `related.ip`, `server.ip`, `server.nat.ip`, `source.ip`, or `source.nat.ip`), and it is only in the _pipeline_ when the user who has _context_ into the data they are parsing, that mapping to the correct ECS-defined field can be done.

docs/index.asciidoc
yaauie: heyyy. perfect.

docs/index.asciidoc
yaauie: I _think_ I have added some clarity in both this paragraph and the one that follows.

karenzone: Yes! Nice clarification. Thank you.

PR #953, opened 2020-07-15
Fix breakage caused by external use of our internal API

yaauie:
Because Logstash Core v7.7.0-v7.8.x override the `use_event_type?` in a
monitoring implelentation's subclass with a required (and ignored) `client`
argument, when we send `use_event_type?` internally, we must send an argument
or else we break compatibility with those versions of Logstash.
    
This change re-adds the required argument and modifies our call-site to send
with an arity of 1, re-names the resulting local variable to be descriptive
of our intent, and adds commentary to prevent future accidental breakage.

### DIFF
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -419,7 +419,7 @@ If you don't set a value for this option:
 ** When Logstash provides a `pipeline.ecs_compatibility` setting, its value is used as the default
 ** Otherwise, the default value is `disabled`.
 
-Controls this plugin's compatibility with the {ecs-ref}[Elastic Common Schema (ECS)],
+Controls this plugin's compatibility with the https://www.elastic.co/guide/en/ecs/current/index.html[Elastic Common Schema (ECS)],
 including the installation of ECS-compatible index templates.
 The value of this setting affects the _default_ values of:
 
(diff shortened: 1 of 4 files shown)


yaauie:
Not sure how or why, but I've pushed changelog and version bump to the branch referenced here and Github is not picking it up -> https://github.com/logstash-plugins/logstash-output-elasticsearch/compare/master...yaauie:fix-external-use-of-internal-api

jsvd:
GitHub is acting up, already happened in another PR

yaauie:
Ah, it seems caught up now.

lib/logstash/outputs/elasticsearch/common.rb
```diff
@@ -60,7 +60,14 @@ def successful_connection?
       !!maximum_seen_major_version
     end
 
-    def use_event_type?
+    ##
+    # WARNING: This method is overridden in a subclass in Logstash Core 7.7-7.8's monitoring,
```
jsvd, line 64: :heart:

yaauie:
Needs a doc fix too before publishing.

docs/index.asciidoc
```diff
@@ -419,7 +419,7 @@ If you don't set a value for this option:
 ** When Logstash provides a `pipeline.ecs_compatibility` setting, its value is used as the default
 ** Otherwise, the default value is `disabled`.
 
-Controls this plugin's compatibility with the {ecs-ref}}[Elastic Common Schema (ECS)],
+Controls this plugin's compatibility with the {ecs-ref}[Elastic Common Schema (ECS)],
```
karenzone, line 422:
```suggestion
Controls this plugin's compatibility with the https://www.elastic.co/guide/en/ecs/current/index.html[Elastic Common Schema (ECS)],
```

Issue #954, opened 2020-07-23
Healthcheck fails in weird ways when response is an HTTP 404


yaauie:
`ManticoreAdapter#perform_request` explicitly counts an HTTP 404 as a non-exceptional condition, which allows the healthcheck to get into a weird state when reviving a connection (which can occur during initialization), specifically when we send `Pool#get_es_version(url)` and expect the response to be a specific shape.

We need to validate that we received a 2XX level response code from the initial healthcheck ping, ideally one that validates the shape of the repsponse to be Elasticsearch-y, and we need to make the whole code more resilient to non-success paths in `Pool#get_es_version(url)`.

TheVastyDeep:
The get_es_version function in the pool.rb certainly makes assumptions about the shape of the response (i.e. that it is valid JSON and the version field contains a number sub-field. If those assumptions are violated the result is an exception that prevents the pipeline starting but which gives absolutely no indication of what the problem is until you enable DEBUG logging.

PR #955, opened 2020-07-23
Reconnection partial failure modes

yaauie:
- Ensures that a url's healthcheck is a _success_ before marking it alive in the pool
 - Ensures that the version of the connected Elasticsearch is retrieved _successfully_ before allowing the connection to be marked as alive
 - Eliminates a crash that can occur when HTTP connection can be established and the connected resource isn't Elasticsearch (e.g., Elastic Cloud returning an HTTP `404 Not Found`)

Resolves: #954


lib/logstash/outputs/elasticsearch/http_client/pool.rb
```diff
@@ -481,8 +486,10 @@ def return_connection(url)
     end
 
     def get_es_version(url)
-      request = perform_request_to_url(url, :get, ROOT_URI_PATH)
-      LogStash::Json.load(request.body)["version"]["number"]
+      response = perform_request_to_url(url, :get, ROOT_URI_PATH)
+      return nil unless (200..299).cover?(response.code)
+
+      LogStash::Json.load(response.body).dig("version","number")
```
kares, line 492: :thinking: theoretically the plugin does not limit itself being installed on LS 5.6 (< 6.0) ...
should we do so, since we're explicitly using a Ruby 2.3 (`Hash#dig`) feature here?

kares (approved):
:+1: semi-tested with ES 7.8 (stubbed out a 404).
Normal ES output working as before, bonus points for sanitized URLs!

lib/logstash/outputs/elasticsearch/http_client/pool.rb
yaauie: Good catch. I would love to start using modern ruby features, but the benefit here isn't worth a major version bump of the plugin.

jsvd:
replaced by #1130

PR #959, opened 2020-07-31
avoid retry loop caused by sending empty bulk requests to Elasticsearch

yaauie:
Under Logstash 7.x's Java Execution Engine, empty batches can be sent to this
output's `multi_receive`, and because Elasticsearch emits an HTTP 400-level
error when it receives an empty bulk request, this plugin ends up in an
unescapable retry loop when it receives an empty batch.


yaauie:
It looks like safeguards against empty payloads already exist at a lower layer, and that the error I'm seeing that _matches_ that of an empty payload can also be triggered by a _corrupt_ payload #960 

I'm closing this unmerged.

PR #962, opened 2020-08-13
Don't set the pipeline param if the sprintf returns an empty value

### LINKED ISSUE #961: pipeline value is sent to elasticsearch even when the value is an empty string
We have configured our logstash pipeline to do this at the end of the filter section:

```
   if ! [@metadata][pipeline] {
     mutate {
           add_field => {"[@metadata][pipeline]" => ""}
       }
     }
```

And then in the elasticsearch output: 

```
   if [type] == "beats" {
     elasticsearch {
...
       pipeline => "%{[@metadata][pipeline]}"
...
     }
   }
```

This results in the following error thrown by the eleasticsearch output:

```
{
         "input" => {
        "type" => "log"
    },
         "agent" => {
            "hostname" => "a0fe929361c6",
                "name" => "flaper87-filebeat-test",
                  "id" => "1d58ef10-b92e-4b12-9c9a-8152ec78901d",
        "ephemeral_id" => "b9b71472-8e7d-4bf4-bb31-ac8eab60081f",
                "type" => "filebeat",
             "version" => "7.6.2"
    },
    "@timestamp" => 2020-08-13T06:28:55.998Z,
           "ecs" => {
        "version" => "1.4.0"
    },
           "log" => {
          "file" => {
            "path" => "/filebeat/logs/my-log-file.log"
        },
        "offset" => 64
    },
     "@metadata" => {
           "version" => "7.6.2",
          "pipeline" => "",
              "beat" => "filebeat",
              "type" => "_doc",
        "ip_address" => "10.128.15.207"
    },
      "@version" => "1",
          "host" => {
        "name" => "flaper87-filebeat-test"
    },
          "type" => "beats",
       "message" => "testing",
        "fields" => {
        "cluster" => "non-k8s"
    },
          "tags" => [
        [0] "local"
    ]
}
[2020-08-13T06:28:58,863][ERROR][logstash.outputs.elasticsearch][elasticsearch] Encountered a retryable error. Will Retry with exponential backoff  {:code=>400, :url=>"https://ccbf29977a6ce7ac1bb01e41c9ff9316.us-west-2.aws.found.io:9243/_bulk"}
[2020-08-13T06:29:00,985][ERROR][logstash.outputs.elasticsearch][elasticsearch] Encountered a retryable error. Will Retry with exponential backoff  {:code=>400, :url=>"https://ccbf29977a6ce7ac1bb01e41c9ff9316.us-west-2.aws.found.io:9243/_bulk"}
[2020-08-13T06:29:05,051][ERROR][logstash.outputs.elasticsearch][elasticsearch] Encountered a retryable error. Will Retry with exponential backoff  {:code=>400, :url=>"https://ccbf29977a6ce7ac1bb01e41c9ff9316.us-west-2.aws.found.io:9243/_bulk"}
```

I would have expected the `elasticsearch-output` to not send the `pipeline` "value" as part of the request, which I am assuming is what it is doing.

flaper87:
closes #961

### DIFF
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -647,8 +647,9 @@ not also set this field. That will raise an error at startup
   * Value type is <<string,string>>
   * Default value is `nil`
 
-Set which ingest pipeline you wish to execute for an event. You can also use event dependent configuration here
-like `pipeline => "%{[@metadata][pipeline]}"`. The pipeline parameter won't be set if the value resolves to empty string ("").
+Set which ingest pipeline you wish to execute for an event. You can also use event dependent configuration
+here like `pipeline => "%{INGEST_PIPELINE}"`. The pipeline parameter won't be set if the option is set to 
+an empty string ("") or `nil`.
 
 [id="plugins-{type}s-{plugin}-pool_max"]
 ===== `pool_max` 


lib/logstash/outputs/elasticsearch/common.rb
```diff
@@ -84,7 +84,10 @@ def event_action_tuple(event)
       params[:_type] = get_event_type(event) if use_event_type?(nil)
 
       if @pipeline
-        params[:pipeline] = event.sprintf(@pipeline)
+        value = event.sprintf(@pipeline)
+        if value
+          params[:pipeline] = value
+        end
```
jsvd, line 90: The expectation is that an empty string results in the parameter not being included. However, in Ruby only `false` and `nil` (the objects, not the strings) are falsey. So an expression with an empty string evalues to true:

```ruby
> value = ""
=> ""
> puts "hi" if value
hi
```

Since `sprintf` always returns a string, we can change this expression to check if it is empty:

```suggestion
        # convention: empty string equates to not using a pipeline
        # this is useful when using a field reference in the pipeline setting, e.g.
        #          pipeline => "%{[metadata][pipeline]}"
        params[:pipeline] = value unless value.empty?
```

flaper87: ha, you can tell I'm used to python since I thought `""` would evaluate to false. thanks for the suggestion.

flaper87:
This was cherry-picked and pushed to the repo: https://github.com/logstash-plugins/logstash-output-elasticsearch/commit/8804933e2b15d867a049319e52ee61223552b6be

jsvd:
Thanks for the PR @flaper87, this has been released as [v10.7.0](https://rubygems.org/gems/logstash-output-elasticsearch/versions/10.7.0-java)

PR #963, opened 2020-08-18
Update common_configs.rb index date format

bukovyn:


robbavey:
Hi @bukovyn Would you mind signing the CLA? We need this to be signed before we can merge your commit.

Thanks!

Issue #958, opened 2020-07-31
Support for new index templates


consulthys:
In Elasticsearch 7.8, index templates have [been refactored](https://www.elastic.co/guide/en/elasticsearch/reference/current/index-templates.html) and this plugin does not yet support them, only the [legacy templates](https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-templates-v1.html).

When trying to send a refactored template, the error message is

[...]

PR #968, opened 2020-09-13
Index Template Support Update

rkbennett:
#958 
Adds support for new index template schema
Adds handling for adding ilm configurations when settings field doesn't exist
Only overwrites index_patterns from template when absolutely necessary
Adds legacy_template option (boolean), which will change paths between _index_template and _template, and also configure ilm fields appropriately


alah64:
Looks like it hasn't been fixed yet

PR #972, opened 2020-10-06
fix: prevent sub-batch 413's from blocking whole batch

yaauie:
The Http Client breaks batches of actions into sub-batches that are up to 20MB
in size, sending larger actions as batches-of-one, and zips the responses
together to emulate a single batch response from the Elasticsearch API.

When an individual HTTP request is rejected by Elasticsearch (or by an
intermediate proxy or load-balancer) with an HTTP/1.1 413, we can emulate
the error response instead of blowing an exception through to the whole batch.

### DIFF
--- a/.ci/elasticsearch-run.sh
+++ b/.ci/elasticsearch-run.sh
@@ -1,4 +1,4 @@
 #!/bin/bash
 set -ex
 
-/usr/share/elasticsearch/bin/elasticsearch -Ediscovery.type=single-node
+/usr/share/elasticsearch/bin/elasticsearch -Ediscovery.type=single-node -Eaction.destructive_requires_name=false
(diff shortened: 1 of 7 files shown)


andsel (commented):
I was wondering if it's feasible to add a unit test that verify the new behavior introduced, overwriting `TARGET_BULK_BYTES` doubling the `LogStash::Outputs::ElasticSearch::HttpClient::Pool` client to return a 413 on the `post` request.

andsel (commented):
I've noticed only a minor repetition in the Changelog.
The changes to me seems OK, but I don't understand why the it fails on [Travis](https://travis-ci.org/github/logstash-plugins/logstash-output-elasticsearch/jobs/741975814#L5344)

andsel (approved):
LGTM

yaauie:
Reviving, and expanding the scope to address a situation where we could send a <20MB payload that expanded beyond Elasticsearch's 100MB buffers (#823). Bulk grouping is now entirely determined by decompressed size.

yaauie:
Failing tests on 8.0-SNAPSHOT are unrelated -- it looks like integration tests do some cleanup of ILM policies without first cleaning up the indices governed by those policies, which will be an error in 8.x. I'll see if fixing is trivial, or will file a separate issue to chase that down outside this PR.

andsel (approved):
LGTM

yaauie:
Theoretically have a fix in place for the failing integration tests against ES 8 snapshot -- They recently made a change to the default value of `action.destructive_requires_name` to `true` in https://github.com/elastic/elasticsearch/pull/66908 , which causes our attempt to wildcard-delete the indices prior to removal of the ILM policies fail.

Once CI goes green, I plan to merge.

PR #973, opened 2020-10-20
separate config options into elasticsearch specific and shared

colinsurprenant:
This is work toward Data Streams integration https://github.com/elastic/logstash/issues/12178

Specific elasticsearch option are now into the main elasticsearch class file and shared option have been moved into the PluginMixins::ElasticSearch::APIConfigs namespace. 

This is essentially a no-op PR with only the config options reorganization. 

I included the removal of both `DEFAULT_INDEX_NAME` and `DEFAULT_ROLLOVER_ALIAS` constants which were unused.


lib/logstash/outputs/elasticsearch/shared_configs.rb
```diff
@@ -0,0 +1,160 @@
+require 'forwardable' # Needed for logstash core SafeURI. We need to patch this in core: https://github.com/elastic/logstash/pull/5978
+
+module LogStash; module Outputs; class ElasticSearch
+  module SharedConfigs
```
yaauie, line 4: yes, these are shared, but _why_ are they shared?

I think `ConnectionConfigs` may be a better name because almost all of them seem to have to do with establishing and maintaining connections to Elasticsearch. There are a couple around retries that don't seem to quite fit there, so maybe `APIConfigs`?

lib/logstash/outputs/elasticsearch/elasticsearch_configs.rb
```diff
@@ -1,15 +1,23 @@
-require 'forwardable' # Needed for logstash core SafeURI. We need to patch this in core: https://github.com/elastic/logstash/pull/5978
-
 module LogStash; module Outputs; class ElasticSearch
-  module CommonConfigs
+  module ElasticsearchConfigs
```
yaauie, line 2: If all of these are only ever used in `LogStash::Outputs::ElasticSearch`, is there value in having them extracted to `LogStash::Outputs::ElasticSearch::ElasticsearchConfigs`? Should we be doubling-down on removing them from directly being on the output plugin's class, or should the ones that only apply to the output be defined on the output?

lib/logstash/outputs/elasticsearch/shared_configs.rb
colinsurprenant: Agreed naming is a bit hard here and SharedConfigs is certainly not the best. +1 APIConfigs.

lib/logstash/outputs/elasticsearch/elasticsearch_configs.rb
colinsurprenant: Good point. So my thinking was that we already had that `CommonConfigs` module which defined *some* options while others were defined in the plugin class and I did not really see any clear indications on the options separation choices, there is certainly some history here. My goad is to extract what will be shared with the data streams output which will depend on this output. I decided to simply put all options which were specific to this output into the ElasticsearchConfigs module and all other which I think will be shared with data streams into the SharedConfigs module. 

I am +1 on having these options back into the plugin class. There is no particular need to have these into a module.

colinsurprenant:
@yaauie Agree with your comments. LMKWYT with the modifications.

colinsurprenant:
Bump /cc @jsvd

yaauie (approved):
LGTM 👍🏼

lib/logstash/outputs/elasticsearch/api_configs.rb
```diff
@@ -0,0 +1,160 @@
+require 'forwardable' # Needed for logstash core SafeURI. We need to patch this in core: https://github.com/elastic/logstash/pull/5978
```
jsvd, line 1: this has been  fixed in logstash since 5.1/6.0.0, so we can remove it
```suggestion
```

lib/logstash/outputs/elasticsearch.rb
```diff
@@ -122,130 +123,132 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
...
+  # you wanted to be able to update it regularly, this option could help there as well.
+  #
+  # Please note that if you are using your own customized version of the Logstash
+  # template (logstash), setting this to true will make Logstash to overwrite
+  # the "logstash" template (i.e. removing all customized settings)
+  config :template_overwrite, :validate => :boolean, :default => false
+
+  # The document ID for the index. Useful for overwriting existing entries in
+  # Elasticsearch with the same ID.
+  config :document_id, :validate => :string
```
jsvd, line 180: @colinsurprenant even with data streams, create actions can take an id to be used as document id, so this belongs in the shared config settings.

lib/logstash/outputs/elasticsearch.rb
```diff
@@ -106,6 +102,11 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
   # ecs_compatibility option, provided by Logstash core or the support adapter.
   include(LogStash::PluginMixins::ECSCompatibilitySupport)
 
+  # Generic/API config options that the data streams output will share
```
jsvd, line 105:
```suggestion
  # Generic/API config options that any document indexer output needs
  # These are to be shared between this generic elasticsearch output and datastream output.
```

jsvd (commented):
Added a couple of notes, otherwise LGTM

For reference, here's the list of settings that are shared:

<details>
  <summary>❯ grep "config :" lib/logstash/outputs/elasticsearch/api_configs.rb | sort | cut -d " " -f 8-</summary>

```  
:api_key, :validate => :password
:bulk_path, :validate => :string
:cacert, :validate => :path
:cloud_auth, :validate => :password
:cloud_id, :validate => :string
:custom_headers, :validate => :hash, :default => {}
:failure_type_logging_whitelist, :validate => :array, :default => []
:healthcheck_path, :validate => :string
:hosts, :validate => :uri, :default => [ DEFAULT_HOST ], :list => true
:http_compression, :validate => :boolean, :default => false
:keystore, :validate => :path
:keystore_password, :validate => :password
:parameters, :validate => :hash
:password, :validate => :password
:path, :validate => :string
:pool_max, :validate => :number, :default => 1000
:pool_max_per_route, :validate => :number, :default => 100
:proxy, :validate => :uri # but empty string is allowed
:resurrect_delay, :validate => :number, :default => 5
:retry_initial_interval, :validate => :number, :default => 2
:retry_max_interval, :validate => :number, :default => 64
:sniffing, :validate => :boolean, :default => false
:sniffing_delay, :validate => :number, :default => 5
:sniffing_path, :validate => :string
:ssl, :validate => :boolean
:ssl_certificate_verification, :validate => :boolean, :default => true
:timeout, :validate => :number, :default => 60
:truststore, :validate => :path
:truststore_password, :validate => :password
:user, :validate => :string
:validate_after_inactivity, :validate => :number, :default => 10000
```
</details>

And the list of configs we'll not use in the datastreams output:

<details>
  <summary>❯ grep "config :" lib/logstash/outputs/elasticsearch.rb | sort | cut -d " " -f 4-</summary>

``` 
:action, :validate => :string, :default => "index"
:doc_as_upsert, :validate => :boolean, :default => false
:document_id, :validate => :string
:document_type,
:ilm_enabled, :validate => [true, false, 'true', 'false', 'auto'], :default => 'auto'
:ilm_pattern, :validate => :string, :default => '{now/d}-000001'
:ilm_policy, :validate => :string, :default => DEFAULT_POLICY
:ilm_rollover_alias, :validate => :string
:index, :validate => :string
:join_field, :validate => :string, :default => nil
:manage_template, :validate => :boolean, :default => true
:parent, :validate => :string, :default => nil
:pipeline, :validate => :string, :default => nil
:retry_on_conflict, :validate => :number, :default => 1
:routing, :validate => :string
:script, :validate => :string, :default => ""
:script_lang, :validate => :string, :default => "painless"
:script_type, :validate => ["inline", 'indexed', "file"], :default => ["inline"]
:script_var_name, :validate => :string, :default => "event"
:scripted_upsert, :validate => :boolean, :default => false
:template, :validate => :path
:template_name, :validate => :string
:template_overwrite, :validate => :boolean, :default => false
:upsert, :validate => :string, :default => ""
:version, :validate => :string
:version_type, :validate => ["internal", 'external', "external_gt", "external_gte", "force"]
```

</details>

colinsurprenant:
@yaauie per discussion in #976 I moved the `APIConfig` into the `PluginMixins` namespace. Could you please double check that last commit to make sure all in good per our discussion. When good, I'll rebase #976 against this and move the commons code in the mixin.

colinsurprenant:
Thanks @yaauie @jsvd for the reviews. This is merged and will be part of 10.8.0.

PR #977, opened 2020-11-06
ci: look for arch-specific snapshots when necessary

yaauie:
Since the 8.x docker images for Logstash are arch-specific (relying on the base install with embedded arch-specific JVM), they no longer will have the same file names we relied on for 7.x causing builds to fail when looking up the URL for the most recent snapshot.

This PR refactors the code that finds the URL for docker image snapshots to find appropriate artifacts, whether the listing includes no-arch or `x86_64`-arch artifacts. It also adds helpful logging when none of the items can be found, and breaks the code into a function for reuse when loading the elasticsearch image.

Prior to this change, failure to load a URL resulted in a blind error:

~~~
null
curl: (6) Could not resolve host: null
~~~

After this change, an error looking up the URL will be much more informative:
~~~
Downloading snapshot docker artifact: logstash-8.0.0-SNAPSHOT-docker-image
Failed to find 'logstash-8.0.0-SNAPSHOT-docker-image.tar.gz'
Failed to find 'logstash-8.0.0-SNAPSHOT-docker-image-BADARCHINTENTIONALFAILURE.tar.gz'
Listing:
[
  "logstash-8.0.0-SNAPSHOT-docker-image-x86_64.tar.gz",
  "logstash-oss-8.0.0-SNAPSHOT-docker-image-x86_64.tar.gz",
  "logstash-ubi8-8.0.0-SNAPSHOT-docker-image-x86_64.tar.gz"
]
[error: 1] 
~~~

### DIFF
--- a/.ci/docker-setup.sh
+++ b/.ci/docker-setup.sh
@@ -6,6 +6,42 @@ set -e
 
 VERSION_URL="https://raw.githubusercontent.com/elastic/logstash/master/ci/logstash_releases.json"
 
+
+download_and_load_docker_snapshot_artifact() {
+  project="${1?project name required}"
+
+  artifact_type="docker-image"
+  artifact_name_base="${project}${DISTRIBUTION_SUFFIX}-${ELASTIC_STACK_VERSION}-${artifact_type}"
+  echo "Downloading snapshot docker image: ${project}${DISTRIBUTION_SUFFIX} (${ELASTIC_STACK_VERSION})"
+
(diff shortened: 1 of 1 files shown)


yaauie:
Although this plugin defines its own `.ci/docker-setup.sh`, it is heavily based on the one defined in `.ci` project. A matching PR for this exists there: https://github.com/logstash-plugins/.ci/pull/19

robbavey (approved):
Tests are green - LGTM

Thanks @yaauie!

PR #984, opened 2020-12-02
fix bug in choosing the NoopLicenseChecker

colinsurprenant:
Hash#fetch returns the default value when the hash does not contain a value for the key but here we set that value to `nil`. This is a safer construct that will work for `nil` value or no-value.


yaauie (approved):
LGTM :+1:

colinsurprenant:
v10.8.1 published.

PR #988, opened 2021-01-18
Feat: data-stream support + logging improvements

kares:
### Data stream support, specification https://github.com/elastic/logstash/issues/12178

- [x] confirm (spec-ified) OSS behavior now defunct
- [x] needs documentation
- [x] name validation could do more checks but should be sufficient for now
- [ ] NOTE: due missing support on ES end we currently do not validate whether a data-stream endpoint is actually a DS
  pending issue: https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/1001
- [x] ~~use at least `require_alias=true` with `_bulk` when using DS?!?~~

DS configuration diverged from the proposed implementation plan:
> Implementation wise, the proposal is to build a proxy class that takes in all settings and decides if they are DS compatible.

Simply because it seemed fine to just detect the configuration directly, given:
- we can check for configuration params from `LogStash::PluginMixins::ElasticSearch::APIConfigs` mixin
  there's a `APIConfigs::CONFIG_PARAMS` convention to make sure introducing a new param will not break DS detection
- we can filter out all super-class (generic) configuration parameters
- there's a `data_stream_` prefix convention for DS specific configuration params

---

Initially I extracted a base-line for the later `include DataStreamSupport` but I've run into (existing) spec problems due mocking. I decided to fix these along the way and also bring back logging which was disabled previously due tests.
Splitting this work into separate PR would require extra work as spec improvements rely on some of the DS related refactorinfs for the updated mocking, thus its all here with minor changes besides DS support, namely:

- re-invent error logging (disabled previously) `happening after successful_connection` (from install template etc.)
- unified logging syntax - use Capitalized messages without `.` or `!` logical ends 
(might be just me but I never appreciate logs 'screaming' at me with `!`)
- split some methods and hide what is meant to be `private` (with respect to LS re-use in monitoring)

### DIFF
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -21,6 +21,10 @@ include::{include_path}/plugin_header.asciidoc[]
 
 ==== Description
 
+Elasticsearch provides near real-time search and analytics for all types of
+data. The Elasticsearch output plugin can store both time series datasets (such
+as logs, events, and metrics) and non-time series data in Elasticsearch.
+
 If you plan to use the Kibana web interface to analyze data transformed by
 Logstash, use the Elasticsearch output plugin to get your data into
 Elasticsearch.
@@ -354,8 +358,6 @@ The Elasticsearch action to perform. Valid actions are:
 - A sprintf style string to change the action based on the content of the event. The value `%{[foo]}`
   would use the foo field for the action
 
-When using data-streams a `create` default is assumed, otherwise the default action is `"index"`.
-
 For more details on actions, check out the {ref}/docs-bulk.html[Elasticsearch bulk API documentation].
 
 [id="plugins-{type}s-{plugin}-api_key"]
@@ -418,7 +420,6 @@ For more details, check out the
 
 Defines whether data will be indexed into an Elasticsearch data stream.
 The other `data_stream_*` settings will be used only if this setting is enabled.
-Defaults to `false` in Logstash 7.x and `auto` starting in Logstash 8.0.
 
 Logstash handles the output as a data stream when the supplied configuration
 is compatible with data streams and this value is set to `auto`.
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -432,7 +432,7 @@ with data streams Logstash will assume to be handling events as if `true` was se
 * Default value is `true`.
 
 Automatically routes events by deriving the data stream name using specific event
-fields with the `%{data_stream.type}-%{data_stream.dataset}-%{data_stream.namespace}` format.
+fields with the `%{[data_stream][type]}-%{[data_stream][dataset]}-%{[data_stream][namespace]}` format.
 
 If enabled, the `data_stream.*` event fields will take precedence over the
 `data_stream_type`, `data_stream_dataset`, and `data_stream_namespace` settings,
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -423,7 +423,6 @@ The other `data_stream_*` settings will be used only if this setting is enabled.
 
 Logstash handles the output as a data stream when the supplied configuration
 is compatible with data streams and this value is set to `auto`.
-with data streams Logstash will assume to be handling events as if `true` was set.
 
 [id="plugins-{type}s-{plugin}-data_stream_auto_routing"]
 ===== `data_stream_auto_routing`
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -341,7 +341,7 @@ output plugins.
 ===== `action`
 
   * Value type is <<string,string>>
-  * Default value depends on whether data-streams are used.
+  * Default value is `create` for data streams, and `index` for non-time series data. 
 
 The Elasticsearch action to perform. Valid actions are:
 
@@ -414,13 +414,14 @@ For more details, check out the
 ===== `data_stream`
 
 * Value can be any of: `true`, `false` and `auto`
-* Default value depends on Logstash version.
+* Default is `false` in Logstash 7.x and `auto` starting in Logstash 8.0.
 
 Defines whether data will be indexed into an Elasticsearch data stream.
-The other `data_stream_*` settings will only be used if this setting is enabled.
+The other `data_stream_*` settings will be used only if this setting is enabled.
 Defaults to `false` in Logstash 7.x and `auto` starting in Logstash 8.0.
 
-The `auto` behavior means that whenever the supplied configuration is compatible
+Logstash handles the output as a data stream when the supplied configuration
+is compatible with data streams and this value is set to `auto`.
 with data streams Logstash will assume to be handling events as if `true` was set.
 
 [id="plugins-{type}s-{plugin}-data_stream_auto_routing"]
@@ -458,11 +459,12 @@ The data stream namespace used to construct the data stream at index time.
 * Value type is <<boolean,boolean>>
 * Default value is `true`
 
-Automatically syncs the `data_stream.*` event fields if they are missing from the
-event, this ensures fields match the data stream name that events are being sent to.
+Automatically adds and syncs the `data_stream.*` event fields if they are missing from the
+event. This ensures that fields match the name of the data stream that is receiving events.
 
-NOTE: when `data_stream_auto_routing` is disabled and existing `data_stream.*` event
-fields do not match the data stream name they will be overwritten (with a warning).
+NOTE: If existing `data_stream.*` event fields do not match the data stream name
+and `data_stream_auto_routing` is disabled, the event fields will be
+overwritten with a warning.
 
 [id="plugins-{type}s-{plugin}-data_stream_type"]
 ===== `data_stream_type`
@@ -471,7 +473,7 @@ fields do not match the data stream name they will be overwritten (with a warnin
 * Default value is `logs`.
 
 The data stream type used to construct the data stream at index time.
-Currently, only `logs` and `metrics` is supported.
+Currently, only `logs` and `metrics`are supported.
 
 [id="plugins-{type}s-{plugin}-doc_as_upsert"]
 ===== `doc_as_upsert`
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -274,6 +274,12 @@ This plugin supports the following configuration options plus the
 | <<plugins-{type}s-{plugin}-cloud_auth>> |<<password,password>>|No
 | <<plugins-{type}s-{plugin}-cloud_id>> |<<string,string>>|No
 | <<plugins-{type}s-{plugin}-custom_headers>> |<<hash,hash>>|No
+| <<plugins-{type}s-{plugin}-data_stream>> |<<string,string>>, one of `["true", "false", "auto"]`|No
+| <<plugins-{type}s-{plugin}-data_stream_auto_routing>> |<<boolean,boolean>>|No
+| <<plugins-{type}s-{plugin}-data_stream_dataset>> |<<string,string>>|No
+| <<plugins-{type}s-{plugin}-data_stream_namespace>> |<<string,string>>|No
+| <<plugins-{type}s-{plugin}-data_stream_sync_fields>> |<<boolean,boolean>>|No
+| <<plugins-{type}s-{plugin}-data_stream_type>> |<<string,string>>|No
 | <<plugins-{type}s-{plugin}-doc_as_upsert>> |<<boolean,boolean>>|No
 | <<plugins-{type}s-{plugin}-document_id>> |<<string,string>>|No
 | <<plugins-{type}s-{plugin}-document_type>> |<<string,string>>|No
@@ -335,23 +341,22 @@ output plugins.
 ===== `action`
 
   * Value type is <<string,string>>
-  * Default value is `"index"`
+  * Default value depends on whether data-streams are used.
 
-Protocol agnostic (i.e. non-http, non-java specific) configs go here
-Protocol agnostic methods
 The Elasticsearch action to perform. Valid actions are:
 
-- index: indexes a document (an event from Logstash).
-- delete: deletes a document by id (An id is required for this action)
-- create: indexes a document, fails if a document by that id already exists in the index.
-- update: updates a document by id. Update has a special case where you can upsert -- update a
+- `index`: indexes a document (an event from Logstash).
+- `delete`: deletes a document by id (An id is required for this action)
+- `create`: indexes a document, fails if a document by that id already exists in the index.
+- `update`: updates a document by id. Update has a special case where you can upsert -- update a
   document if not already present. See the `doc_as_upsert` option. NOTE: This does not work and is not supported
   in Elasticsearch 1.x. Please upgrade to ES 2.x or greater to use this feature with Logstash!
 - A sprintf style string to change the action based on the content of the event. The value `%{[foo]}`
   would use the foo field for the action
 
-For more details on actions, check out the {ref}/docs-bulk.html[Elasticsearch
-bulk API documentation].
+When using data-streams a `create` default is assumed, otherwise the default action is `"index"`.
+
+For more details on actions, check out the {ref}/docs-bulk.html[Elasticsearch bulk API documentation].
 
 [id="plugins-{type}s-{plugin}-api_key"]
 ===== `api_key`
@@ -404,6 +409,69 @@ Cloud ID, from the Elastic Cloud web console. If set `hosts` should not be used.
 
 For more details, check out the
 {logstash-ref}/connecting-to-cloud.html[Logstash-to-Cloud documentation].
+
+[id="plugins-{type}s-{plugin}-data_stream"]
+===== `data_stream`
+
+* Value can be any of: `true`, `false` and `auto`
+* Default value depends on Logstash version.
+
+Defines whether data will be indexed into an Elasticsearch data stream.
+The other `data_stream_*` settings will only be used if this setting is enabled.
+Defaults to `false` in Logstash 7.x and `auto` starting in Logstash 8.0.
+
+The `auto` behavior means that whenever the supplied configuration is compatible
+with data streams Logstash will assume to be handling events as if `true` was set.
+
+[id="plugins-{type}s-{plugin}-data_stream_auto_routing"]
+===== `data_stream_auto_routing`
+
+* Value type is <<boolean,boolean>>
+* Default value is `true`.
+
+Automatically routes events by deriving the data stream name using specific event
+fields with the `%{data_stream.type}-%{data_stream.dataset}-%{data_stream.namespace}` format.
+
+If enabled, the `data_stream.*` event fields will take precedence over the
+`data_stream_type`, `data_stream_dataset`, and `data_stream_namespace` settings,
+but will fall back to them if any of the fields are missing from the event.
+
+[id="plugins-{type}s-{plugin}-data_stream_dataset"]
+===== `data_stream_dataset`
+
+* Value type is <<string,string>>
+* Default value is `generic`.
+
+The data stream dataset used to construct the data stream at index time.
+
+[id="plugins-{type}s-{plugin}-data_stream_namespace"]
+===== `data_stream_namespace`
+
+* Value type is <<string,string>>
+* Default value is `default`.
+
+The data stream namespace used to construct the data stream at index time.
+
+[id="plugins-{type}s-{plugin}-data_stream_sync_fields"]
+===== `data_stream_sync_fields`
+
+* Value type is <<boolean,boolean>>
+* Default value is `true`
+
+Automatically syncs the `data_stream.*` event fields if they are missing from the
+event, this ensures fields match the data stream name that events are being sent to.
+
+NOTE: when `data_stream_auto_routing` is disabled and existing `data_stream.*` event
+fields do not match the data stream name they will be overwritten (with a warning).
+
+[id="plugins-{type}s-{plugin}-data_stream_type"]
+===== `data_stream_type`
+
+* Value type is <<string,string>>
+* Default value is `logs`.
+
+The data stream type used to construct the data stream at index time.
+Currently, only `logs` and `metrics` is supported.
 
 [id="plugins-{type}s-{plugin}-doc_as_upsert"]
 ===== `doc_as_upsert`
--- a/lib/logstash/outputs/elasticsearch.rb
+++ b/lib/logstash/outputs/elasticsearch.rb
@@ -252,7 +252,9 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
   config :ilm_policy, :validate => :string, :default => DEFAULT_POLICY
 
   attr_reader :client
-  attr_reader :default_index, :default_ilm_rollover_alias, :default_template_name
+  attr_reader :default_index
+  attr_reader :default_ilm_rollover_alias
+  attr_reader :default_template_name
 
   def initialize(*params)
     super
--- a/lib/logstash/outputs/elasticsearch.rb
+++ b/lib/logstash/outputs/elasticsearch.rb
@@ -450,6 +450,8 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
 
   def install_template
     TemplateManager.install_template(self)
+  rescue => e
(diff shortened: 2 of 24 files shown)


kares:
will be :green_circle: after https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/993

lib/logstash/outputs/elasticsearch/template_manager.rb
```diff
@@ -13,10 +13,8 @@ def self.install_template(plugin)
       end
 
       add_ilm_settings_to_template(plugin, template) if plugin.ilm_in_use?
-      plugin.logger.info("Attempting to install template", :manage_template => template)
+      plugin.logger.debug("Attempting to install template", template: template)
```
kares, line 16: there's already a `Using mapping template from|Using a default mapping template` info line.

spec/unit/outputs/elasticsearch/data_stream_support_spec.rb
```diff
@@ -0,0 +1,497 @@
...
+
+      let(:event) do
+        super().tap do |event|
+          event.set '[data_stream][type]', 'foo'
+          event.set '[data_stream][dataset]', false
+          event.set '[data_stream][extra]', 0
+        end
+      end
+
+      it 'overwrites invalid data_stream sub-fields' do
```
kares, line 377: there was a previous discussion on the topic: https://github.com/logstash-plugins/logstash-output-elasticsearch_data_streams/issues/2
the outcome isn't clear but given the importannce of `data_stream.*` fields to be in sync -> behavior is to overwrite the `dataset/namespace/type` fields if the existing values do not match (a warning is printed in the log when doing so)

spec/unit/outputs/elasticsearch/data_stream_support_spec.rb
```diff
@@ -0,0 +1,497 @@
...
+    #   expect( subject.logger ).to receive(:warn) do |msg|
+    #     msg.index "Configuration is data_stream compliant but won't be used"
+    #   end
+    #   change_constant :LOGSTASH_VERSION, '8.0.1' do
+    #     change_constant :OSS, true, target: LogStash do
+    #       expect( subject.data_stream_config? ).to be false
+    #     end
+    #   end
+    # end
+    it "uses data-streams on LS 8.0 (OSS)" do
```
kares, line 113: :scissors: here as a reminder that we diverged from the original plan

> Logstash 8.0 OSS + Data Streams compatible settings: Default to index, warn / log to suggest use of default ES + data stream for better experience

we now behave the same regardless of whether LS is OSS

spec/unit/outputs/elasticsearch/data_stream_support_spec.rb
```diff
@@ -0,0 +1,497 @@
...
+      it "fails to start on LS 8.0" do
+        change_constant :LOGSTASH_VERSION, '8.0.0' do
+          change_constant :OSS, false, target: LogStash do
+            expect { subject.data_stream_config? }.to raise_error(LogStash::ConfigurationError, /data_stream is only supported since Elasticsearch/)
+          end
+        end
+      end
+
+      # it "uses index-ing on OSS LS 8.0" do
+      #   change_constant :LOGSTASH_VERSION, '8.0.0' do
```
kares, line 134: :scissors: here as a reminder that we diverged from the original plan

> Logstash 8.0 OSS + Data Streams compatible settings: Default to index, warn / log to suggest use of default ES + data stream for better experience

we now behave the same regardless of whether LS is OSS

docs/index.asciidoc
```diff
@@ -405,6 +410,69 @@ Cloud ID, from the Elastic Cloud web console. If set `hosts` should not be used.
 For more details, check out the
 {logstash-ref}/connecting-to-cloud.html[Logstash-to-Cloud documentation].
 
+[id="plugins-{type}s-{plugin}-data_stream"]
+===== `data_stream`
```
kares, line 414: these are mostly just :scissors: from the [spec](https://github.com/elastic/logstash/issues/12178) could certainly use more data-stream related chit-chat ... // cc @karenzone

spec/unit/outputs/elasticsearch/data_stream_support_spec.rb
```diff
@@ -0,0 +1,529 @@
...
+    end
+
+    before { allow(subject).to receive(:last_es_version).and_return(es_version) }
+
+    it "does not use data-streams by default" do
+      expect( subject.logger ).to receive(:warn) do |msg|
+        expect(msg).to include "Ambiguous configuration, data stream settings have no effect"
+      end
+      change_constant :LOGSTASH_VERSION, '7.10.2' do
+        expect { subject.data_stream_config? }.to raise_error(LogStash::ConfigurationError, /Ambiguous configuration/i)
```
kares, line 165: slightly more aggressive than specified initially - should be the best for the user.
on LS 7.x the default behavior is to use index-ing but given the user sets a data_stream_ param e.g.

`data_stream_dataset => test ...` in such cases we rather force to specify `data_stream => true/false`

lib/logstash/outputs/elasticsearch/data_stream_support.rb
```diff
@@ -0,0 +1,249 @@
...
+      def validate_value(value, validator)
+        if validator == :dataset_identifier
+          return validate_dataset_identifier(value)
+        end
+        if validator == :namespace_identifier
+          return validate_namespace_identifier(value)
+        end
+        return super unless validator == :ds_identifier_string
+
+        super(value, validator)
```
yaauie, line 210: This bit is redundant:
~~~
        return super unless validator == :ds_identifier_string

        super(value, validator)
~~~

Addititonally, I think this can be made more clear by using a `case` statement:

```suggestion
        case validator
        when :dataset_identifier   then validate_dataset_identifier(value)
        when :namespace_identifier then validate_namespace_identifier(value)
        else super
        end
```

The comment at the top about which validators it provides needs updating since `:ds_identifier_string` is no longer useful.

lib/logstash/outputs/elasticsearch/ilm.rb
```diff
@@ -4,21 +4,12 @@ module Ilm
     ILM_POLICY_PATH = "default-ilm-policy.json"
 
     def setup_ilm
-      return unless ilm_in_use?
```
yaauie, line 7: nice removal of redundant check

lib/logstash/plugin_mixins/elasticsearch/api_configs.rb
```diff
@@ -5,159 +5,163 @@ module APIConfigs
...
+
+        # Set initial interval in seconds between bulk retries. Doubled on each retry up to `retry_max_interval`
+        :retry_initial_interval => { :validate => :number, :default => 2 },
+
+        # Set max interval in seconds between bulk retries.
+        :retry_max_interval => { :validate => :number, :default => 64 }
+    }.freeze
+
+    def self.included(base)
+      CONFIG_PARAMS.each { |name, opts| base.config(name, opts) }
```
yaauie, line 164: clever 👍🏼

lib/logstash/plugin_mixins/elasticsearch/common.rb
```diff
@@ -339,5 +344,15 @@ def dlq_enabled?
       respond_to?(:execution_context) && execution_context.respond_to?(:dlq_writer) &&
         !execution_context.dlq_writer.inner_writer.is_a?(::LogStash::Util::DummyDeadLetterQueueWriter)
     end
+
+    EMPTY_HASH = {}.freeze
+
+    def dig_value(val, *keys)
```
yaauie, line 350: this implementation won't work if the thing at the key being dug is actually an empty hash -- it'll return `nil` instead of the object that exists at that address.

Although our one specific use is not likely to hit this issue (or if it does, we already have other problems), I'm wary of implementing a `dig` that behaves differently than the one we know and love from Ruby 2.3+.

Here is an implementation that borrows heavily from [ruby backports](https://github.com/marcandre/backports/blob/71bd2a5ba0c64e1f7ef90503d3f7d3ea167b5028/lib/backports/2.3.0/hash/dig.rb) without monkey-patching `Hash`:

~~~ ruby
def dig_value(val, first_key, *rest_keys)
  fail(TypeError, "cannot dig value from #{val.class}") unless val.kind_of?(Hash)
  val = self[first_key]
  return val if rest_keys.empty? || val == nil
  dig_value(val, *rest_keys)
end
~~~

yaauie (commented):
First pass: 🎉🎉🎉

There is a lot going on here and from what I can tell in my first pass it does what you say it does. I have left a whole lot of comments in-line, many of which are nitpicks or suggestions for readability.

The two main take-aways I have are:
 - mutating the `event` in the output is generally not desirable, because this can end up surprising people who have pipelines with multiple outputs.
 - some of the new synchronization I think is only useful if we are already experiencing double-sends of `register`, which is likely to cause other problems.

lib/logstash/outputs/elasticsearch.rb
```diff
@@ -267,18 +271,31 @@ def register
...
     end
+
     @bulk_request_metrics = metric.namespace(:bulk_requests)
     @document_level_metrics = metric.namespace(:documents)
     @logger.info("New Elasticsearch output", :class => self.class.name, :hosts => @hosts.map(&:sanitized).map(&:to_s))
   end
 
+  # @override post-register when ES connection established
+  def finish_register
+    synchronize do # due @i-vars being set on the plugin
```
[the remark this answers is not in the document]
kares: so the idea here is to impose a barrier on the field writes - not to prevent concurrent access.

> due @i-vars being set on the plugin

essentially `after_successful_connection` executes on a separate thread and the plugin might be waiting on that thread to complete (from `multi_receive`) by checking a volatile variable. we've been previously setting instance state on a plugin such as `@index = ...` from the after connection thread. in theory there's no visibility guarantee - a mutex should resolve these concerns without resorting several i-vars to atomic references (as there's now more instance state being updated from `finish_register`).

otherwise yeah I understand how this is confusing and probably still not 100% safe.

lib/logstash/outputs/elasticsearch/data_stream_support.rb
```diff
@@ -0,0 +1,249 @@
...
+      namespace = data_stream['namespace'] || data_stream_namespace
+      "#{type}-#{dataset}-#{namespace}"
+    end
+
+    # @param params the user configuration for the ES output
+    # @note LS initialized configuration (with filled defaults) won't detect as data-stream
+    # compatible, only explicit (`original_params`) config should be tested.
+    # @return [TrueClass|FalseClass] whether given configuration is data-stream compatible
+    def check_data_stream_config!(params = original_params)
+      use_data_stream = data_stream_explicit?
```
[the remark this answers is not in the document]
kares: sorry about this I tend to use `method?` as `true`/`false` and maybe `nil` - quite a habit of mine. updated!

lib/logstash/outputs/elasticsearch/data_stream_support.rb
```diff
@@ -0,0 +1,249 @@
...
+          @logger.warn "Invalid 'data_stream' field type, due fields sync will overwrite", value: data_stream, event: event.to_hash
+        end
+        data_stream = Hash.new
+      end
+
+      data_stream['type'] ||= data_stream_type
+      data_stream['dataset'] ||= data_stream_dataset
+      data_stream['namespace'] ||= data_stream_namespace
+
+      event.set('data_stream', data_stream)
```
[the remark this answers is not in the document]
kares: :cupid:  for the hint - did not realize we should not be updating events in an output.

lib/logstash/outputs/elasticsearch.rb
```diff
@@ -247,17 +251,17 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
   # ILM policy to use, if undefined the default policy will be used.
   config :ilm_policy, :validate => :string, :default => DEFAULT_POLICY
 
-  attr_reader :default_index
-  attr_reader :default_ilm_rollover_alias
-  attr_reader :default_template_name
+  attr_reader :client
```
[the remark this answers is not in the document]
kares: :+1: have double checked the usage from LS before doing the change, both use-cases are fine:
```ruby
      def build_client
        es = LogStash::Outputs::ElasticSearch.new(@es_options)
        es.instance_variable_set :@logger, logger
        es.build_client
      end
```
and https://github.com/elastic/logstash/blob/v7.11.1/x-pack/lib/config_management/elasticsearch_source.rb#L122-L125

lib/logstash/outputs/elasticsearch/data_stream_support.rb
```diff
@@ -0,0 +1,249 @@
...
+    # @note LS initialized configuration (with filled defaults) won't detect as data-stream
+    # compatible, only explicit (`original_params`) config should be tested.
+    # @return [TrueClass|FalseClass] whether given configuration is data-stream compatible
+    def check_data_stream_config!(params = original_params)
+      use_data_stream = data_stream_explicit?
+      data_stream_params = params.select { |name, _| name.start_with?('data_stream_') } # exclude data_stream =>
+      invalid_data_stream_params = invalid_data_stream_params(params)
+
+      if use_data_stream.eql?(false) && data_stream_params.any?
+        @logger.debug "Ignoring data stream specific configuration (due data_stream => false)", data_stream_params
```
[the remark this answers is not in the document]
kares: sorry my bad - went back and forth overthinking user expectations.
changed this back to a warn-ing, the [idea](https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/988#discussion_r587735157) (as spec-ed) is that having:
- `'hosts' => ['es.host:9200'] 'data_stream_auto_routing' => false` is an error (we do not default to DS on 7.x)
- `'hosts' => ['es.host:9200'] 'data_stream_auto_routing' => false 'data_stream' => false` is a warn-ing

the original intent for not warn-ing was that with `'data_stream' => false` user is clearly aware what's going on :monkey_face:

docs/index.asciidoc
```diff
@@ -405,6 +410,69 @@ Cloud ID, from the Elastic Cloud web console. If set `hosts` should not be used.
...
+
+* Value can be any of: `true`, `false` and `auto`
+* Default value depends on Logstash version.
+
+Defines whether data will be indexed into an Elasticsearch data stream.
+The other `data_stream_*` settings will only be used if this setting is enabled.
+Defaults to `false` in Logstash 7.x and `auto` starting in Logstash 8.0.
+
+The `auto` behavior means that whenever the supplied configuration is compatible
```
karenzone, line 423:
```suggestion
Logstash handles the output as a data stream when the supplied configuration
is compatible with data streams and this value is set to `auto`.
```

docs/index.asciidoc
```diff
@@ -405,6 +410,69 @@ Cloud ID, from the Elastic Cloud web console. If set `hosts` should not be used.
...
+but will fall back to them if any of the fields are missing from the event.
+
+[id="plugins-{type}s-{plugin}-data_stream_dataset"]
+===== `data_stream_dataset`
+
+* Value type is <<string,string>>
+* Default value is `generic`.
+
+The data stream dataset used to construct the data stream at index time.
+
```
karenzone, line 446: Are there other valid entries at this time?

docs/index.asciidoc
```diff
@@ -405,6 +410,69 @@ Cloud ID, from the Elastic Cloud web console. If set `hosts` should not be used.
...
+The data stream namespace used to construct the data stream at index time.
+
+[id="plugins-{type}s-{plugin}-data_stream_sync_fields"]
+===== `data_stream_sync_fields`
+
+* Value type is <<boolean,boolean>>
+* Default value is `true`
+
+Automatically syncs the `data_stream.*` event fields if they are missing from the
+event, this ensures fields match the data stream name that events are being sent to.
```
karenzone, line 462:
```suggestion
Automatically adds and syncs the `data_stream.*` event fields if they are missing from the
event. This ensures that fields match the name of the data stream that is receiving events.
```

docs/index.asciidoc
```diff
@@ -405,6 +410,69 @@ Cloud ID, from the Elastic Cloud web console. If set `hosts` should not be used.
...
+
+NOTE: when `data_stream_auto_routing` is disabled and existing `data_stream.*` event
+fields do not match the data stream name they will be overwritten (with a warning).
```
karenzone, line 465:
```suggestion
NOTE: If existing `data_stream.*` event fields do not match the data stream name
and `data_stream_auto_routing` is disabled, the event fields will be
overwritten with a warning.
```

docs/index.asciidoc
```diff
@@ -405,6 +410,69 @@ Cloud ID, from the Elastic Cloud web console. If set `hosts` should not be used.
...
+
+[id="plugins-{type}s-{plugin}-data_stream_type"]
+===== `data_stream_type`
+
+* Value type is <<string,string>>
+* Default value is `logs`.
+
+The data stream type used to construct the data stream at index time.
+Currently, only `logs` and `metrics` is supported.
```
karenzone, line 474:
```suggestion
Currently, only `logs` and `metrics`are supported.
```

karenzone (commented):
Thanks for the great start you have on docs.  I left some comments and suggestions inline for your consideration. 

Also, will you please put this text near the beginning of the document, starting immediately after "==== Description"?

> Elasticsearch provides near real-time search and analytics for all types of
> data. The Elasticsearch output plugin can store both time series datasets (such
> as logs, events, and metrics) and non-time series data in Elasticsearch.

docs/index.asciidoc
kares: Believe anything is valid (as long as it's a valid name) in ES. 
Only the `data_stream_type` is contrained with enumerated values.

kares:
Hey @karenzone sorry took so long, have adressed your documentation concerns and added:
> Elasticsearch provides near real-time search and analytics for all types of
data. The Elasticsearch output plugin can store both time series datasets (such
as logs, events, and metrics) and non-time series data in Elasticsearch.

Please double check for me or if you have anything else to add :heart:

karenzone:
Thanks for updating the docs, @kares.  Docs look good for now. 👍

lib/logstash/outputs/elasticsearch.rb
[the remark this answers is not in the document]
yaauie: 👍🏼 so many layers to think through here, good analysis.

lib/logstash/outputs/elasticsearch/data_stream_support.rb
```diff
@@ -0,0 +1,247 @@
...
+    # @note LS initialized configuration (with filled defaults) won't detect as data-stream
+    # compatible, only explicit (`original_params`) config should be tested.
+    # @return [TrueClass|FalseClass] whether given configuration is data-stream compatible
+    def check_data_stream_config!(params = original_params)
+      use_data_stream = data_stream_explicit_value
+      data_stream_params = params.select { |name, _| name.start_with?('data_stream_') } # exclude data_stream =>
+      invalid_data_stream_params = invalid_data_stream_params(params)
+
+      if use_data_stream.eql?(false) && data_stream_params.any?
+        @logger.warn "Ignoring data stream specific configuration (due data_stream => false)", data_stream_params
```
yaauie, line 60: I think a refactor dropped my first-pass review bits on this behaviour.

When a user explicitly specifies `data_stream => false`, and also provides `data_stream`-specific settings, I would consider it invalid configuration.

A warning here will easily be missed and recovery of data may be non-trivial, and we can easily prevent the user from accidentally getting themselves in this situation.

lib/logstash/outputs/elasticsearch/data_stream_support.rb
```diff
@@ -0,0 +1,247 @@
...
+      end
+
+      if use_data_stream.nil?
+        use_data_stream = data_stream_default(invalid_data_stream_params.empty?)
+        if !use_data_stream && data_stream_params.any?
+          # DS (auto) disabled but there's still some data-stream parameters (and no `data_stream => false`)
+          @logger.warn "Ambiguous configuration, data stream settings have no effect", data_stream_params
+          raise LogStash::ConfigurationError, "Ambiguous configuration, please set data_stream => true/false " +
```
yaauie, line 68: If this is run in the `finish_register` hook in another thread, I don't think we can guarantee that we properly cascade the error and prevent the pipeline from starting. For example, if there is some latency, `Plugin#register` could have already returned before a connection is established and the framework we have for dealing with `LogStash::ConfigurationError` would have been bypassed and we will instead crash the output after it has possibly begun processing events.

We _may_ need to flip things around and block in `register` until the `finish_register` has succeeded.

The other option I can see is to _not_ rely on the state of Elasticsearch and to simply default to using datastreams in Logstash 8 regardless of what we are connected to, possibly guarded by whether Logstash itself is Licensed or OSS-only.

yaauie (commented):
Second pass: overall looking on track.

I've raised a question about validation happening out-of-band once a connection has been established. I'd like to see how that behaviour plays out in a real pipeline with multiple outputs to ensure that a bad config doesn't leave a pipeline in a partially-up state. Ideally, we prevent pipelines from starting at all if they have bad config, so relying on the external state of Elasticsearch makes this a bit precarious.

lib/logstash/outputs/elasticsearch/data_stream_support.rb
kares: Good point, we could retrieve the error on the LS worker thread but at this point I would rather try:
> We may need to flip things around and block in register until the finish_register has succeeded.

kares: Looked into this and waiting for a successful connection in `register` is problematic - by blocking we effectively disable the pipeline setup. Now we could try to be smart with a timeout but it's making things even more complicated.

A reasonable resolution is for the DS setup to **NOT** rely on ES version check and effectively always default to `data_stream => false` (assuming no `data_stream...` configuration present) on LS 7.x, potentially with an info message that the default will change in LS 8.0. I think the attempt to play things smart wasn't a good idea ...

spec/integration/outputs/ilm_spec.rb
```diff
@@ -198,12 +198,16 @@
     let (:settings) { super().merge!({ 'ilm_enabled' => true }) }
 
     it 'should raise a configuration error' do
+      # TODO should be refactored not to rely on plugin internals
```
kares, line 201: integration test was relying on a raise from the other thread to propagate, which is problematic.
the new logic to rescue everything from the async thread and propagate exceptions back as `thread.value` - making it a bit hard to test -> spec example should be rewritten or migrated to a unit test.

yaauie (commented):
👍🏼 good cleanup on the async auto-DS.

I've got some suggestions for making the validation a little easier to read, perhaps detangling some of the logic that ended up getting a bit twisty.

karenzone (commented):
Covers the basics and builds cleanly. I recommend that we merge this huge body of work when it's ready, and immediately follow up with a docs-only PR. Suggestions for enhancements are here: https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/1006. These kinds of suggestions/enhancements are hard to notate and implement in the github workflow. LMKWYT about this approach.

UPDATED: Please add a data streams section (similar to the ILM section), and include the two configs from the [issue](https://github.com/elastic/logstash/issues/12178). We can handle and track the other doc improvements as part of the follow up issue: https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/1006

lib/logstash/outputs/elasticsearch/data_stream_support.rb
```diff
@@ -0,0 +1,235 @@
...
+        if !use_data_stream && data_stream_params.any?
+          # DS (auto) disabled but there's still some data-stream parameters (and no `data_stream => false`)
+          @logger.warn "Ambiguous configuration, data stream settings have no effect", data_stream_params
+          raise LogStash::ConfigurationError, "Ambiguous configuration, please set data_stream => true " +
+                                              "or remove data stream specific settings: #{data_stream_params.keys}"
+        end
+      end
+
+      if use_data_stream
+        if invalid_data_stream_params.any?
```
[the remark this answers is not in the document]
kares: :cupid: very nice catch, it started with more logic originally but was removed and makes so much sense to simplify!

yaauie (approved):
LGTM :tada:

PR #992, opened 2021-02-16
Add example under hosts setting

roaksoax:
Adds an example under hosts setting showing users how to use an environment variable with multiple white-space delimited hosts.

### DIFF
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -523,6 +523,13 @@ should reference only data or client nodes in Elasticsearch.
 Any special characters present in the URLs here MUST be URL escaped! This means
 `#` should be put in as `%23` for instance.
 
+This setting supports defining an environment variable `hosts => '${ES_HOSTS:}',
+with multiple {logstash-ref}/plugin-concepts.html[whitespace-delimited URIs], e.g.:
+
+```
+ES_HOSTS="es1.example.com es2.example.com:9201 es3.example.com:9201"
+```
+
 [id="plugins-{type}s-{plugin}-http_compression"]
 ===== `http_compression`
 


docs/index.asciidoc
```diff
@@ -523,6 +523,13 @@ should reference only data or client nodes in Elasticsearch.
 Any special characters present in the URLs here MUST be URL escaped! This means
 `#` should be put in as `%23` for instance.
 
+This setting supports defining an environment variable `hosts => '${ES_HOSTS:}',
```
robbavey, line 526: Do we need to add a note that this only supported on Logstash `7.9` and above?

docs/index.asciidoc
```diff
@@ -523,6 +523,13 @@ should reference only data or client nodes in Elasticsearch.
 Any special characters present in the URLs here MUST be URL escaped! This means
 `#` should be put in as `%23` for instance.
 
+When using Logstash 7.9+, this setting supports defining an environment variable
+`hosts => '${ES_HOSTS:}' with multiple {logstash-ref}/plugin-concepts.html[whitespace-delimited URIs], e.g.:
```
robbavey, line 527: Maybe we should link to the specific anchor for the subject (`{logstash-ref}/plugin-concepts.html#space-delimited-uris-in-list-params`), rather than the general plugin concepts page? The page looks like a general one that could be expanded with further entries over time.

docs/index.asciidoc
```diff
@@ -523,6 +523,13 @@ should reference only data or client nodes in Elasticsearch.
...
+
+```
+ES_HOSTS="es1.example.com es2.example.com:9201 es3.example.com:9201"
```
robbavey, line 530: Does it make sense to add further examples using the different schemes detailed above?

docs/index.asciidoc
robbavey, line 527: Do you mean `ES_HOSTS` rather than `ES_HOSTS:`?

robbavey, line 527: Typo: Non-matching backticks/quotes

robbavey (changes requested):
Couple of typos, and some content queries

docs/index.asciidoc
```diff
@@ -523,6 +523,13 @@ should reference only data or client nodes in Elasticsearch.
...
+```
```
karenzone, line 531: Suggested text/example:

>{ls} 7.9 and later supports defining an environment variable `hosts =>
'${ES_HOSTS}' with multiple whitespace delimited URIs. 
This plugin automatically expands `hosts` strings that contain multiple
white-space delimited URs into separate entries. 

>**Example**
>```
>ES_HOSTS="es1.example.com es2.example.com:9201 es3.example.com:9201"
>```

karenzone (commented):
Thank you for adding this helpful example. It definitely makes the docs better.  I left some suggested text inline. Note that I removed the link to the concept page.  Instead, I lifted the critical part of the text and included it here.

docs/index.asciidoc
karenzone: ++ On adding further examples for the other uses.

docs/index.asciidoc
roaksoax: "This plugin automatically expands hosts strings that contain multiple
white-space delimited URs into separate entries." - I feel like this is very developer centric and less easy to understand.

robbavey: What about something like 

> This setting supports defining an environment variable containing one or more hostnames, each separated by whitespace 

Followed by an example of the environment variable definition(s), and how to use it/them in the configuration. With a note that prior to `7.9`, the environment variable would only support a single hostname?

karenzone: ++

docs/index.asciidoc
```diff
@@ -523,6 +523,16 @@ should reference only data or client nodes in Elasticsearch.
 Any special characters present in the URLs here MUST be URL escaped! This means
 `#` should be put in as `%23` for instance.
 
+When using Logstash 7.9+, this setting supports defining an environment variable
+`hosts => '${S}'` with multiple {logstash-ref}/plugin-concepts.html#space-delimited-uris-in-list-params[whitespace-delimited URIs]:
```
karenzone, line 527:
```suggestion
This option can accept an environment variable containing one or more hostnames separated by whitespace.
Strings separated by whitespace are treated as separate entries.
```

karenzone (commented):
Added suggestion based on what we discussed above.

Issue #915, opened 2020-02-01
Specifying types in bulk requests is deprecated


fdartayre:
**Version:** logstash-output-elasticsearch 10.2.3 (shipping with Logstash 7.5.2)

[Starting from Elasticsearch 7.0](https://www.elastic.co/guide/en/elasticsearch/reference/current/removal-of-types.html#_document_apis), types should also no longer appear in the body of bulk requests.

Including a type causes a deprecation warning to get logged in Elasticsearch:
```
[2020-01-31T10:58:52,319][WARN ][o.e.d.a.b.BulkRequestParser] [hostname] [types removal] Specifying types in bulk requests is deprecated.
```
The plugin shouldn't use types anymore.

andsel:
With commit https://github.com/logstash-plugins/logstash-output-elasticsearch/commit/c93abd1e6255d5ba519e942a85ac708c05d472a7 present in version `10.2.2` type was removed but with https://github.com/logstash-plugins/logstash-output-elasticsearch/commit/39a4db99008704553afb885a8ad9f0fbd44984fb was added because the `_monitoring/bulk` operation required it to continue work. With the LS 7.7 the end point should be switched and then removed

rolffokkens:
Currently at LS 7.10, end the ES (also 7.10) logs are flouding my filesystem. No idea what the previous comment means.

barosch78:
> Currently at LS 7.10, end the ES (also 7.10) logs are flouding my filesystem. No idea what the previous comment means. The flouding ("[types removal] Specifying types in bulk requests is deprecated.") started after upgrading ES from 7.9 to 7.10.

I have the same problem after upgrading to 7.10

nosferatu314:
Same problem here. Flooding can be stopped by changing log level of deprecation warnings.

From: https://www.elastic.co/guide/en/elasticsearch/reference/current/logging.html

You can disable it in the config/log4j2.properties file by setting the deprecation log level to error like this:
```
logger.deprecation.name = org.elasticsearch.deprecation
logger.deprecation.level = error
```

This is merely a workaround and I would appreciate a fix.

MakoWish:
Any updates on this issue?

andsel:
Could you share some information about this problem please, which `logstash-output-elasticseach` configuration are you using?Did you remove the `type` from your config definition when pushing data to Elasticsearch?

MakoWish:
Hi @andsel,

[...]

PR #994, opened 2021-02-18
Remove `type` when connecting to ES 7 and user doesn't specify it

andsel:
With ES 7.10.0 a deprecation log line about types removal is printed for each request (index/bulk) that contains the deprecated index `_type`.
This led to possible flood of deprecation logs on Elasticsearch, tracked in https://github.com/elastic/elasticsearch/issues/69188.

This commit removes the implicit set of type when connected to ES7 unless the user doesn't specify explicitly the type.

Related to #915

### DIFF
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -40,7 +40,8 @@ the website landing page or in the {ref}[Elasticsearch documentation].
 [NOTE]
 ================================================================================
 When connected to Elasticsearch 7.x, modern versions of this plugin
-use the required `_doc` document-type when inserting documents.
+don't use the document-type when inserting documents, unless the user
+explicitly sets <<plugins-{type}s-{plugin}-document_type>>.
 
 If you are using an earlier version of Logstash and wish to connect to
 Elasticsearch 7.x, first upgrade Logstash to version 6.8 to ensure it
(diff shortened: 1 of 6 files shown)


jsvd (approved):
LGTM
I manually tested a few scenarios:
* plain `output { elasticsearch {}}`: no deprecation logs, documents indexed with _doc
* `output { elasticsearch { document_type => meh index => test }}`:  confirmed "meh
 was used as the document type AND that ES logged a deprecation warning
* `output { elasticsearch { index => test }}`: confirmed that in the absence of type, documents were indexed with the index's type (in this case, "meh"). No deprecation log was emitted.

docs/index.asciidoc
```diff
@@ -40,7 +40,8 @@ the website landing page or in the {ref}[Elasticsearch documentation].
 [NOTE]
 ================================================================================
 When connected to Elasticsearch 7.x, modern versions of this plugin
-use the required `_doc` document-type when inserting documents.
+don't use the document-type when inserting documents, unless the user
+ecplicitly sets <<plugins-{type}s-{plugin}-document_type>>.
```
jsvd, line 44:
```suggestion
ecplicitly sets <<plugins-{type}s-{plugin}-document_type>>.
```
```suggestion
explicitly sets <<plugins-{type}s-{plugin}-document_type>>.
```

PR #1005, opened 2021-04-07
Feat: always check ES license

### LINKED ISSUE #1004: LS should always perform ES license check
Historically the output did not do a license check on OSS Logstash.
However with the recent ES licensing change (since 7.11) there's no OSS ES artifact anymore.
Thus the distinction makes no sense.

The license check is really simple basically assuring the returned JSON has [`license.status == 'active'`](https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/v10.8.4/lib/logstash/outputs/elasticsearch/license_checker.rb#L39)

A resolution should simply be removing a line to [skip the check under OSS](https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/v10.8.4/lib/logstash/outputs/elasticsearch/license_checker.rb#L14).

Potential hiccups: ES versions older than 7.11. 
The plugin would need to check the version first and skip the license check (under OSS) for ES < 7.11 for backwards compatibility.

kares:
Historically the output did not do a license check on OSS Logstash.
However with the recent ES licensing change (since 7.11) there's no OSS ES artifact anymore.
Thus the distinction makes no sense.

see #1004


robbavey (approved):
Code LGTM - providing @acchen97 and/or @roaksoax approve the change in deprecation warning text

roaksoax (approved):
lgtm!

roaksoax (changes requested):
This PR needs to ensure we are sending data to licensed ES.

roaksoax (approved):
LGTM!

Issue #1006, opened 2021-04-07
[Doc] Expand documentation for data streams


karenzone:
Related: https://github.com/elastic/logstash/issues/12178

Enhancements for consideration: 
- [ ] Expand descriptions/plugin behavior for time series and non-time series data
    - [ ] Do we need to explain difference in indexing strategy? 
- [ ] Include examples for basic default config, custom data stream name. Other examples? 
- [ ] Expand ECS section to list of ES-compatible plugins and caveat against introducing fields that conflict with ECS 
- [ ] Explain special treatment for Elastic Agent

While we're making doc changes, perhaps we can resolve items in #978 at the same time.

karenzone:
cc:/ @kares

PR #1009, opened 2021-04-12
Doc: Add data stream section and config samples

karenzone:
Related: #1006


docs/index.asciidoc
```diff
@@ -74,6 +62,60 @@ By having an ECS-compatible template in place, we can ensure that Elasticsearch
...
+
+**Example: Customize data stream name**
+
+[source,sh]
+-----
+output {
+    elasticsearch {
+        hosts => "hostname"
+        data_stream => "true"
+        data_stream_timestamp => "@timestamp"
```
kares, line 108:
```suggestion
```
we (should but) do not support customizing the timestamp field in the initial implementation available from #1006

kares (approved):
:+1: w one note on potentially not mentioning `data_stream_timestamp` for now

PR #1012, opened 2021-04-20
Fix: DLQ expects us to pass down an event

kares:
since [Data Streams](https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/988), we refactored to pass down event data directly instead of the [previous `event.to_hash` mapping in `safe_bulk`](https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/988/files#diff-221ddb62bde6d62966fab1051387ded7513fffd575cf0706a5f3a94ad2d31307L270) (doing so potentially on every retry)

there seemed to have been no reason to keep the original event around (well turns except with DLQ).

besides practical reasons the early `event.to_hash` also addressed a requirement to ... [sync data stream fields without effecting the original event](https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/v11.0.0/lib/logstash/outputs/elasticsearch/data_stream_support.rb#L148).

unfortunately this broke DLQ support, which expects a `LS::Event` passed on [`write(event, reason)`](https://github.com/elastic/logstash/blob/v7.12.0/logstash-core/src/main/java/org/logstash/common/AbstractDeadLetterQueueWriterExt.java#L195-L202)

resolves https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/1013

### DIFF
--- a/lib/logstash/outputs/elasticsearch.rb
+++ b/lib/logstash/outputs/elasticsearch.rb
@@ -392,7 +392,21 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
     params[:version] = event.sprintf(@version) if @version
     params[:version_type] = event.sprintf(@version_type) if @version_type
 
-    [action, params, event.to_hash]
+    EventActionTuple.new(action, params, event)
+  end
+
+  class EventActionTuple < Array # TODO: acting as an array for compatibility
+
+    def initialize(action, params, event, event_data = nil)
+      super(3)
+      self[0] = action
+      self[1] = params
+      self[2] = event_data || event.to_hash
+      @event = event
+    end
+
+    attr_reader :event
+
   end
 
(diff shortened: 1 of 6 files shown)


lib/logstash/outputs/elasticsearch.rb
```diff
@@ -392,7 +392,21 @@ def event_action_tuple(event)
     params[:version] = event.sprintf(@version) if @version
     params[:version_type] = event.sprintf(@version_type) if @version_type
 
-    [action, params, event.to_hash]
+    EventActionTuple.new(action, params, event)
+  end
+
+  class EventActionTuple < Array # TODO: acting as an array for compatibility
```
kares, line 398: ugly but I did not want to do a proper refactoring (replacing the `Array` with a proper struct) at this point, 
given this is meant to be a quick fix ...

andsel (approved):
LGTM

PR #1015, opened 2021-05-10
Validate that required functionality in Elasticsearch is available upon initial connection

kares:
Ensure Logstash connects to a compatible version of Elasticsearch.


robcowart:
@kares I don't see where this PR actually validates any functionality. It looks like it is just a License check. What is the point if the user is not using any features that aren't Apache-2.0?

PR #1018, opened 2021-06-04
Restore support for Elasticsearch OSS distributions 7.0.x-7.10.2

dblock:
Change #1005 had a side effect of removing compatibility with OSS distributions of Elasticsearch 7.0.x-7.10.2. This change restores Logstash-Elasticsearch compatibility outlined in the [product compatibility matrix](https://www.elastic.co/support/matrix#matrix_compatibility) by adding version-awareness prior to the license check.


bensebborn:
Any update on this please, it's breaking quite a few of our Logstash instances communicating with ES 7.9

daparky (approved):

bensebborn:
@dblock @roaksoax any update on when this will be merged in please? Trying to avoid having to edit the source code manually to work around this on production servers. Thanks for upadte.

dblock:
> @dblock @roaksoax any update on when this will be merged in please? Trying to avoid having to edit the source code manually to work around this on production servers. Thanks for upadte.

I do not have merge rights in this repo, but please do let me know if there's anything I can/need to do to get this PR merged.

FYI, see https://github.com/elastic/beats/pull/26305 for a similar proposed change in Beats.

jsvd:
The change in #1005 was introduced to simplify the logic of this plugin moving forward since Elasticsearch has a single distribution starting with 7.11. That change isn’t an isolated one, there’s a lot more we can do to make this output not have to behave differently when installed in a default or OSS distribution of Logstash.
For example, I opened up a couple of follow-up PRs that build on top of the current master:

- #1020 removes a lot of code that is no longer needed because the plugin only needs to work and be tested against 1 distribution
- #1021 builds on the PR above and further removes lingering code and tests we had to work with ES 1.x (indeed!), 2.x and 5.x

Hopefully this demonstrates the kind of simplification we’re trying to achieve and that is possible with a single Elasticsearch distribution.

For those still on a <= 7.10 ES deployment and using the OSS distribution we suggest moving to the default distribution or to keep the deployed stack in sync by using the same version of ES and Logstash.

bensebborn:
@jsvd Can you confirm the other PRs fix the current breaking issue where logstash won’t work with ES versions <=7.10.2?

Some of us are on AWS and can’t upgrade past that now. I’m really hoping this change isn’t a political one to cause problems for those on AWS (Especially as we’re a paying ElasticCloud customer too!)

dblock:
> @jsvd Can you confirm the other PRs fix the current breaking issue where logstash won’t work with ES versions <=7.10.2?

I can confirm that those PRs do not. License checks have not been removed.

bensebborn:
Well I think Elastic have well and truly shown their colours here. 

We have a split workload, with some on Elastic Cloud and some on AWS Elasticsearch. The support from Elastic is awful so we will be moving our remaining clusters over to AWS.

jsvd:
hey @bensebborn sorry to hear that you're having a bad time with support, I'll let the folks know so they can reach out and learn more about you pain points. As for using Logstash to send data to other destinations, in this case AWS maintains an output plugin for their own service, if you're having any problems with that you can create an issue in that project's repository.

mtuckerb:
It would be really handy to have a link to that output plugin.

dblock:
> It would be really handy to have a link to that output plugin (here in this issue).

https://github.com/opensearch-project/logstash-output-opensearch/tree/develop 
https://docs-beta.opensearch.org/clients/agents-and-ingestion-tools/index/

jsvd:
Also [logstash-output-amazon_es](https://github.com/awslabs/logstash-output-amazon_es) for the plugin I was mentioning. Service compatibility is also documented [in the plugin docs](https://github.com/awslabs/logstash-output-amazon_es#compatibility).

PR #1020, opened 2021-06-16
simplify plugin by removing distribution specific code

jsvd:
With ES having a single distribution we can greatly simplify the implementation and testing. This is possible because this plugin will behave the same on either default or oss Logstash.


andsel (approved):
LGTM

PR #1021, opened 2021-06-16
remove 1x, 2x and 5x specific code 

jsvd:
this change depends on https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/1020


lib/logstash/outputs/elasticsearch.rb
```diff
@@ -456,9 +456,7 @@ def get_event_type(event)
              event.sprintf(@document_type)
            else
              major_version = maximum_seen_major_version
-             if major_version < 6
-               es5_event_type(event)
```
kares, line 460: believe the (private) `es5_event_type(event)` method could be removed as well

kares (commented):
LGTM, this enabled us to simplify the code - considerably.

One minor suggestion about removing a potential dead method.

lib/logstash/outputs/elasticsearch.rb
jsvd: 🤦 removal of that method was unstaged in my local repo, solved

PR #1026, opened 2021-07-14
Add preflight check on Elasticsearch before connecting

andsel:
Update client behavior to match functionality of 7.14+

----

## Tasks
- [x] change error message from `LogStash::ConfigurationError: Not a valid Elasticsearch` to `could not connect to a compatible version of Elasticsearch`

### DIFF
--- a/lib/logstash/outputs/elasticsearch/http_client/pool.rb
+++ b/lib/logstash/outputs/elasticsearch/http_client/pool.rb
@@ -251,7 +251,12 @@ module LogStash; module Outputs; class ElasticSearch; class HttpClient;
             end
           else
             logger.info("URL doesn't point to Elasticsearch service", url: url.sanitized.to_s )
-            @state_mutex.synchronize { meta[:state] = :dead }
+            es_version = get_es_version(url)
+            @state_mutex.synchronize do
+              meta[:version] = es_version
+              set_last_es_version(es_version, url)
+              meta[:state] = :dead
+            end
           end
         rescue HostUnreachableError, BadResponseCodeError => e
           logger.warn("Attempted to resurrect connection to dead ES instance, but got an error", url: url.sanitized.to_s, exception: e.class, message: e.message)
--- a/lib/logstash/outputs/elasticsearch/http_client/pool.rb
+++ b/lib/logstash/outputs/elasticsearch/http_client/pool.rb
@@ -281,8 +281,8 @@ module LogStash; module Outputs; class ElasticSearch; class HttpClient;
       if version >= Gem::Version.new('6.0.0') && version < Gem::Version.new('7.0.0')
         return valid_tagline?(version_info)
       elsif version >= Gem::Version.new('7.0.0') && version < Gem::Version.new('7.14.0')
-        build_flavour = version_info["version"]['build_flavour']
-        return false if build_flavour.nil? || build_flavour != 'default' || !valid_tagline?(version_info)
+        build_flavor = version_info["version"]['build_flavor']
+        return false if build_flavor.nil? || build_flavor != 'default' || !valid_tagline?(version_info)
       else
         # case >= 7.14
         lower_headers = response.headers.transform_keys {|key| key.to_s.downcase }
--- a/lib/logstash/outputs/elasticsearch/http_client/pool.rb
+++ b/lib/logstash/outputs/elasticsearch/http_client/pool.rb
@@ -297,7 +297,7 @@ module LogStash; module Outputs; class ElasticSearch; class HttpClient;
 
     def valid_tagline?(version_info)
       tagline = version_info['tagline']
-     tagline == "You Know, for Search"
+      tagline == "You Know, for Search"
     end
 
     def stop_resurrectionist
--- a/lib/logstash/outputs/elasticsearch/http_client/pool.rb
+++ b/lib/logstash/outputs/elasticsearch/http_client/pool.rb
@@ -265,8 +265,12 @@ module LogStash; module Outputs; class ElasticSearch; class HttpClient;
     end
 
     def elasticsearch?(url)
-      response = perform_request_to_url(url, :get, "/")
-      return false if response.code == 401 || response.code == 403
+      begin
+        response = perform_request_to_url(url, :get, "/")
+      rescue ::LogStash::Outputs::ElasticSearch::HttpClient::Pool::BadResponseCodeError => e
+        return false if response.code == 401 || response.code == 403
+        raise e
+      end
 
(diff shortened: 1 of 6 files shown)


lib/logstash/outputs/elasticsearch/http_client/pool.rb
```diff
@@ -254,6 +261,43 @@ def healthcheck!
...
+
+      version_info = LogStash::Json.load(response.body)
+      return false if version_info['version'].nil?
+
+      version = Gem::Version.new(version_info["version"]['number'])
+      return false if version < Gem::Version.new('6.0.0')
+
+      if version >= Gem::Version.new('6.0.0') && version < Gem::Version.new('7.0.0')
+        return valid_tagline?(version_info)
+      elsif version >= Gem::Version.new('7.0.0') && version < Gem::Version.new('7.14.0')
```
jsvd, line 280: We could use Gem::Requirements here to avoid creating new objects everytime:

```ruby
     VERSION_6_TO_7 = Gem::Requirement.new([">= 6.0.0", "< 7.0.0"])
     VERSION_7_TO_7_14 = Gem::Requirement.new([">= 7.0.0", "< 7.14.0"])
```

And then:

```suggestion
      if VERSION_6_TO_7.satisfied_by?(version)
        return valid_tagline?(version_info)
      elsif VERSION_7_TO_7_14.satisfied_by?(version)
```

lib/logstash/outputs/elasticsearch/http_client/pool.rb
```diff
@@ -232,11 +232,18 @@ def health_check_request(url)
...
 
-    def healthcheck!
+    def healthcheck!(register_phase = true)
       # Try to keep locking granularity low such that we don't affect IO...
       @state_mutex.synchronize { @url_info.select {|url,meta| meta[:state] != :alive } }.each do |url,meta|
         begin
           health_check_request(url)
+
+          # when called from resurrectionist skip the product check done during register phase
+          if register_phase
```
jsvd, line 242: This makes me wonder if we could just make the `elasticsearch?` check everytime instead of HEAD'ing the `/`?

Right after we do HEAD then we request '/' in `es_version = get_es_version(url)`.

With this change we'll be doing 4 requests to ES:
1. HEAD request
2. GET '/' for `elasticsearch?` on `register_phase`
3. GET '/' to get elasticsearch version
4. GET '/_license'

We can just do instead:

1. GET '/' for healthcheck (and always confirm it's ES)
2. GET '/_license'

lib/logstash/outputs/elasticsearch/http_client/pool.rb
```diff
@@ -254,6 +261,43 @@ def healthcheck!
       end
     end
 
+    def elasticsearch?(url)
+      begin
+        response = perform_request_to_url(url, :get, "/")
```
jsvd, line 266:
```suggestion
        response = perform_request_to_url(url, :get, ROOT_URI_PATH)
```

jsvd (commented):
I believe we can simplify the healthcheck process by reducing the number of calls to ES, and always performing the product check, which shouldn't have a performance impact.

lib/logstash/outputs/elasticsearch/http_client/pool.rb
andsel: Nice, thanks! Applied

lib/logstash/outputs/elasticsearch/http_client/pool.rb
andsel: I agree with you, I think that point 2 and 3 could be merged or at least do one one HTTP GET call to `/`.
About the HTTP HEAD, that call could use a customized path from configuration (`healthcheck_path`) that some users could have customized. If we drop the HEAD call is like ignoring that setting and we should remove it.
I'm ok with removing the HEAD call, because actually seems to be used just to ping the server for reachability and then it immediately send a GET.
We could split this work in 2 steps:
- first collapse the GET call for point 2 and 3
- create a follow-up PR to drop the HEAD call and deprecate the `healthcheck_path` param

lib/logstash/outputs/elasticsearch/http_client/pool.rb
```diff
@@ -254,6 +264,43 @@ def healthcheck!
...
+      return true
+    rescue => e
+      logger.error("Unable to version version information", url: url.sanitized.to_s, exception: e.class, message: e.message)
+      false
+    end
+
+    def valid_tagline?(version_info)
+      tagline = version_info['tagline']
+      return false if tagline.nil? || tagline != "You Know, for Search"
+      true
```
jsvd, line 301: similar to the suggestion above, we can just return the result of comparing with the right tag.
```suggestion
     tagline == "You Know, for Search"
```

lib/logstash/outputs/elasticsearch/http_client/pool.rb
```diff
@@ -254,6 +264,43 @@ def healthcheck!
...
+      if VERSION_6_TO_7.satisfied_by?(version)
+        return valid_tagline?(version_info)
+      elsif VERSION_7_TO_7_14.satisfied_by?(version)
+        build_flavor = version_info["version"]['build_flavor']
+        return false if build_flavor.nil? || build_flavor != 'default' || !valid_tagline?(version_info)
+      else
+        # case >= 7.14
+        lower_headers = response.headers.transform_keys {|key| key.to_s.downcase }
+        product_header = lower_headers['x-elastic-product']
+        return false if product_header.nil? || product_header != 'Elasticsearch'
```
jsvd, line 290: If product_header is nil, it's also not equal to 'Elasticsearch':

```suggestion
        return false if product_header != 'Elasticsearch'
```

spec/unit/outputs/elasticsearch/http_client/pool_spec.rb
```diff
@@ -319,3 +351,120 @@
     end
   end
 end
+
+describe "#elasticsearch?" do
```
jsvd, line 355: <3 for all the case testing

jsvd (commented):
other than a few cosmetic refactors the change LGTM but we need a better error message, just like in the ES input and filter.

lib/logstash/outputs/elasticsearch/http_client/pool.rb
jsvd: > create a follow-up PR to drop the HEAD call and deprecate the healthcheck_path param

It's fine if there's only an issue instead to track the goal of reducing the number of network calls for a healthcheck.

PR #1032, opened 2021-08-16
Fixed SSL handshake hang indefinitely with proxy setup

kaisecheng:
Fix: #1033

The socket timeout of Apache HTTP client is unset in handshake phase (upstream [issue](https://issues.apache.org/jira/browse/HTTPCLIENT-1478?focusedCommentId=15998229&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-15998229))
This commit set the default socket config to `PoolingHttpClientConnectionManager` explicitly, so the timeout is non zero


lib/manticore/patches/client.rb
```diff
@@ -0,0 +1,21 @@
...
+        @max_pool_size = options.fetch(:pool_max, DEFAULT_MAX_POOL_SIZE)
+        pool_builder(options).tap do |cm|
+          cm.set_validate_after_inactivity options.fetch(:check_connection_timeout, 2_000)
+          cm.set_default_max_per_route options.fetch(:pool_max_per_route, @max_pool_size)
+          cm.set_max_total @max_pool_size
+
+          socket_config_builder = SocketConfig.custom
+          socket_config_builder.set_so_timeout(options.fetch(:socket_timeout, DEFAULT_SOCKET_TIMEOUT) * 1000)
+          socket_config_builder.set_tcp_no_delay(options.fetch(:tcp_no_delay, true))
+          cm.set_default_socket_config socket_config_builder.build
```
kaisecheng, line 14: set the socket config to connection manager explicitly

lib/manticore/patches/client.rb
```diff
@@ -0,0 +1,21 @@
+module Manticore
+  class Client
```
yaauie, line 2: If possible, I would prefer we create a module to contain our patch, and to mix that module into `Manticore::Client` (instead of monkey-patching `Manticore::Client` directly). I would also prefer for that module to be scoped within the plugin, so if we have to chase it down from some other plugin's use of manticore, we can see the patch on the client's ancestor chain.

lib/logstash/outputs/elasticsearch/http_client/manticore_adapter.rb
```diff
@@ -1,4 +1,5 @@
 require 'manticore'
+require_relative '../../../../manticore/patches/client'
```
yaauie, line 2: Generally a `require_relative` with more than one `../` in it is a warning sign to me that the file we are requiring probably isn't as well-scoped as it should be.

yaauie (commented):
:+1: to getting an upstream fix merged already 🎉 .

I've added some commentary here, which may be obviated if we can get @cheald to put together a manticore release with [this and other fixes](https://github.com/cheald/manticore/compare/v0.7.0...master) 🤞🏼 .

cheald:
>  to getting an upstream fix merged already  .
> 
> I've added some commentary here, which may be obviated if we can get @cheald to put together a manticore release with [this and other fixes](https://github.com/cheald/manticore/compare/v0.7.0...master) 🤞🏼 .

I'll get a new release pushed out!

cheald:
https://rubygems.org/gems/manticore/versions/0.7.1-java - v0.7.1 released.

spec/es_spec_helper.rb
```diff
@@ -20,7 +20,9 @@ def get_host_port
   end
 
   def get_client
-    Elasticsearch::Client.new(:hosts => [get_host_port])
+    client = Elasticsearch::Client.new(:hosts => [get_host_port])
+    client.instance_variable_set("@verified", true) # bypass client side version checking
+    client
```
yaauie, line 25: While either approach is technically messing with Elasticsearch's private internals, I greatly prefer mocking methods over directly manipulating the ivar state of an object we don't own.

Both are liable to create problems down the track if the Elasticsearch gem decides to change their internals, but with a method mock we have both a greater sense of the _intent_ behind the patch and a somewhat smaller likelihood of creating problems (e.g., if the ivar takes on a different semantic meaning we risk stomping on it with something invalid; while if the method signature changes we risk not intercepting).

```suggestion
    Elasticsearch::Client.new(:hosts => [get_host_port]).tap do |client|
      allow(client).to receive(:verify_elasticsearch).and_return(true)
    end
```

kaisecheng: Thanks for the suggestions. The client team plans to release a version that relaxes the checking in 7.16. This hack is for a short-term transition. But I think it is no harm to take your suggestion too

yaauie (approved):
LGTM

PR #1038, opened 2021-09-06
Introduced 'user-agent' header

### LINKED ISSUE #1037: Add a customizable `user-agent` header to the HTTP request sent
When the plugin communicates with ES, the user-agent is manticore's, given that it the library used to communicate to Elasticsearch.

As, lets properly identify logstash by setting the user-agent when making HTTP requests to ES with the following format:

`logstash/7.14.1 (OS=Linux-5.4.0-84-generic-amd64; JVM=AdoptOpenJDK-11.0.11) logstash-output-elasticsearch/11.0.1`

andsel:
## Release notes

Feat: add `user-agent` header passed to the Elasticsearch HTTP connection

## What does this PR do?

Pass an option to the underlying `Manticore` library containing the `user-agent` header. That header contains:
- Logstah version
- plugin version, asking to the `Gem` library
- OS name, version and CPU architechture
- JVM vendor name and version

Example:
```
logstash/7.14.1 (OS=Linux-5.4.0-84-generic-amd64; JVM=AdoptOpenJDK-11.0.11) logstash-output-elasticsearch/11.0.1
```

## Checklist

- [ ] ~~I have made corresponding change to the default configuration files (and/or docker env variables)~~

## Author's Checklist

- [ ]

## Related issues

- Closes #1037

### DIFF
--- a/spec/unit/outputs/elasticsearch/http_client_spec.rb
+++ b/spec/unit/outputs/elasticsearch/http_client_spec.rb
@@ -348,7 +348,6 @@ describe LogStash::Outputs::ElasticSearch::HttpClient do
       it "should contains the default Logstash and plugin version" do
         adapter = client.build_adapter(client.options)
         adapter.perform_request(::LogStash::Util::SafeURI.new("http://localhost:#{webserver.port}"), :get, "/headers_check")
-        plugin_version = Gem.loaded_specs['logstash-output-elasticsearch'].version
 
         request = webserver.wait_receive_request
 
--- a/spec/unit/outputs/elasticsearch/http_client_spec.rb
+++ b/spec/unit/outputs/elasticsearch/http_client_spec.rb
@@ -1,6 +1,8 @@
 require_relative "../../../../spec/spec_helper"
 require "logstash/outputs/elasticsearch/http_client"
 require "cabin"
+require "webrick"
+require "java"
 
 describe LogStash::Outputs::ElasticSearch::HttpClient do
   let(:ssl) { nil }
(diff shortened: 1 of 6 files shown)


andsel:
The red CI on 
```
INTEGRATION=true
ELASTIC_STACK_VERSION=8.x
SNAPSHOT=true
```
seems not related to this PR, it happens also on `master` branch

roaksoax (changes requested):
Hi Andrea,

Thanks for adding the ability to set the user-agent. However, the intention of this is NOT to add a customizable user-agent. Rather, the intention of this is that Logstash doesn't inhering the user agent for the library (e.g. manticore) and sets its information always.

As such, we need to change this in such a way that logstash sets the suer-agent to something similar to:

Logstash/7.x.y (logstash-output-elasticsearch/a.b.c)

Ideally, this should also include the platform. Lets further discuss this.

spec/unit/outputs/elasticsearch/http_client_spec.rb
```diff
@@ -287,4 +289,70 @@
...
+      @first_request = nil
+
+      @t = java.lang.Thread.new(
+        proc do
+          begin
+            @server = WEBrick::HTTPServer.new :Port => @port, :DocumentRoot => ".",
+                     :Logger => Cabin::Channel.get, # silence WEBrick logging
+                     :StartCallback => Proc.new {
+                           queue.push("started")
+                         }
```
yaauie, line 310: It may, however, be better to let WEBrick bind to a random _available_ port instead of guessing one at random.

If we specify `:Port => 0`, then the resulting `WEBrick::HTTPServer` will bind to an _available_ port and its `config` will include and a `:Port` attribute containing the actual port that was bound.

(source: https://github.com/ruby/webrick/blob/v1.7.0/lib/webrick/server.rb#L112-L114)

```suggestion
            @server = WEBrick::HTTPServer.new :Port => 0, :DocumentRoot => ".",
                     :Logger => Cabin::Channel.get, # silence WEBrick logging
                     :StartCallback => Proc.new {
                           queue.push("started")
                         }
            @port = @server.config[:Port]
```

spec/unit/outputs/elasticsearch/http_client_spec.rb
```diff
@@ -287,4 +289,70 @@
...
+    end
+
+    context "when don't customize 'user_agent' option then the 'user-agent' header" do
+      it "should contains the default Logstash and plugin version" do
+        adapter = client.build_adapter(client.options)
+        adapter.perform_request(::LogStash::Util::SafeURI.new("http://localhost:#{webserver.port}"), :get, "/headers_check")
+
+        request = webserver.wait_receive_request
+
+        expect(request.header['user-agent']).to include(client.prepare_user_agent)
```
yaauie, line 354: nit: we don't have any specs on `Client#prepare_user_agent`, so ensuring that the user agent we received includes the prepared user agent doesn't sufficiently safeguard us (for example, if `Client#prepare_user_agent` returned an empty string, _any_ string would match).

I would prefer that we set more refined expectations here.

```suggestion
        transmitted_user_agent = request.header['user-agent']
        
        expect(transmitted_user_agent).to include("Logstash/#{LOGSTASH_VERSION}")
        expect(transmitted_user_agent).to match(%r{logstash-output-elasticsearch/[0-9.]+}")
```

yaauie (commented):
I've added a few nitpicks, but the implementation looks good.

spec/unit/outputs/elasticsearch/http_client_spec.rb
```diff
@@ -287,4 +289,70 @@
       end
     end
   end
+
+  class StoppableServer
+
+    attr_reader :port
+
+    def initialize()
+      @port = rand(65535-1024) + 1025
```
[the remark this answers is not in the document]
andsel: This was solved grabbing the port from the HTTP server. The HTTP server binds to the first free port and the testing code retrieve that port

roaksoax (approved):
From my perspective this LGTM! Not sure if @yaauie needs another review?

yaauie (approved):
LGTM 🌮

PR #1039, opened 2021-09-10
Adapt CI to run with security enabled when testing with Elasticsearch > 8

andsel:
## What does this PR do?
With Elasticsearch 8 the `xpack.security` is enabled by default, so updates the CI run with `SECURE_INTEGRATION`.

This PR contains some fixes to expose the value of `ELASTIC_STACK_VERSION` in the Ruby's `ENV` and fixes for `nil` checks when String versions are empty

## How to test
- export some environment variables:
```
export SECURE_INTEGRATION=true && \
export INTEGRATION=true && \
export ELASTIC_STACK_VERSION=8.x && \
export SNAPSHOT=true
```
- run the docker, after copied the scripts from https://github.com/logstash-plugins/.ci/tree/1.x  with : 
```
.ci/docker-setup.sh && .ci/docker-run.sh
```


.ci/docker-compose.override.yml
```diff
@@ -10,6 +10,7 @@ services:
     environment:
         - INTEGRATION=${INTEGRATION:-false}
         - SECURE_INTEGRATION=${SECURE_INTEGRATION:-false}
+        - ELASTIC_STACK_VERSION=$ELASTIC_STACK_VERSION
```
andsel, line 13: Note for reviewer: without this the `ENV['ELASTIC_STACK_VERSION']` in Ruby code sees an empty String

spec/es_spec_helper.rb
```diff
@@ -80,7 +80,7 @@ def self.es_version
   end
 
   def self.es_version_satisfies?(*requirement)
-    es_version = RSpec.configuration.filter[:es_version] || ENV['ES_VERSION'] || ENV['ELASTIC_STACK_VERSION']
+    es_version = nilify(RSpec.configuration.filter[:es_version]) || nilify(ENV['ES_VERSION']) || nilify(ENV['ELASTIC_STACK_VERSION'])
```
andsel, line 83: Note for reviewer: if the single part of the concatenated `||` operator is an empty string then the result is that empty string. So if in the `ENV` there is a `ES_VERSION` but it's empty and the same for the first particle then the `es_version` assume the `""` instead of the version of `ELASTIC_STACK_VERSION`

robbavey (approved):
lgtm

PR #1042, opened 2021-09-28
Fixed bad version bumpd 11.1.0 instead of 11.6.0

andsel:
Fix bad version bump (never published) happened with #1038


jsvd (approved):
LGTM

PR #1044, opened 2021-10-14
Fix referencing Gem classes from global lexical

andsel:
Some Gem classes was referenced from open Logstash lexical scope while it should refer from the global.


yaauie (approved):
LGTM ✨

PR #1046, opened 2021-10-25
Data streams require ECS

yaauie:
When data streams were initially added to this output plugin, we had planned on ECS being on-by-default in Logstash 8, so it made sense to also have data streams be the default target. We had temporarily punted the switch for ECS in Logstash to a later date, which revealed that running very simple configs in the LS8 betas cause runtime mapping errors (typically caused by upstream plugins, also not operating in ECS-mode, introducing ECS-conflicts implicitly). Even though we are back on track for the undrerlying ECS changes to go into Logstash 8, it will be helpful to codify this requirement.

In order to set our users up for success, this plugin should avoid routing to data streams unless it is being run in an ECS compatibility mode (preferably derived from the `pipeline.ecs_compatibility` setting, indicating that the upstream plugins that are the likely _source_ of the conflicts are operating in ECS mode).

To address this, this changeset does two things:
  - `data_stream => auto` resolves to _false_ when the plugin's `ecs_compatibility` is `disabled`.
  - `data_stream => true` now requires the effective `ecs_compatibility` to be non-`disabled` (**breaking? discuss**).

This change also makes necessary fixes to use ECS v8, and to specifying `ecs_compatibility` explicitly on Logstash 6.

TODO:
 - [x] code changes
 - ~[ ] docs changes~
 - [x] version bump
 - [x] changelog

### DIFF
--- a/lib/logstash/outputs/elasticsearch/data_stream_support.rb
+++ b/lib/logstash/outputs/elasticsearch/data_stream_support.rb
@@ -2,6 +2,15 @@ module LogStash module Outputs class ElasticSearch
   # DS specific behavior/configuration.
   module DataStreamSupport
 
+    # @api private
+    ENABLING_ECS_GUIDANCE = <<~END.tr("\n", " ")
+      Elasticsearch data streams require that events adhere to the Elastic Common Schema.
+      While `ecs_compatibility` can be set for this individual Elasticsearch output plugin, doing so will not fix schema conflicts caused by upstream plugins in your pipeline.
+      To avoid mapping conflicts, you will need to use ECS-compatible field names and datatypes throughout your pipeline.
+      Many plugins support an `ecs_compatibility` mode, and the `pipeline.ecs_compatibility` setting can be used to opt-in for all plugins in a pipeline.
+    END
+    private_constant :ENABLING_ECS_GUIDANCE
+
     def self.included(base)
       # Defines whether data will be indexed into an Elasticsearch data stream,
       # `data_stream_*` settings will only be used if this setting is enabled!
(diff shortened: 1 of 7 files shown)


lib/logstash/outputs/elasticsearch/data_stream_support.rb
```diff
@@ -93,6 +112,7 @@ def invalid_data_stream_params(params)
           true
         when 'data_stream'
           value.to_s == 'true'
+        when 'ecs_compatibility' then true # reqiuired for LS <= 6.x
```
robbavey, line 115: Nit: typo

lib/logstash/outputs/elasticsearch/data_stream_support.rb
```diff
@@ -123,18 +143,28 @@ def assert_es_version_supports_data_streams
     DATA_STREAMS_ENABLED_BY_DEFAULT_LS_VERSION = '8.0.0'
 
     # when data_stream => is either 'auto' or not set
-    def data_stream_default(data_stream_params, valid_data_stream_config)
-      ds_default = Gem::Version.create(LOGSTASH_VERSION) >= Gem::Version.create(DATA_STREAMS_ENABLED_BY_DEFAULT_LS_VERSION)
+    # @param data_stream_params [#any?]
+    # @param invalid_data_stream_config [#any?#inspect]
+    def data_stream_default(data_stream_params, invalid_data_stream_config)
```
robbavey, line 148: I'm wondering if it makes sense to log whether or not we're using data streams at `info` rather than `debug`, when using `auto` - it's new behaviour and it may be surprising to users. WDYT?

robbavey (approved):
LGTM, just a nit on a typo, and a question on logging

lib/logstash/outputs/elasticsearch/data_stream_support.rb
yaauie: Resolved in 311b1dd

lib/logstash/outputs/elasticsearch/data_stream_support.rb
yaauie: Resolved in 38627fd

yaauie:
flaky test looks unrelated. I've kicked off the sub-build again.

robbavey (approved):
LGTM

PR #1047, opened 2021-10-27
ecs: update v1 to ECS v1.12.1

yaauie:
Updates our vendoring definition so that when `ecs_compatibility => v1` is specified the ECS v1.12.1 templates are loaded.


kares (approved):
lgtm

PR #1049, opened 2021-11-02
ecs: pin v1 to upstream's v1.10.0 to avoid pulling breaking field types

yaauie:
Earlier today we released v11.2.2 of this plugin, which included broken templates for ECS v1 on Elasticsearch 6.x or 7.1.0-7.2.x, since they include the `flattened` field type introduced in Elasticsearch 7.3.

https://github.com/elastic/ecs/issues/1649

Because we do not have the ability to pin ECS minors to ES minors, this change pins the ECS templates to v1.10.0, the latest available that can be installed into 6.x or _any_ version of 7.x (including 7.1-7.2).

I will separately work to refine our ability to target specific versions of ECS based on major/minor of the ES we are connected to, but in the meantime this patch gets us as up-to-date as possible without shipping broken templates to our users running older Elasticsearches.


robbavey (approved):
lgtm

robbavey (changes requested):
Actually, I'm taking back the LGTM

The commit message and changelog message don't match the code. The commit message and changelog appear to be correct

Rakefile
```diff
@@ -1,7 +1,7 @@
 require "logstash/devutils/rake"
 
 ECS_VERSIONS = {
-    v1: 'v1.12.1'
+    v1: 'v1.10.1', # WARNING: v1.11 breaks 6.x (see: https://github.com/elastic/ecs/issues/1649)
```
robbavey, line 4: This should be `v1.10.0`, no?

yaauie: yes it should :facepal

robbavey (approved):
Tested locally, template installed correctly on Elasticsearch `6.8`.

Assuming green, LGTM

PR #1053, opened 2021-11-20
Doc: Fix note format

karenzone:
Current note formatting works, but isn't the best approach for simple notes. Fixing because formatting gets perpetuated in other places in docs.


karenzone (commented):
Built locally to validate fix.

PR #1055, opened 2021-11-29
Feat: add ssl_supported_protocols option

kares:
The new `ssl_supported_protocols` option default is aligned with what the Manticore/HttpClient defaults to 
-> whatever the SSL engine defaults to -> on up to date Java (>= 8) it's `TLSv1.3,TLSv1.2`

Having a configuration option (similar to ES' `xpack.security.http.ssl.supported_protocols: TLSv1.3,TLSv1.2`) allows for more flexibility.

This PR also establishes confidence that TLSv1.3 is working and supported on recent LS (JVM) versions.


lib/logstash/plugin_mixins/elasticsearch/api_configs.rb
```diff
@@ -66,6 +66,8 @@ module APIConfigs
         # Set the keystore password
         :keystore_password => { :validate => :password },
 
+        :ssl_supported_protocols => { :validate => ['TLSv1.1', 'TLSv1.2', 'TLSv1.3'], :default => [], :list => true },
```
kares, line 69: [ES](https://www.elastic.co/guide/en/elasticsearch/reference/current/security-settings.html) also (still) supports: `SSLv3`, `TLSv1` and `SSLv2Hello` but since these are disabled by default in an up-to-date Java I do not think we need to support them give even an old 6.0 cluster is expected to run on at least Java 8.

jsvd (commented):
I reran the travis tests and there's still one configuration failing: `SECURE_INTEGRATION=true INTEGRATION=true ELASTIC_STACK_VERSION=8.x SNAPSHOT=true LOG_LEVEL=info`

The logs don't seem to show anything relevant, may require some investigation to rule out as unrelated.

kares:
> I reran the travis tests and there's still one configuration failing: `SECURE_INTEGRATION=true INTEGRATION=true ELASTIC_STACK_VERSION=8.x SNAPSHOT=true LOG_LEVEL=info`

good catch, the issue was simply 8.x requiring auth credentials (for _refresh and friends), this slipped under the sleeve with the :red_circle: storm on CI. now all :green_circle:

docs/index.asciidoc
```diff
@@ -1004,6 +1005,23 @@ Option to validate the server's certificate. Disabling this severely compromises
 For more information on disabling certificate verification please read
 https://www.cs.utexas.edu/~shmat/shmat_ccs12.pdf
 
+[id="plugins-{type}s-{plugin}-ssl_supported_protocols"]
+===== `ssl_supported_protocols`
+
+  * Value type is <<string,string>>
+  * Allowed values are: `'TLSv1.1'`, `'TLSv1.2'`, `'TLSv1.3'`
+  * Default depends on the JDK being used. With up-to-date Logstash, the default is `['TLSv1.2', 'TLSv1.3']`.
```
robbavey, line 1013: I wonder if it makes sense to have a Logstash page equivalent to https://www.elastic.co/guide/en/elasticsearch/reference/current/jdk-tls-versions.html for Logstash for a single point of reference, given that this is copied across multiple plugins. 
cc @karenzone

lib/logstash/plugin_mixins/elasticsearch/api_configs.rb
robbavey: I'm slightly wary of removing something that is a valid option with Elasticsearch, maybe we should keep it but warn?

robbavey (commented):
Code looks good, I'd like to get feedback from @karenzone on the docs changes, and from @jsvd on whether we need to continue supporting TLS < v1.1 as it is a valid configuration for Elasticsearch, or that we can safely drop support

lib/logstash/plugin_mixins/elasticsearch/api_configs.rb
kares: we had a discussion with @jsvd on the topic and part of that outcome was that we start with `>= TLSv1.1` (for all plugins), in this case despite ES' support for legacy `SSLv3` and `TLSv1`.
existing plugin versions using TLS <= 1.1, on an up-to-date JDK, need tampering with the java.security files so if they're upgrading they either upgrade to secure TLS or we'll hear them complaining on the feature per plugin basis (the default - not setting the option is expected to work for them as before).

jsvd:
>  @jsvd on whether we need to continue supporting TLS < v1.1 

While it makes sense for ES to still allow the possibility of a client using tls1.0 because you many need to have a component sending to ES directly that can only do tls1.0 or lower, on the Logstash ES output we are essentially connecting only to ES, and all ES versions we support also support tls1.1+. We can certainly but a higher bar on the encryption parameters used between Elastic components.
If anything we should make tls1.1 deprecated and remove support afterwards, but we can tackle that more broadly in a separate effort.

docs/index.asciidoc
jsvd: It does @robbavey ! That refactor should happen separately from that PR, but I agree we should centralize this information. Pls open a docs issue so the idea isn't lost.

robbavey:
Thanks for the explanations @kares, @jsvd - makes sense. Code LGTM, leaving to @karenzone for final approval on doc

docs/index.asciidoc
robbavey: Issue created - https://github.com/elastic/logstash/issues/13962

karenzone (approved):
LGTM

Issue #1052, opened 2021-11-10
Add support for 'traces' data stream type


ultrr:
Currently, the plugin supports only following data stream types: 'logs', 'metrics', 'synthetics'. Other applications, for example APM Server also supports output to 'traces' type.

In our use case, we will like to use APM Server to place traces into ElasticSearch data stream via Logstash and at the moment this way to do so isn't possible.

kares:
Makes sense, the `type` naming already includes `'traces'` with [DS in Fleet Agent](https://www.elastic.co/guide/en/fleet/8.0/data-streams.html#data-streams-naming-scheme) as well ...

Issue #1056, opened 2021-12-01
ILM integration specs failing against 8.1-SNAPSHOT


kares:
```
logstash_1_7ef3b6d38fa0 | Failures:

logstash_1_7ef3b6d38fa0 | 

logstash_1_7ef3b6d38fa0 |   1) Elasticsearch has index lifecycle management enabled with ilm disabled behaves like an ILM disabled Logstash should not write the ILM settings into the template

logstash_1_7ef3b6d38fa0 |      Failure/Error: t = client.indices.get_index_template(name: name)

logstash_1_7ef3b6d38fa0 |      

logstash_1_7ef3b6d38fa0 |      Elasticsearch::Transport::Transport::Errors::NotFound:

logstash_1_7ef3b6d38fa0 |        [404] {"error":{"root_cause":[{"type":"resource_not_found_exception","reason":"index template matching [logstash] not found"}],"type":"resource_not_found_exception","reason":"index template matching [logstash] not found"},"status":404}

logstash_1_7ef3b6d38fa0 |      Shared Example Group: "an ILM disabled Logstash" called from ./spec/integration/outputs/ilm_spec.rb:524

logstash_1_7ef3b6d38fa0 |      # ./spec/es_spec_helper.rb:184:in `get_template'

logstash_1_7ef3b6d38fa0 |      # ./spec/integration/outputs/ilm_spec.rb:112:in `block in <main>'

logstash_1_7ef3b6d38fa0 | 

logstash_1_7ef3b6d38fa0 |   2) Elasticsearch has index lifecycle management enabled with ilm disabled behaves like an ILM disabled Logstash with a custom template name should not write the ILM settings into the template

logstash_1_7ef3b6d38fa0 |      Failure/Error: t = client.indices.get_index_template(name: name)

logstash_1_7ef3b6d38fa0 |      

logstash_1_7ef3b6d38fa0 |      Elasticsearch::Transport::Transport::Errors::NotFound:

logstash_1_7ef3b6d38fa0 |        [404] {"error":{"root_cause":[{"type":"resource_not_found_exception","reason":"index template matching [logstash_custom_template_name] not found"}],"type":"resource_not_found_exception","reason":"index template matching [logstash_custom_template_name] not found"},"status":404}

logstash_1_7ef3b6d38fa0 |      Shared Example Group: "an ILM disabled Logstash" called from ./spec/integration/outputs/ilm_spec.rb:524

[...]

PR #1057, opened 2021-12-01
Feat: add support for 'traces' data stream type

ultrr:
Adds support for 'traces' data stream type.

Related issue: https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/1052


kares:
:red_circle: is due https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/1056

PR #1059, opened 2021-12-21
Doc: Add v8 to ecs_compatibility options

karenzone:


karenzone:
@kares or @yaauie: Is this a change that we need to make?  Thanks for looking.

kares (commented):
:+1: we do want this to be documented.

the template itself will be v1 even for v8 up till 8.0 is actually released (ECS release 8.0), unsure here whether that is worth mentioning - in my opinion it's transient and not necessary.

kares:
> the template itself will be v1 even for v8 up till 8.0 is actually released (ECS release 8.0), unsure here whether that is worth mentioning - in my opinion it's transient and not necessary.

actually, since https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/1062 there's not a BEAT v8 template being shipped.
v1 template also got an update but this isn't mentioned (whether it's worth specifying might be up for discussion)

karenzone:
@kares 
>the template itself will be v1 even for v8 up till 8.0 is actually released (ECS release 8.0)

Is this accurate now that 8.0 has shipped?

kares:
> Is this accurate now that 8.0 has shipped?

That has changed recently with https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/1062 ... seems like we generated the v8 from the upstream (unreleased at that time) ECS v8 files.

kares (approved):

karenzone:
PENDING PUBLISHING: CI not green

Issue #1060, opened 2021-12-22
TASK: update ECS template vendoring scripts


yaauie:
We currently provide pre-release ECS v8 support for Elasticsearch 8 by transforming the ECS-provided "legacy" templates generated for Elasticsearch 7 into the Elasticsearch 8 format.

With https://github.com/elastic/ecs/pull/1700 ECS now only generates ECS 8 templates targeted at Elasticsearch 8, so we need to provide a reverse of that transformation to vendor templates targeted for Elasticsearch 7.

Additionally, with ECS becoming stack-aligned, the file structure of the generated artifacts in ECS 8 is changing
 - from `generated/elasticsearch/${elasticsearch_major}/template.json`, where `elasticsearch_major` is either `6` or `7`
 - to `generated/elasticsearch/legacy/template.json`, where "_legacy_" refers  the template not being the composable template format, but the artifact is targeted for same-version Elasticsearch (e.g., this file on the 8.3 branch will target Elasticsearch 8.3).

When complete, our vendoring scripts should:
  - [ ] vendor _frozen_ ECS 8.0.0 template for targeting Elasticsearch 7.x (avoid providing new 8.x features that will not be supported on Elasticsearch 7.x)
  - [ ] vendor tracked ECS 8.last template for targeting Elasticsearch 8.x 
  - [ ] vendor latest ECS 1.? for Elasticsearch 6, 7, and transformed-7-as-8

***

This change does _not_ block ECS from back-porting https://github.com/elastic/ecs/pull/1700 to `8.0` branch, as our current vendoring strategy for ECS v8 support relies on a pinned git commit, and our ECS v1 support relies on 1.x branches.

PR #1062, opened 2021-12-22
ECS Templates: stack alignment

yaauie:
Resolution of #1060.

With ECS 8.0, ECS is becoming Stack-aligned, and no longer generates template artifacts for other Elasticsearch majors. This is causing some paths to change in the upstream project from which we vendor our templates.

They continue to generate a monolith "legacy" template, for use with the Elasticsearch "legacy" template API (a.k.a., v1), but this plugin _automatically_ uses the v2 template endpoint when persisting templates to Elasticsearch 8.

Historically, this plugin has provided ECS v1 templates for (unreleased) Elasticsearch 8 by transforming them into the v2 format prior to vendoring. This changeset builds upon that work, renaming the relevant methods to be more descriptive and separating the concerns of transforming the template from those of defining which versions top vendor.

It also includes a number of build fixes.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,36 +1,57 @@
 require "logstash/devutils/rake"
 
-ECS_VERSIONS = {
-    v1: 'v1.10.0', # WARNING: v1.11 breaks 6.x (see: https://github.com/elastic/ecs/issues/1649)
+task :'vendor-ecs-schemata' do
+  # WARNING: v1.11 breaks 6.x (see: https://github.com/elastic/ecs/issues/1649)
+  download_ecs_schemata(:v1, elasticsearch_major: 6, ecs_release_tag: 'v1.10.0')
+  download_ecs_schemata(:v1, elasticsearch_major: 7, ecs_release_tag: 'v1.10.0')
+  download_ecs_schemata(:v1, elasticsearch_major: 8, ecs_release_tag: 'v1.10.0', generated_for: 7)
 
-    # PRERELEASE: 8.0@{2021-11-02T20:09:42Z}
-    # when pinning to released tag, remove BETA warning.
-    v8: 'b3dace43afb0dce743605cdd4cc067a3a6b8131c',
-}
+  # PRERELEASE: 8.0 branch
+  # when pinning to released tag, remove BETA warning.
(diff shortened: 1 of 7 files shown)


Rakefile
```diff
@@ -1,36 +1,52 @@
...
-    v1: 'v1.10.0', # WARNING: v1.11 breaks 6.x (see: https://github.com/elastic/ecs/issues/1649)
+task :'vendor-ecs-schemata' do
+  download_ecs_schemata(:v1, elasticsearch_major: 6, ecs_release_tag: 'v1.10.0') # WARNING: v1.11 breaks 6.x (see: https://github.com/elastic/ecs/issues/1649)
+  download_ecs_schemata(:v1, elasticsearch_major: 7, ecs_release_tag: 'v1.12.1')
+  download_ecs_schemata(:v1, elasticsearch_major: 8, ecs_release_tag: 'v1.12.1', generated_for: 7) { |template| transform_for_es8!(template) }
+
+  # PRERELEASE: 8.0 branch
+  # when pinning to released tag, remove BETA warning.
+  download_ecs_schemata(:v8, elasticsearch_major: 7, ecs_release_tag: '8.0') { |template| transform_for_es7!(template) if template.include?("template") }
+  download_ecs_schemata(:v8, elasticsearch_major: 8, ecs_release_tag: '8.0') { |template| transform_for_es8!(template) unless template.include?("template") }
```
yaauie, line 11: The in-block conditionals are meant to make this draft PR resilient to upstream changes in ECS, so that it will work both before and after they migrate ES7-style templates to ES8-style.

Rakefile
```diff
@@ -63,3 +79,20 @@ def transform_for_es8!(template)
 
   nil
 end
+
+# destructively transforms an ES8-style legacy index_template into an ES7-compatible legacy template.
+def transform_for_es7!(template)
```
yaauie, line 84: As long as ECS generates ES7-style legacy templates on the 8.0 branch, this will be orphan, unreachable code.

Rakefile
yaauie: Upstream ECS has clarified that they will be generating monolith templates only for the _legacy_ template API, even after stack alignment.

andsel (approved):
Nice,
LGTM!

PR #1063, opened 2021-12-27
Feat: upgrade manticore (http-client) library

kares:
Manticore 0.8.0 includes an upgrade of HttpClient to latest (4.5.13).
Also the `verify: :disable` logic has been changed in the library to skip all verification (including hostname checks).

Expected to address all remaining concerns from https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/433


kares:
:green_circle: expect [8.x SNAPSHOT](https://app.travis-ci.com/github/logstash-plugins/logstash-output-elasticsearch/jobs/553582916) due https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/1056

spec/integration/outputs/index_spec.rb
```diff
@@ -55,10 +56,18 @@
...
-  let(:index_url) {"#{es_url}/#{index}"}
-  let(:http_client_options) { {} }
-  let(:http_client) do
-    Manticore::Client.new(http_client_options)
+  let(:index_url) { "#{es_url}/#{index}" }
+
+  let(:curl_opts) { nil }
+
+  def curl_and_get_json_response(url, method: :get); require 'open3'
+    stdout, status = Open3.capture2("curl #{curl_opts} -X #{method.to_s.upcase} -k #{url}")
```
kares, line 64: we already rely on `curl` being present in our integration test setup

andsel: If `curl` is not installed `capture2` raise `Errno::ENOENT` . Could we catch that and have an assertion that cleanly exposes the problem of the missed tool. 
Being explicit would have trouble shooting a possible problem when tested on OSes that doesn't provide it.

andsel (approved):
Great work on the testing part, LGTM!

I've left a comment on how we could improve the check of the `curl` existence in the testing platform. Feel free to ignore, just a sugegstion

PR #1066, opened 2022-01-26
params: don't require explicit 'ssl=>true' when cloud_id or all-HTTPS hosts are present

### LINKED ISSUE #935: SSL Settings ignored if when using cloud_id unless explicitly setting `ssl => true`
When connecting via `cloud_id` and not explicitly setting `ssl => true`, the `params["hosts"]` is unpopulated causing us to fail to apply SSL-related settings:

https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/v10.4.1/lib/logstash/outputs/elasticsearch/http_client_builder.rb#L107-L108

Connections established in this way are still made over SSL, since the resulting host protocol is `https`, but we fail to apply the SSL-related settings.

### LINKED ISSUE #1065: Don't require ssl => true when using cloud_id
If an output is configured with cloud_id only:
```
output {
  elasticsearch {
    cloud_id => '....'
    api_key => 'id:key' 
  }
}
```

we get an error:

```
[2022-01-19T10:36:54,206][ERROR][logstash.javapipeline    ][main] Pipeline error {:pipeline_id=>"main", :exception=>#<LogStash::ConfigurationError: Using api_key authentication requires SSL/TLS secured communication using the `ssl => true` option>, 
```

This is an unnecessary error since all of the Elastic cloud deployments must have TLS enabled, so TLS should be enabled automatically when cloud.id is set.

yaauie:
Resolves #1065
Resolves #935 :tada:


spec/unit/outputs/elasticsearch_spec.rb
```diff
@@ -873,32 +878,64 @@
...
+      it "should raise a configuration error" do
+        expect { subject.register }.to raise_error LogStash::ConfigurationError, /requires SSL\/TLS/
       end
     end
 
+    context "when set" do
+      let(:options) { { "ssl" => true, "api_key" => ::LogStash::Util::Password.new(api_key) } }
+
+      it_behaves_like 'secure api-key authenticated client'
+    end
```
jsvd, line 937: Here's a suggestion to group the tests when api_key is set.

```suggestion
    context "when set" do
      let(:options) { { "api_key" => ::LogStash::Util::Password.new(api_key) } }

      context "with ssl => true" do
        let(:options) { super().merge("ssl" => true) }
        it_behaves_like 'secure api-key authenticated client'
      end

      context "with ssl => false" do
        let(:options) { super().merge("ssl" => "false") }

        let(:do_register) { false } # this is what we want to test, so we disable the before(:each) call
        it "should raise a configuration error" do
          expect { subject.register }.to raise_error LogStash::ConfigurationError, /requires SSL\/TLS/
        end
      end

      context "without ssl specified" do
        context "with an https host" do
          let(:options) { super().merge("hosts" => ["https://some.host.com"]) }
          it_behaves_like 'secure api-key authenticated client'
        end
        context "with an http host`" do
          let(:do_register) { false } # this is what we want to test, so we disable the before(:each) call
          let(:options) { { "hosts" => ["http://some.host.com"], "api_key" => api_key } }

          it "should raise a configuration error" do
            expect { subject.register }.to raise_error LogStash::ConfigurationError, /requires SSL\/TLS/
          end
        end
      end
    end
```

lib/logstash/plugin_mixins/elasticsearch/common.rb
```diff
@@ -64,6 +66,15 @@ def setup_hosts
       end
     end
 
+    def effectively_ssl?
+      return @ssl unless @ssl.nil?
+
+      hosts = Array(@hosts)
+      return nil if hosts.nil? || hosts.empty?
```
jsvd, line 73: `effectively_ssl?` being a predicate I'd rather only return true/false here:
```suggestion
      return false if hosts.nil? || hosts.empty?
```

lib/logstash/plugin_mixins/elasticsearch/common.rb
```diff
@@ -64,6 +66,15 @@ def setup_hosts
       end
     end
 
+    def effectively_ssl?
```
jsvd, line 69: Seems like we only need the `effectively_ssl?` predicate for this:

```
      if @api_key && @api_key.value && !effectively_ssl?
        raise(LogStash::ConfigurationError, "Using api_key authentication requires SSL/TLS secured communication using the `ssl => true` option")
      end
```
Which could be moved to the [HttpClientBuilder.setup_api_key](https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/882c9dda75819c8411f3b4291a22146179aa20c2/lib/logstash/outputs/elasticsearch/http_client_builder.rb#L154).
We also use it to set `params['ssl']` but that can also be set in the [HttpClientBuilder.setup_ssl](https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/882c9dda75819c8411f3b4291a22146179aa20c2/lib/logstash/outputs/elasticsearch/http_client_builder.rb#L109-L110).

yaauie: I'm not sure we should be raising `LogStash::ConfigurationError` from within `HttpClientBuilder` -- that seems like the wrong side of a boundary.

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+<<<<<<< HEAD
```
yaauie, line 1: snip

```suggestion
```

docs/index.asciidoc
```diff
@@ -988,11 +988,10 @@ do not use full URL here, only paths, e.g. "/sniff/_nodes/http"
...
 
-Enable SSL/TLS secured communication to Elasticsearch cluster. Leaving this
-unspecified will use whatever scheme is specified in the URLs listed in 'hosts'.
-If no explicit protocol is specified plain HTTP will be used. If SSL is
-explicitly disabled here the plugin will refuse to start if an HTTPS URL is
-given in 'hosts'
+Enable SSL/TLS secured communication to Elasticsearch cluster.
+Leaving this unspecified will use whatever scheme is specified in the URLs listed in <<plugins-{type}s-{plugin}-hosts>> or extracted from the <<plugins-{type}s-{plugin}-cloud_id>>.
+If no explicit protocol is specified plain HTTP will be used.
+If SSL is explicitly disabled here the plugin will refuse to start if an HTTPS URL is given in 'hosts'
```
jsvd, line 994: > If SSL is explicitly disabled here the plugin will refuse to start if an HTTPS URL is given in 'hosts'

This is only true if the api_key is set, otherwise a configuration such as `output { elasticsearch { hosts => ['https://localhost:9300'] ssl => false }}` will not raise an error at startup.

PR #1067, opened 2022-02-16
Adds a new `dlq_custom_codes` option to customize DLQ codes

### LINKED ISSUE #697: Make 403 response configurable for going into DLQ
When a 403 response is encountered, this is retried forever. This blocks the pipeline from continuing to process (https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/140). This is good in the event of a major misconfiguration since the entire pipeline becomes blocked. This can be bad in the case where just a couple messages are routed to indices that might be closed out due to automatic curation. The behavior of a 403 messages should be configurable, so it could go into the DLQ.

```
[2017-10-24T15:31:30,232][INFO ][logstash.outputs.elasticsearch] retrying failed action with response code: 403 ({"type"=>"index_closed_exception", "reason"=>"closed", "index_uuid"=>"Lnvl85mrRtSOGDzZnb86gw", "index"=>"test_dlq"})
```

Discussed with @jordansissel, and agree this should not be the default behavior - since putting in the DLQ only extends the amount of time before the pipeline is stalled (due to a full DLQ). However, it should be a choice an end user can configure for desired behavior.

andsel:
## Release notes
Adds `dlq_custom_codes` option to customize the list of HTTP codes that are considered eligible for DLQ.

## What does this PR do?
Introduced a new configuration option, named `dlq_custom_codes` which accepts a list of HTTP error codes.
The list is to be considered as an addition to the already existing codes that the plugin uses (DLQ_CODES, DOC_SUCCESS_CODES, DOC_CONFLICT_CODE defined in module `Common`). If the list has an intersection with those error codes then it's reported as a configuration error.

This is a feature requested in #697, in some cases some error codes returned by Elasticsearch should be handled as DLQ messaged and not retried for ever. For example writing to a read only index report a 403, the index status would never change automatically so it's best to move these events to DLQ for a later reprocessing.
Under normal conditions the user shouldn't customize this list, but it's an helper in some circumstances.

## Checklist

- ~~[ ] I have commented my code, particularly in hard-to-understand areas~~

## Author's Checklist

- [x] test a version of the plugin with a real ES

- launch an ES instance, for example `docker-compose -f es_and_kb.yml up` 
[es_and_kb.txt](https://github.com/logstash-plugins/logstash-output-elasticsearch/files/8081186/es_and_kb.txt)
- from http://localhost:5601 Kibana UI (credential: elastic, changeme) create and index and put into read only:
```
PUT /test_index
PUT test_index/_settings 
{ 
  "index.blocks.read_only" : true 
}
```
- configure a pipeline to write into:
```
input {
	stdin {}
}

output {
	elasticsearch {
		index => "test_index"
		hosts => "http://localhost:9200"
		user => "elastic"
		password => "changeme"
		dlq_custom_codes => [403]
	}
}	
```
- enable DLQ in `config/logstash.yml` : 
```
dead_letter_queue.enable: true
```
- start the pipeline and generate an input and stop
- check that the string you typed is present in DLQ 
```
strings data/dead_letter_queue/main/1.log
```

## Related issues

- Fixes #697
- Related https://github.com/elastic/logstash/issues/10493

## Use cases

HTTP error codes used to decide if an event index reply has to be moved into the DLQ are 400 and 404. These codes are hard coded, this PR aims the user to customize this list extending other than 400 and 404.

### DIFF
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -197,7 +197,7 @@ re-indexed to Elasticsearch. If the DLQ is not enabled, and a mapping error
 happens, the problem is logged as a warning, and the event is dropped. See
 <<dead-letter-queues>> for more information about processing events in the DLQ.
 The list of error codes accepted for DLQ could be customized with <<plugins-{type}s-{plugin}-dlq_custom_codes>>
-but should be done only in motivated cases.
+but should be used only in motivated cases.
 
 [id="plugins-{type}s-{plugin}-ilm"]
 ==== Index Lifecycle Management
@@ -527,7 +527,8 @@ Currently, only `logs`, `metrics`, `synthetics` and `traces` are supported.
 * Value type is <<number,number>>
 * Default value is `[]`.
 
-List extra HTTP's error codes that are considered valid to move the events into the dead letter queue.
+List HTTP's error codes that are considered valid to move the events into the dead letter queue.
+This list is an addition to the ordinary error codes considered for this feature, 400 and 404.
 It's considered a configuration error to re-use the same predefined codes for success, DLQ or conflict.
 The option accepts a list of natural numbers corresponding to HTTP errors codes.
 
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -527,7 +527,7 @@ Currently, only `logs`, `metrics`, `synthetics` and `traces` are supported.
 * Value type is <<number,number>>
 * Default value is `[]`.
 
-List HTTP's error codes that are considered valid to move the events into the dead letter queue.
+List single-action error codes from Elasticsearch's Bulk API that are considered valid to move the events into the dead letter queue.
 This list is an addition to the ordinary error codes considered for this feature, 400 and 404.
 It's considered a configuration error to re-use the same predefined codes for success, DLQ or conflict.
 The option accepts a list of natural numbers corresponding to HTTP errors codes.
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -196,6 +196,8 @@ processed at a later time. Often times, the offending field can be removed and
 re-indexed to Elasticsearch. If the DLQ is not enabled, and a mapping error
 happens, the problem is logged as a warning, and the event is dropped. See
 <<dead-letter-queues>> for more information about processing events in the DLQ.
+The list of error codes accepted for DLQ could be customized with <<plugins-{type}s-{plugin}-dlq_custom_codes>>
+but should be done only in motivated cases.
 
 [id="plugins-{type}s-{plugin}-ilm"]
 ==== Index Lifecycle Management
@@ -317,6 +319,7 @@ This plugin supports the following configuration options plus the
 | <<plugins-{type}s-{plugin}-data_stream_namespace>> |<<string,string>>|No
 | <<plugins-{type}s-{plugin}-data_stream_sync_fields>> |<<boolean,boolean>>|No
 | <<plugins-{type}s-{plugin}-data_stream_type>> |<<string,string>>|No
+| <<plugins-{type}s-{plugin}-dlq_custom_codes>> |<<number,number>>|No
 | <<plugins-{type}s-{plugin}-doc_as_upsert>> |<<boolean,boolean>>|No
 | <<plugins-{type}s-{plugin}-document_id>> |<<string,string>>|No
 | <<plugins-{type}s-{plugin}-document_type>> |<<string,string>>|No
@@ -517,6 +520,16 @@ overwritten with a warning.
 
 The data stream type used to construct the data stream at index time.
 Currently, only `logs`, `metrics`, `synthetics` and `traces` are supported.
+
+[id="plugins-{type}s-{plugin}-dlq_custom_codes"]
+===== `dlq_custom_codes`
+
+* Value type is <<number,number>>
+* Default value is `[]`.
+
+List extra HTTP's error codes that are considered valid to move the events into the dead letter queue.
+It's considered a configuration error to re-use the same predefined codes for success, DLQ or conflict.
+The option accepts a list of natural numbers corresponding to HTTP errors codes.
 
 [id="plugins-{type}s-{plugin}-doc_as_upsert"]
 ===== `doc_as_upsert`
--- a/lib/logstash/outputs/elasticsearch.rb
+++ b/lib/logstash/outputs/elasticsearch.rb
@@ -299,10 +299,11 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
     # To support BWC, we check if DLQ exists in core (< 5.4). If it doesn't, we use nil to resort to previous behavior.
     @dlq_writer = dlq_enabled? ? execution_context.dlq_writer : nil
 
+    @dlq_codes = DOC_DLQ_CODES.to_set
+
     if dlq_enabled?
       check_dlq_custom_codes
-      @dlq_codes = dlq_custom_codes.to_set
-      @dlq_codes.merge(DOC_DLQ_CODES)
+      @dlq_codes.merge(dlq_custom_codes)
     else
       raise LogStash::ConfigurationError, "DLQ feature (dlq_custom_codes) is configured while DLQ is not enabled" unless dlq_custom_codes.empty?
     end
--- a/lib/logstash/outputs/elasticsearch.rb
+++ b/lib/logstash/outputs/elasticsearch.rb
@@ -255,6 +255,9 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
   # ILM policy to use, if undefined the default policy will be used.
   config :ilm_policy, :validate => :string, :default => DEFAULT_POLICY
 
+  # List extra HTTP's error codes that are considered valid to move the events into the dead letter queue.
+  # It's considered a configuration error to re-use the same predefined codes for success, DLQ or conflict.
+  # The option accepts a list of natural numbers corresponding to HTTP errors codes.
   config :dlq_custom_codes, :validate => :number, :list => true, :default => []
 
   attr_reader :client
--- a/lib/logstash/outputs/elasticsearch.rb
+++ b/lib/logstash/outputs/elasticsearch.rb
@@ -300,6 +300,8 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
 
     if dlq_enabled?
       check_dlq_custom_codes
+    else
+      raise LogStash::ConfigurationError, "DLQ feature (dlq_custom_codes) is configured while DLQ is not enabled" unless dlq_custom_codes.empty?
     end
 
     if data_stream_config?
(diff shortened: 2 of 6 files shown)


docs/index.asciidoc
```diff
@@ -518,6 +521,17 @@ overwritten with a warning.
 The data stream type used to construct the data stream at index time.
 Currently, only `logs`, `metrics`, `synthetics` and `traces` are supported.
 
+[id="plugins-{type}s-{plugin}-dlq_custom_codes"]
+===== `dlq_custom_codes`
+
+* Value type is <<number,number>>
+* Default value is `[]`.
+
+List HTTP's error codes that are considered valid to move the events into the dead letter queue.
```
yaauie, line 530: While the per-action response codes _mirror_ http status codes, we only handle retries and DLQ routing for failed actions from bulk requests whose HTTP status code is 2xx. This has caused a fair bit of confusion in the past, so I believe we need to be careful to call these out as action response codes and avoid any explicit association with HTTP.

lib/logstash/plugin_mixins/elasticsearch/common.rb
```diff
@@ -268,7 +268,7 @@ def submit(actions)
           @document_level_metrics.increment(:non_retryable_failures)
           @logger.warn "Failed action", status: status, action: action, response: response if log_failure_type?(error)
           next
-        elsif DOC_DLQ_CODES.include?(status)
+        elsif DOC_DLQ_CODES.include?(status) || dlq_custom_codes.include?(status)
```
yaauie, line 271: I would prefer using a `Set` (loaded once at plugin registration) that contained the union of these two, loaded into an instance variable, instead of adding a per-event overhead.

lib/logstash/outputs/elasticsearch.rb
```diff
@@ -531,4 +540,15 @@ def check_action_validity
...
+  def check_dlq_custom_codes
+    intersection = dlq_custom_codes & DOC_DLQ_CODES
+    raise LogStash::ConfigurationError, "dlq_custom_codes contains error codes already used: #{intersection}" unless intersection.empty?
+
+    intersection = dlq_custom_codes & DOC_SUCCESS_CODES
+    raise LogStash::ConfigurationError, "dlq_custom_codes contains error codes already defined as success: #{intersection}" unless intersection.empty?
+
+    intersection = dlq_custom_codes & [DOC_CONFLICT_CODE]
+    raise LogStash::ConfigurationError, "dlq_custom_codes contains error codes already defined as conflict: #{intersection}" unless intersection.empty?
+  end
```
kaisecheng, line 553: besides success, conflict and dlq codes, there is a `failure_type_logging_whitelist`. Do you think it should be checked here?

lib/logstash/outputs/elasticsearch.rb
```diff
@@ -531,4 +540,15 @@ def check_action_validity
 
     raise LogStash::ConfigurationError, "Action '#{@action}' is invalid! Pick one of #{valid_actions} or use a sprintf style statement"
   end
+
+  def check_dlq_custom_codes
+    intersection = dlq_custom_codes & DOC_DLQ_CODES
+    raise LogStash::ConfigurationError, "dlq_custom_codes contains error codes already used: #{intersection}" unless intersection.empty?
+
+    intersection = dlq_custom_codes & DOC_SUCCESS_CODES
+    raise LogStash::ConfigurationError, "dlq_custom_codes contains error codes already defined as success: #{intersection}" unless intersection.empty?
```
kaisecheng, line 549: suggest a rephase for you to consider "#{intersection} are success codes which cannot be redefined in dlq_custom_codes"

docs/index.asciidoc
andsel: I think is sufficient to rephrase it as:
```
List response or single action error codes that...
```
WDYT?

lib/logstash/outputs/elasticsearch.rb
```diff
@@ -255,6 +255,11 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
   # ILM policy to use, if undefined the default policy will be used.
   config :ilm_policy, :validate => :string, :default => DEFAULT_POLICY
 
+  # List extra HTTP's error codes that are considered valid to move the events into the dead letter queue.
+  # It's considered a configuration error to re-use the same predefined codes for success, DLQ or conflict.
+  # The option accepts a list of natural numbers corresponding to HTTP errors codes.
+  config :dlq_custom_codes, :validate => :number, :list => true, :default => []
```
kaisecheng, line 261: about the naming, I understand there are fixed codes 400 and 404, so name the changeable part as custom codes. As a user, I think `dlq_codes` is more approachable and easy to understand.

docs/index.asciidoc
yaauie: This specific line here, but we also have references to HTTP codes throughout.

```suggestion
List single-action error codes from Elasticsearch's Bulk API that are considered valid to move the events into the dead letter queue.
```

docs/index.asciidoc
```diff
@@ -518,6 +521,17 @@ overwritten with a warning.
...
+This list is an addition to the ordinary error codes considered for this feature, 400 and 404.
+It's considered a configuration error to re-use the same predefined codes for success, DLQ or conflict.
```
yaauie, line 532: It looks like we already handle `200`/`201` as explicit success, `400`/`404` as DLQ, and `409` as conflict, and are declaring it a configuration error to re-use any of these.

If a user can't use this to intercept `409` conflicts, or 404's attempting to update records that don't exist, what error codes do we expect? Only authorization-related ones like `403`? Others? Do we have real-world use-cases driving this feature?

lib/logstash/outputs/elasticsearch.rb
yaauie: I favor `dlq_custom_codes`, because I would assume `dlq_codes` to require passing a complete list to _override_ the list of codes we naturally route to DLQ, whereas the `custom` bit implies that it is in _addition_ to the natural routing.

lib/logstash/outputs/elasticsearch.rb
```diff
@@ -293,6 +298,10 @@ def register
     # To support BWC, we check if DLQ exists in core (< 5.4). If it doesn't, we use nil to resort to previous behavior.
     @dlq_writer = dlq_enabled? ? execution_context.dlq_writer : nil
 
+    if dlq_enabled?
```
yaauie, line 301: if dlq is not enabled for a pipeline, and a user explicitly tells us how to route things to the DLQ, I would consider that a configuration error too.

docs/index.asciidoc
andsel: This PR aim to solve request #697 which explicitly state that `403` error code could be eventually added as a DLQ trigger code.
While intersection with existing DLQ codes (`400` and `404`) could be seen as not necessarily a conflict, because both trigger the same feature, using error codes already configured as success (`200`/`201`) document conflict error (`409`) generates confusion in the precedence order when the [code process responses](https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/1cab1ad2a41b28ac569cb77809b48ed6b66d2b39/lib/logstash/plugin_mixins/elasticsearch/common.rb#L264-L280).

Suppose that also `409` is customized as a `dlq_custom_code` what should take precedence? Processing it as a failed action or as DLQ event?

To keep the things simple I imagined to keep out of the set the codes that already has a meaning in the plugin.

lib/logstash/outputs/elasticsearch.rb
andsel: The intention was to inform the user that's not the full overriding list but an addition list to the default behavior.

lib/logstash/outputs/elasticsearch.rb
andsel: `failure_type_logging_whitelist` is used to log or not errors during the execution of the queries, while the method `#check_dlq_custom_codes` is used during registration to report configuration error. I think are different aspect, one concern about a configuration problem that should be logged or raised independently of `failure_type_logging_whitelist` which is a setting used to limit the number of logs.

lib/logstash/outputs/elasticsearch.rb
andsel: Fixed by 5a81f0a328dbdcaa683cf7fd059677f3791a0dbe

lib/logstash/outputs/elasticsearch.rb
andsel: nice point!

andsel: Fixed by 8e98c89a7d618698e7d2871001d39bec44857fa5

lib/logstash/plugin_mixins/elasticsearch/common.rb
andsel: Fixed by ec57f24c27fdb595b9c78e13a2c67ce754fd7926

lib/logstash/plugin_mixins/elasticsearch/common.rb
```diff
@@ -268,7 +268,7 @@ def submit(actions)
           @document_level_metrics.increment(:non_retryable_failures)
           @logger.warn "Failed action", status: status, action: action, response: response if log_failure_type?(error)
           next
-        elsif DOC_DLQ_CODES.include?(status)
+        elsif @dlq_codes.include?(status)
```
kaisecheng, line 271: It is causing red CI in integration test version 6.x

kaisecheng (commented):
Once the CI is fixed, I can approve.

lib/logstash/plugin_mixins/elasticsearch/common.rb
andsel: the error is:
```
undefined method `include?' for nil:NilClass\nDid you mean?  include_class", :exception=>NoMethodError
```
So `@dlq_codes` is `nil` fo r `6.x`. I think it's related to the way `dlq_enabled?` works on LS `6.x`

andsel: My guess was that the way `dlq_enabled?` operates on classes to determine if DLQ is active involved some fields that wasn't present on `6.8` but [ExecutionContextExt.dlqWriter](https://github.com/elastic/logstash/blob/50b43ebbaf35a042b9d1848118dcffde264dfbc9/logstash-core/src/main/java/org/logstash/execution/ExecutionContextExt.java#L39) exists in `6.8`

docs/index.asciidoc
roaksoax: @yaauie seems that this needs your feedback?

yaauie: Yes. I can unblock this PR.

We can consider a later feature to enable this explicit custom-code routing to also intercept `400`/`404`/`409` that are already routed by the implementation, and if and when we do so we have a path to do it in a non-breaking way.

yaauie (approved):
LGTM 👍🏼

yaauie:
I've merged in upstream and resolved conflicting version bump in changelog/gemspec, and will merge/release on successful CI.

PR #1080, opened 2022-08-08
Validate the action before sending the bulk request to ES.

### LINKED ISSUE #1079: Bulk request without checking supported action causing ES 400.
## Short Description
Plugin receives `400 Validation Failed: 1: no requests added` error when sending any random action.

## Expectation
Plugin should validate the action before sending bulk request to ES.

**Logstash information**:

Please include the following information:

1. Logstash version (e.g. `bin/logstash --version`)
2. Logstash installation source (e.g. built from source, with a package manager: DEB/RPM, expanded from tar or zip archive, docker)
3. How is Logstash being run (e.g. as a service/service manager: systemd, upstart, etc. Via command line, docker/kubernetes)
4. How was the Logstash Plugin installed

**JVM** (e.g. `java -version`):

If the affected version of Logstash is 7.9 (or earlier), or if it is NOT using the bundled JDK or using the 'no-jdk' version in 7.10 (or higher), please provide the following information:

1. JVM version (`java -version`)
2. JVM installation source (e.g. from the Operating System's package manager, from source, etc).
3. Value of the `JAVA_HOME` environment variable if set.

**OS version** (`uname -a` if on a Unix-like system):

**Steps to reproduce**:

Please include a *minimal* but *complete* recreation of the problem,
including (e.g.) pipeline definition(s), settings, locale, etc.  The easier
you make for us to reproduce it, the more likely that somebody will take the
time to look at it.

 1. Setup any Logstash version (tested 7.15.1 & 8.3.0) and read json file
 ```
 input {
    file {
        path => "/path/to/files/*.json"
        sincedb_path => "/since/db/path/sincedb"
        start_position => "beginning"
        codec => "json"
    }
}
 ```
 3. `logstash-output-elasticsearch` `11.6.0` version
 4. Set any value in file for action (ex: `partialupdate`) except `index`, `delete`, `create` & `update` and save
 ```
{
    "event":
    {
        "name": "TEST",
        "type": "TEST",
        "action": "partialupdate",
        "source": "template",
        "version": 1
    },
    "eventData":
    {
        "hello": "world"
    }
}
 ```

Plugin receives `400 Validation Failed: 1: no requests added` error.

```
[2022-08-08T11:58:50,017][ERROR][logstash.outputs.elasticsearch][main][1] Encountered a retryable error (will retry with exponential backoff) {:code=>400, :url=>"https://{ES_HOST}/_bulk", :content_length=>882, :body=>"{\"error\":{\"root_cause\":[{\"type\":\"action_request_validation_exception\",\"reason\":\"Validation Failed: 1: no requests added;\"}],\"type\":\"action_request_validation_exception\",\"reason\":\"Validation Failed: 1: no requests added;\"},\"status\":400}"}
```

mashhurs:
# This PR is under process
We have common logics with #1084, rebasing and leveraging the changes. 

## What this PR does?
Plugin receives `400 Validation Failed: 1: no requests added` exception when requesting unsupported bulk action.
This PR introduces a validation check on action before sending bulk request to ES.

Closes #1079 

### Manual testing
  - Used configuration to read from JSON file and send to ES. File contains an action field which is not in [`index`, `create`, `update`, `delete`]
```
   input {
      file {
          path => "/path/to/files/*.json"
          start_position => "beginning"
          codec => "json"
      }
   }
   output {
     elasticsearch {
       ....
       action => "%{[@metadata][action]}"
     }
   }
```
 - Result
```
[2022-09-09T11:59:44,810][WARN ][logstash.outputs.elasticsearch][main][ca10f590b6ab3366bc38f754293adbd2499c9d97aedd7933d11542cd26cd056d] Could not index event to Elasticsearch because its action is not supported. Action: ["partialupdate", {:_id=>"5412e1fe-7822-4440-86b6-5b508ab99229", :_index=>"logstash-test-output", :routing=>nil}, {"eventData"=>....}]
```

### Unit testing
  - Run the unit test
  `rspec spec/unit/outputs/elasticsearch_spec.rb`

### Logs
```
[2022-09-21T09:55:33,170][WARN ][logstash.outputs.elasticsearch][main][02e543d9daaf1e2af9190a6aef2cb91aa012fabaf7756093f487d85b2b9224de] Badly formatted index, after interpolation still contains placeholder: [%{[index_name]}], {"action"=>"index", "host"=>{"name"=>"dhcp-128-189-70-158"}, "name"=>"TEST", "source"=>"template", "@timestamp"=>2022-09-21T16:55:33.045630Z, "type"=>"TEST", "version"=>1, "@version"=>"1", "log"=>{"file"=>{"path"=>"/Users/mashhur/Dev/elastic/SDH/test2.json"}}, "event"=>{"original"=>"{\"name\":\"TEST\",\"type\":\"TEST\",\"action\":\"index\",\"source\":\"template\",\"version\":1}"}}

[2022-09-21T09:55:33,171][WARN ][logstash.outputs.elasticsearch][main][02e543d9daaf1e2af9190a6aef2cb91aa012fabaf7756093f487d85b2b9224de] Elasticsearch doesn't support [partialupdate] action, {"action"=>"partialupdate", "host"=>{"name"=>"dhcp-128-189-70-158"}, "source"=>"template", "type"=>"TEST", "index_name"=>"logstash-output", "@version"=>"1", "log"=>{"file"=>{"path"=>"/Users/mashhur/Dev/elastic/SDH/test2.json"}}, "event"=>{"original"=>"{\"name\":\"TEST\",\"type\":\"TEST\",\"action\":\"partialupdate\",\"source\":\"template\",\"version\":1,\"index_name\":\"logstash-output\"}"}, "name"=>"TEST", "@timestamp"=>2022-09-21T16:55:33.052311Z, "version"=>1}


[2022-09-21T09:55:33,171][WARN ][logstash.outputs.elasticsearch][main][02e543d9daaf1e2af9190a6aef2cb91aa012fabaf7756093f487d85b2b9224de] Can't map the event, needs to configure the pipeline settings according to the input. The event sent to DLQ: Badly formatted index, after interpolation still contains placeholder: [%{[index_name]}], {"action"=>"index", "host"=>{"name"=>"dhcp-128-189-70-158."}, "name"=>"TEST", "source"=>"template", "@timestamp"=>2022-09-21T16:55:33.045630Z, "type"=>"TEST", "version"=>1, "@version"=>"1", "log"=>{"file"=>{"path"=>"/Users/mashhur/Dev/elastic/SDH/test2.json"}}, "event"=>{"original"=>"{\"name\":\"TEST\",\"type\":\"TEST\",\"action\":\"index\",\"source\":\"template\",\"version\":1}"}}


[2022-09-21T09:55:33,171][WARN ][logstash.outputs.elasticsearch][main][02e543d9daaf1e2af9190a6aef2cb91aa012fabaf7756093f487d85b2b9224de] Can't map the event, needs to configure the pipeline settings according to the input. The event sent to DLQ: Elasticsearch doesn't support [partialupdate] action, {"action"=>"partialupdate", "host"=>{"name"=>"dhcp-128-189-70-158"}, "source"=>"template", "type"=>"TEST", "index_name"=>"logstash-output", "@version"=>"1", "log"=>{"file"=>{"path"=>"/Users/mashhur/Dev/elastic/SDH/test2.json"}}, "event"=>{"original"=>"{\"name\":\"TEST\",\"type\":\"TEST\",\"action\":\"partialupdate\",\"source\":\"template\",\"version\":1,\"index_name\":\"logstash-output\"}"}, "name"=>"TEST", "@timestamp"=>2022-09-21T16:55:33.052311Z, "version"=>1}
```

### DIFF
--- a/lib/logstash/outputs/elasticsearch.rb
+++ b/lib/logstash/outputs/elasticsearch.rb
@@ -367,36 +367,37 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
     wait_for_successful_connection if @after_successful_connection_done
     events_mapped = safe_interpolation_map_events(events)
     retrying_submit(events_mapped.successful_events)
-    unless events_mapped.failed_events.empty?
-      @logger.error("Can't map some events, needs to be handled by DLQ #{events_mapped.failed_events}")
-      send_failed_resolutions_to_dlq(events_mapped.failed_events)
+    unless events_mapped.event_mapping_errors.empty?
+      handle_event_mapping_errors(events_mapped.event_mapping_errors)
     end
   end
 
-  # @param: Arrays of EventActionTuple
+  # @param: Arrays of EventActionTuple with error messages
   private
-  def send_failed_resolutions_to_dlq(failed_action_tuples)
-    failed_action_tuples.each do |action|
-      handle_dlq_status(action, "warn", "Could not resolve dynamic index")
+  def handle_event_mapping_errors(event_mapping_errors)
+    event_mapping_errors.each do |action, message|
+      detail_message = message + ", " + action.to_s
+      handle_dlq_status(action.event, :warn, detail_message)
+      @document_level_metrics.increment(:non_retryable_failures) unless @dlq_writer
     end
   end
 
-  MapEventsResult = Struct.new(:successful_events, :failed_events)
+  MapEventsResult = Struct.new(:successful_events, :event_mapping_errors)
 
   private
   def safe_interpolation_map_events(events)
     successful_events = [] # list of LogStash::Outputs::ElasticSearch::EventActionTuple
-    failed_events = [] # list of LogStash::Event
+    event_mapping_errors = [] # list of LogStash::Outputs::ElasticSearch::EventActionTuple with error messages
     events.each do |event|
-       begin
+      begin
         successful_events << @event_mapper.call(event)
-       rescue IndexInterpolationError, e
-         action = event.sprintf(@action || 'index')
-         event_action_tuple = EventActionTuple.new(action, [], event)
-         failed_events << event_action_tuple
+      rescue EventMappingError => ie
+        action = event.sprintf(@action || 'index')
+        event_action_tuple = EventActionTuple.new(action, [], event)
+        event_mapping_errors << [event_action_tuple, ie.message]
        end
     end
-    MapEventsResult.new(successful_events, failed_events)
+    MapEventsResult.new(successful_events, event_mapping_errors)
   end
 
   public
@@ -449,6 +450,7 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
     end
 
     action = event.sprintf(@action || 'index')
+    raise UnsupportedActionError, action unless VALID_HTTP_ACTIONS.include?(action)
 
     if action == 'update'
       params[:_upsert] = LogStash::Json.load(event.sprintf(@upsert)) if @upsert != ""
@@ -476,13 +478,21 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
 
   end
 
-  class IndexInterpolationError < ArgumentError
-    attr_reader :bad_formatted_index
+  class EventMappingError < ArgumentError
+    def initialize(msg = nil)
+      super
+    end
+  end
 
+  class IndexInterpolationError < EventMappingError
     def initialize(bad_formatted_index)
       super("Badly formatted index, after interpolation still contains placeholder: [#{bad_formatted_index}]")
+    end
+  end
 
-      @bad_formatted_index = bad_formatted_index
+  class UnsupportedActionError < EventMappingError
+    def initialize(bad_action)
+      super("Elasticsearch doesn't support [#{bad_action}] action")
     end
   end
 
--- a/lib/logstash/outputs/elasticsearch.rb
+++ b/lib/logstash/outputs/elasticsearch.rb
@@ -372,29 +372,33 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
     end
   end
 
-  # @param: Arrays of EventActionTuple with error messages
+  # @param: Arrays of FailedEventMapping
   private
   def handle_event_mapping_errors(event_mapping_errors)
-    event_mapping_errors.each do |action, message|
-      detail_message = message + ", " + action.to_s
-      handle_dlq_status(action.event, :warn, detail_message)
-      @document_level_metrics.increment(:non_retryable_failures) unless @dlq_writer
+    # if DQL is enabled, log the events to provide issue insights to users.
+    if @dlq_writer
+      @logger.warn("Events could not be indexed and routing to DLQ, count: #{event_mapping_errors.size}")
     end
+
+    event_mapping_errors.each do |event_mapping_error|
+      detailed_message = "#{event_mapping_error.message}; event: `#{event_mapping_error.event.to_hash_with_metadata}`"
+      handle_dlq_status(event_mapping_error.event, :warn, detailed_message)
+    end
+    @document_level_metrics.increment(:non_retryable_failures, event_mapping_errors.size)
   end
 
   MapEventsResult = Struct.new(:successful_events, :event_mapping_errors)
+  FailedEventMapping = Struct.new(:event, :message)
 
   private
   def safe_interpolation_map_events(events)
     successful_events = [] # list of LogStash::Outputs::ElasticSearch::EventActionTuple
-    event_mapping_errors = [] # list of LogStash::Outputs::ElasticSearch::EventActionTuple with error messages
+    event_mapping_errors = [] # list of FailedEventMapping
     events.each do |event|
       begin
         successful_events << @event_mapper.call(event)
       rescue EventMappingError => ie
-        action = event.sprintf(@action || 'index')
-        event_action_tuple = EventActionTuple.new(action, [], event)
-        event_mapping_errors << [event_action_tuple, ie.message]
+        event_mapping_errors << FailedEventMapping.new(event, ie.message)
        end
     end
     MapEventsResult.new(successful_events, event_mapping_errors)
--- a/lib/logstash/outputs/elasticsearch/http_client.rb
+++ b/lib/logstash/outputs/elasticsearch/http_client.rb
@@ -151,7 +151,7 @@ module LogStash; module Outputs; class ElasticSearch;
                    :payload_size => stream_writer.pos,
                    :content_length => body_stream.size,
                    :batch_offset => (actions.size - batch_actions.size))
-      bulk_responses << bulk_send(body_stream, batch_actions)
+      bulk_responses << bulk_send(body_stream, batch_actions) if body_stream.size > 0
 
       body_stream.close if !http_compression
       join_bulk_responses(bulk_responses)
--- a/lib/logstash/outputs/elasticsearch/http_client.rb
+++ b/lib/logstash/outputs/elasticsearch/http_client.rb
@@ -103,22 +103,24 @@ module LogStash; module Outputs; class ElasticSearch;
 
       return if actions.empty?
 
-      bulk_responses = []
+      unsupported_action_indexes = []
+      index = 0
+
       bulk_actions = actions.collect do |action, args, source|
         args, source = update_action_builder(args, source) if action == 'update'
 
+        # do not filter out the action, we have a logic in common.rb -> submit() to match action index with response index
         unless ElasticSearch::VALID_HTTP_ACTIONS.include?(action)
-          logger.warn("Bulk request rejected because of unsupported action: ", :action => action)
-          bulk_responses << emulate_batch_error_response([[{action => args}, source].compact], 400, "unsupported action")
-          next nil
+          unsupported_action_indexes << index
         end
+        index = index + 1
 
         if source && action != 'delete'
           next [ { action => args }, source ]
         else
           next { action => args }
         end
-      end.compact
+      end
 
       body_stream = StringIO.new
       if http_compression
@@ -127,8 +129,15 @@ module LogStash; module Outputs; class ElasticSearch;
       else
         stream_writer = body_stream
       end
+
+      bulk_responses = []
       batch_actions = []
       bulk_actions.each_with_index do |action, index|
+        if unsupported_action_indexes.include?(index)
+          bulk_responses << emulate_batch_error_response([action], 422, "unsupported action")
+          next
+        end
+
         as_json = action.is_a?(Array) ?
                     action.map {|line| LogStash::Json.dump(line)}.join("\n") :
                     LogStash::Json.dump(action)
@@ -148,7 +157,9 @@ module LogStash; module Outputs; class ElasticSearch;
         stream_writer.write(as_json)
         batch_actions << action
       end
+
       stream_writer.close if http_compression
+
       if body_stream.size > 0
         logger.debug("Sending final bulk request for batch.",
                      :action_count => batch_actions.size,
--- a/lib/logstash/outputs/elasticsearch/http_client.rb
+++ b/lib/logstash/outputs/elasticsearch/http_client.rb
@@ -103,17 +103,8 @@ module LogStash; module Outputs; class ElasticSearch;
 
       return if actions.empty?
 
-      unsupported_action_indexes = []
-      index = 0
-
       bulk_actions = actions.collect do |action, args, source|
         args, source = update_action_builder(args, source) if action == 'update'
-
-        # do not filter out the action, we have a logic in common.rb -> submit() to match action index with response index
-        unless ElasticSearch::VALID_HTTP_ACTIONS.include?(action)
-          unsupported_action_indexes << index
-        end
-        index = index + 1
 
         if source && action != 'delete'
           next [ { action => args }, source ]
@@ -133,11 +124,6 @@ module LogStash; module Outputs; class ElasticSearch;
       bulk_responses = []
       batch_actions = []
       bulk_actions.each_with_index do |action, index|
-        if unsupported_action_indexes.include?(index)
-          bulk_responses << emulate_batch_error_response([action], 422, "unsupported action")
-          next
-        end
-
         as_json = action.is_a?(Array) ?
                     action.map {|line| LogStash::Json.dump(line)}.join("\n") :
                     LogStash::Json.dump(action)
@@ -160,14 +146,13 @@ module LogStash; module Outputs; class ElasticSearch;
 
       stream_writer.close if http_compression
 
-      if body_stream.size > 0
-        logger.debug("Sending final bulk request for batch.",
-                     :action_count => batch_actions.size,
-                     :payload_size => stream_writer.pos,
-                     :content_length => body_stream.size,
-                     :batch_offset => (actions.size - batch_actions.size))
-        bulk_responses << bulk_send(body_stream, batch_actions)
-      end
+      logger.debug("Sending final bulk request for batch.",
+                   :action_count => batch_actions.size,
+                   :payload_size => stream_writer.pos,
+                   :content_length => body_stream.size,
+                   :batch_offset => (actions.size - batch_actions.size))
+      bulk_responses << bulk_send(body_stream, batch_actions)
+
       body_stream.close if !http_compression
       join_bulk_responses(bulk_responses)
     end
--- a/lib/logstash/outputs/elasticsearch/http_client.rb
+++ b/lib/logstash/outputs/elasticsearch/http_client.rb
@@ -103,15 +103,22 @@ module LogStash; module Outputs; class ElasticSearch;
 
       return if actions.empty?
 
+      bulk_responses = []
       bulk_actions = actions.collect do |action, args, source|
         args, source = update_action_builder(args, source) if action == 'update'
+
+        unless ElasticSearch::VALID_HTTP_ACTIONS.include?(action)
+          logger.warn("Bulk request rejected because of unsupported action: ", :action => action)
+          bulk_responses << emulate_batch_error_response([[{action => args}, source].compact], 400, "unsupported action")
+          next nil
+        end
 
         if source && action != 'delete'
           next [ { action => args }, source ]
         else
           next { action => args }
         end
-      end
+      end.compact
 
       body_stream = StringIO.new
       if http_compression
@@ -120,7 +127,6 @@ module LogStash; module Outputs; class ElasticSearch;
       else
         stream_writer = body_stream
       end
-      bulk_responses = []
       batch_actions = []
       bulk_actions.each_with_index do |action, index|
         as_json = action.is_a?(Array) ?
@@ -143,12 +149,14 @@ module LogStash; module Outputs; class ElasticSearch;
         batch_actions << action
       end
       stream_writer.close if http_compression
-      logger.debug("Sending final bulk request for batch.",
-                   :action_count => batch_actions.size,
-                   :payload_size => stream_writer.pos,
-                   :content_length => body_stream.size,
-                   :batch_offset => (actions.size - batch_actions.size))
-      bulk_responses << bulk_send(body_stream, batch_actions) if body_stream.size > 0
+      if body_stream.size > 0
+        logger.debug("Sending final bulk request for batch.",
+                     :action_count => batch_actions.size,
+                     :payload_size => stream_writer.pos,
+                     :content_length => body_stream.size,
+                     :batch_offset => (actions.size - batch_actions.size))
+        bulk_responses << bulk_send(body_stream, batch_actions)
+      end
       body_stream.close if !http_compression
       join_bulk_responses(bulk_responses)
     end
--- a/lib/logstash/plugin_mixins/elasticsearch/common.rb
+++ b/lib/logstash/plugin_mixins/elasticsearch/common.rb
@@ -165,7 +165,7 @@ module LogStash; module PluginMixins; module ElasticSearch
     end
 
     def retrying_submit(actions)
-      # filter out unsupported ES actions
+      # reject unsupported ES actions and keep only valid actions to send to ES
       submit_actions = reject_unsupported_actions(actions)
       sleep_interval = @retry_initial_interval
 
--- a/lib/logstash/plugin_mixins/elasticsearch/common.rb
+++ b/lib/logstash/plugin_mixins/elasticsearch/common.rb
@@ -166,7 +166,7 @@ module LogStash; module PluginMixins; module ElasticSearch
 
     def retrying_submit(actions)
       # filter out unsupported ES actions
-      submit_actions = filter_unsupported_actions(actions)
+      submit_actions = reject_unsupported_actions(actions)
       sleep_interval = @retry_initial_interval
 
       # Initially we submit the full list of valid actions
@@ -192,24 +192,23 @@ module LogStash; module PluginMixins; module ElasticSearch
       end
     end
 
-    # we shouldn't send unsupported actions to ES
-    # method filters out unsupported actions by warning, writes event to DQL if enabled
+    # Do not send unsupported actions to ES
+    # method rejects unsupported actions before sending to ES, warns and writes event to DQL if enabled
     # @returns valid actions
-    def filter_unsupported_actions(actions)
-      return if actions.nil? || actions.size < 1
+    def reject_unsupported_actions(actions)
       supported_actions, unsupported_actions = actions.partition { |action, _, _| LogStash::Outputs::ElasticSearch::VALID_HTTP_ACTIONS.include?(action) }
-      if !unsupported_actions.nil? && unsupported_actions.size > 0
-        handle_dlq_or_drop(unsupported_actions)
-        @logger.warn("Number of requests filtered out before sending to Elasticsearch because of unsupported action, ", size: actions.size)
-      end
+      dlq_or_drop_unsupported_actions(unsupported_actions) if unsupported_actions.any?
       supported_actions
     end
 
-    def handle_dlq_or_drop(unsupported_actions)
-      return if @dlq_writer.nil?
-
+    def dlq_or_drop_unsupported_actions(unsupported_actions)
       unsupported_actions.each do |action|
-        @dlq_writer.write(action.event, "Unsupported event.")
+        if @dlq_writer
+          @dlq_writer.write(action.event, "Unsupported action.")
+        else
+          @document_level_metrics.increment(:non_retryable_failures)
+          @logger.warn("Could not index event to Elasticsearch because its action is not supported.", action: action)
+        end
       end
     end
 
--- a/lib/logstash/plugin_mixins/elasticsearch/common.rb
+++ b/lib/logstash/plugin_mixins/elasticsearch/common.rb
@@ -198,7 +198,7 @@ module LogStash; module PluginMixins; module ElasticSearch
     def filter_unsupported_actions(actions)
       return if actions.nil? || actions.size < 1
       supported_actions, unsupported_actions = actions.partition { |action, _, _| LogStash::Outputs::ElasticSearch::VALID_HTTP_ACTIONS.include?(action) }
-      unless unsupported_actions.nil? && unsupported_actions.size == 0
+      if !unsupported_actions.nil? && unsupported_actions.size > 0
         handle_dlq_or_drop(unsupported_actions)
         @logger.warn("Number of requests filtered out before sending to Elasticsearch because of unsupported action, ", size: actions.size)
       end
--- a/lib/logstash/plugin_mixins/elasticsearch/common.rb
+++ b/lib/logstash/plugin_mixins/elasticsearch/common.rb
@@ -165,11 +165,11 @@ module LogStash; module PluginMixins; module ElasticSearch
     end
 
     def retrying_submit(actions)
-      # Initially we submit the full list of actions
-      # filter out invalid ES unsupported actions
+      # filter out unsupported ES actions
       submit_actions = filter_unsupported_actions(actions)
       sleep_interval = @retry_initial_interval
 
+      # Initially we submit the full list of valid actions
       while submit_actions && submit_actions.size > 0
 
         # We retry with whatever is didn't succeed
@@ -196,7 +196,7 @@ module LogStash; module PluginMixins; module ElasticSearch
     # method filters out unsupported actions by warning, writes event to DQL if enabled
     # @returns valid actions
     def filter_unsupported_actions(actions)
-      return if actions.empty? || actions.size < 1
+      return if actions.nil? || actions.size < 1
       grouped_actions = actions.group_by { |action, arg, source | LogStash::Outputs::ElasticSearch::VALID_HTTP_ACTIONS.include?(action) }
       unless grouped_actions[false].nil?
         @logger.warn("Number of requests filtered out before sending to Elasticsearch because of unsupported action, ", size: grouped_actions[false].size)
--- a/lib/logstash/plugin_mixins/elasticsearch/common.rb
+++ b/lib/logstash/plugin_mixins/elasticsearch/common.rb
@@ -192,22 +192,23 @@ module LogStash; module PluginMixins; module ElasticSearch
       end
     end
 
-    # we shouldn't unsupported actions to ES
+    # we shouldn't send unsupported actions to ES
     # method filters out unsupported actions by warning, writes event to DQL if enabled
     # @returns valid actions
     def filter_unsupported_actions(actions)
       return if actions.nil? || actions.size < 1
-      grouped_actions = actions.group_by { |action, arg, source | LogStash::Outputs::ElasticSearch::VALID_HTTP_ACTIONS.include?(action) }
-      unless grouped_actions[false].nil?
-        @logger.warn("Number of requests filtered out before sending to Elasticsearch because of unsupported action, ", size: grouped_actions[false].size)
-        write_unsupported_events_to_dql(grouped_actions[false])
+      supported_actions, unsupported_actions = actions.partition { |action, _, _| LogStash::Outputs::ElasticSearch::VALID_HTTP_ACTIONS.include?(action) }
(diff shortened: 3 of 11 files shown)


spec/unit/outputs/elasticsearch/http_client_spec.rb
```diff
@@ -266,6 +266,18 @@
...
+          let(:actions) { [
+            ["index", {:_id=>nil, :_index=>"logstash"}, {"message"=> "hello"}],
+            ["unsupported_action", {:_id=>nil, :_index=>"logstash"}, {"message"=> "world!"}],
+          ]}
+          it "executes one bulk_send operation" do
+            allow(subject).to receive(:join_bulk_responses)
+            expect(subject).to receive(:bulk_send).once
+            s = subject.send(:bulk, actions)
+          end
+        end
```
yaauie, line 280: we should also make assertions about what we expect back. The emulation of an error response and zipping back together with the actual bulk requests that are sent are currently untested.

lib/logstash/outputs/elasticsearch/http_client.rb
```diff
@@ -103,15 +103,22 @@ def bulk(actions)
...
+          next nil
+        end
+
         if source && action != 'delete'
           next [ { action => args }, source ]
         else
           next { action => args }
         end
-      end
+      end.compact
```
yaauie, line 121: I worry that this message combined with our emulated error in the batch's combined-bulks response will cause users to assume that _Elasticsearch_ is rejecting the event, and that a batch with many such events will _flood_ the logs with identical messages. We want it to be clear that the event is being _excluded_ from being sent, and to rely on our per-event handling of errors to include event-specific context in the logs.

Speaking of which, we have an ordering problem. The return value of this method must contain its per-event responses in the [same order as it received the actions](https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/1cab1ad2a41b28ac569cb77809b48ed6b66d2b39/lib/logstash/plugin_mixins/elasticsearch/common.rb#L251-L257) so that it can correctly map which events to retry.

By picking off the invalid actions here we are preventing the order from being maintained.


***

The following is what I was considering _before_ realizing that we have an ordering problem, and is _NOT_ a valid suggestion but perhaps worth including for context.

We may want to pre-partition our actions into those that are valid and those that are not valid, and emit a single message that XX events were not sent due to unsupported actions. We can rely on the downstream handling of the value we return (including the emulated error from ES) to log the individual events that are getting retried/dropped, which includes additional context about the event itself.

If I were to tackle this, I would probably break this collect block out into a private method like `build_action_tuple`, then do something like:

~~~ ruby
      bulk_actions, unsupported_actions = actions.partition do |action, _, _|
        ElasticSearch::VALID_HTTP_ACTIONS.include?(action)
      end.collect do |partition_actions|
        partition_actions.collect do |action, args, source|
          build_action_tuple(action, args, source)
        end
      end
      if unsupported_actions.any?
        logger.warn("events were excluded from batch because they had an unsupported action", event_count: unsupported_actions.size)
        bulk_responses << emulate_batch_error_response(unsupported_actions, 400, "unsupported action")
      end
~~~

yaauie (changes requested):
We have an ordering problem, since the per-event responses must be emitted in the same order as the actions we received.

lib/logstash/outputs/elasticsearch/http_client.rb
```diff
@@ -103,9 +103,18 @@ def bulk(actions)
...
+      index = 0
+
       bulk_actions = actions.collect do |action, args, source|
         args, source = update_action_builder(args, source) if action == 'update'
 
+        # do not filter out the action, we have a logic in common.rb -> submit() to match action index with response index
+        unless ElasticSearch::VALID_HTTP_ACTIONS.include?(action)
+          unsupported_action_indexes << index
+        end
+        index = index + 1
```
mashhurs, line 116: somehow don't like this plain incremental approach but when I tried `actions.collect.with_index`, return data (`bulk_actions`) messed up. Any way to simplify?

lib/logstash/plugin_mixins/elasticsearch/common.rb
```diff
@@ -280,6 +284,10 @@ def submit(actions)
         end
       end
 
+      if unsupported_actions.size > 0
+        @logger.warn("Bulk requests rejected before sending to Elasticsearch because of unsupported action, ", size: unsupported_actions.size)
+      end
```
mashhurs, line 289: in case checking: should we increment any metrics for this case?

spec/unit/outputs/elasticsearch/http_client_spec.rb
mashhurs: please take a look at changes.

lib/logstash/outputs/elasticsearch/http_client.rb
mashhurs: Thanks @yaauie , 
It is not really visible [the logic where maps the requested action with response](https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/1cab1ad2a41b28ac569cb77809b48ed6b66d2b39/lib/logstash/plugin_mixins/elasticsearch/common.rb#L257) and I think for the long term we should avoid (maybe Hashing mechanism would be better?) to mitigate such risks.
I have updated the PR with new change and please give your feedbacks when you will have a time. Thanks.

lib/logstash/plugin_mixins/elasticsearch/common.rb
```diff
@@ -192,6 +192,26 @@ def retrying_submit(actions)
       end
     end
 
+    # we shouldn't unsupported actions to ES
+    # method filters out unsupported actions by warning, writes event to DQL if enabled
+    # @returns valid actions
+    def filter_unsupported_actions(actions)
```
mashhurs, line 198: _(unit tests will fail, confirming if this approach makes sense)_

@yaauie, as we took offline, time complexity becomes little extended but we overcome the order mismatch case. If you let me know your feedback I will test more and update the unit tests. Thanks.

lib/logstash/plugin_mixins/elasticsearch/common.rb
```diff
@@ -192,6 +192,26 @@ def retrying_submit(actions)
...
+      grouped_actions = actions.group_by { |action, arg, source | LogStash::Outputs::ElasticSearch::VALID_HTTP_ACTIONS.include?(action) }
+      unless grouped_actions[false].nil?
+        @logger.warn("Number of requests filtered out before sending to Elasticsearch because of unsupported action, ", size: grouped_actions[false].size)
+        write_unsupported_events_to_dql(grouped_actions[false])
+      end
+      grouped_actions[true].nil? ? [] : grouped_actions[true]
+    end
+
+    def write_unsupported_events_to_dql(actions)
+      return if @dlq_writer.nil?
```
mashhurs, line 209: `pipelines.yml` -> `dead_letter_queue.enable: true` didn't help. Is there any way I can make `common.rb` -> `dlq_enabled?` true?

lib/logstash/plugin_mixins/elasticsearch/common.rb
```diff
@@ -192,6 +192,26 @@ def retrying_submit(actions)
...
+    # @returns valid actions
+    def filter_unsupported_actions(actions)
+      return if actions.empty? || actions.size < 1
+      grouped_actions = actions.group_by { |action, arg, source | LogStash::Outputs::ElasticSearch::VALID_HTTP_ACTIONS.include?(action) }
+      unless grouped_actions[false].nil?
+        @logger.warn("Number of requests filtered out before sending to Elasticsearch because of unsupported action, ", size: grouped_actions[false].size)
+        write_unsupported_events_to_dql(grouped_actions[false])
+      end
+      grouped_actions[true].nil? ? [] : grouped_actions[true]
+    end
```
yaauie, line 206: I would recommend using `Enumerable#partition` instead of `Enumerable#group_by` since the result can be bound to two local variables that can be handled directly, instead of digging into a hash repeatedly.

```suggestion
    def filter_unsupported_actions(actions)
      supported_actions, unsupported_actions = actions.partition { |action, _, _| VALID_HTTP_ACTIONS.include?(action) }
      
      handle_dlq_or_drop(unsupported_actions)
      
      supported_actions
    end
    
    def handle_dlq_or_drop(unsupported_actions)
      return if unsupported_actions.nil? || unsupported_actions.empty?
      
      # ... implement routing to dlq or drop.
      # logging in data-loss situations should mirror our existing dropped events logging in detail.
    end
```

lib/logstash/plugin_mixins/elasticsearch/common.rb
yaauie: I would rely on `handle_dlq_status` implementation here, whose `action` argument is an `EventActionTuple` (an entry in the `actions` enumerator passed to `retrying_submit` that extends `Array` so it "splats" to an `action`,`args`,`source` but also has an `EventActionTuple#event` method to get the `LogStash::Event`)

From there, we can stub out the `handle_dlq_status` and validate it receives the args we expect.

lib/logstash/outputs/elasticsearch/http_client.rb
```diff
@@ -142,13 +143,16 @@ def bulk(actions)
...
+
       stream_writer.close if http_compression
+
       logger.debug("Sending final bulk request for batch.",
                    :action_count => batch_actions.size,
                    :payload_size => stream_writer.pos,
                    :content_length => body_stream.size,
                    :batch_offset => (actions.size - batch_actions.size))
-      bulk_responses << bulk_send(body_stream, batch_actions) if body_stream.size > 0
+      bulk_responses << bulk_send(body_stream, batch_actions)
```
yaauie: IIRC this `if body_stream.size > 0` was a safeguard against sending zero-entry bulks. I haven't validated whether those are still possible.

mashhurs: Yeah I accidentally removed it, put back in my previous commit.

lib/logstash/plugin_mixins/elasticsearch/common.rb
mashhurs: Well, I would agree to utilize `handle_dlq_status` if it _does not expect response information_. We are not sending any request to ES/3rd-party or cannot forcefully imitate the response (ex: with 400/0 code). It also logs the event and throws if DQL is not enabled where we are introducing log flood.

lib/logstash/plugin_mixins/elasticsearch/common.rb
mashhurs: Not sure why `VALID_HTTP_ACTIONS` is not injected, getting following error:
```
[ERROR][logstash.javapipeline    ][main] Pipeline worker error, the pipeline will be stopped {:pipeline_id=>"main",
 :error=>"(NameError) uninitialized constant LogStash::PluginMixins::ElasticSearch::Common::VALID_HTTP_ACTIONS", :exception=>Java::OrgJrubyExceptions::NameError,
```

lib/logstash/plugin_mixins/elasticsearch/common.rb
yaauie: We make a very serious effort in this plugin and others to not lose data, but when we need to drop data, at the very least we ensure that we are not silent about it. The log flood caused by routing through `handle_dlq_status ` is both intentional and necessary.

lib/logstash/plugin_mixins/elasticsearch/common.rb
```diff
@@ -192,6 +192,27 @@ def retrying_submit(actions)
...
+    # we shouldn't send unsupported actions to ES
+    # method filters out unsupported actions by warning, writes event to DQL if enabled
+    # @returns valid actions
+    def filter_unsupported_actions(actions)
+      return if actions.nil? || actions.size < 1
+      supported_actions, unsupported_actions = actions.partition { |action, _, _| LogStash::Outputs::ElasticSearch::VALID_HTTP_ACTIONS.include?(action) }
+      if !unsupported_actions.nil? && unsupported_actions.size > 0
+        handle_dlq_or_drop(unsupported_actions)
+        @logger.warn("Number of requests filtered out before sending to Elasticsearch because of unsupported action, ", size: actions.size)
+      end
```
yaauie, line 204: - `Enumerable#partition` will _always_ return 2 arrays, so we don't need a nil-check here.
 - because we should be logging the actual events that are being dropped (or not logging when DLQ is enabled and can capture them), the additional warning here isn't particularly meaningful or actionable.

```suggestion
      if unsupported_actions.any?
        handle_dlq_or_drop(unsupported_actions)
        @logger.warn("Number of requests filtered out before sending to Elasticsearch because of unsupported action, ", size: unsupported_actions.size)
      end
```

Or, since we log the skipped actions in the DLQ-or-drop handling:

```suggestion
      handle_dlq_or_drop(unsupported_actions) if unsupported_actions.any?
```

And if the `handle_or_drop(unsupported_actions)` method is implemented as-suggested by using `Enumerable#each`, we don't need to do the `Enumerable#any?` check here either.

```suggestion
      handle_dlq_or_drop(unsupported_actions)
```

lib/logstash/plugin_mixins/elasticsearch/common.rb
```diff
@@ -192,6 +192,27 @@ def retrying_submit(actions)
...
+      supported_actions
+    end
+
+    def handle_dlq_or_drop(unsupported_actions)
```
yaauie, line 208: nitpick: method name doesn't indicate that this has semantics specific to unsupported actions, so we risk accidental use in later development

spec/unit/outputs/elasticsearch/http_client_spec.rb
```diff
@@ -266,6 +266,22 @@
             end
           end
         end
+
+        context "with partial unsupported action" do
```
yaauie, line 270: testing scenarios:
 - several valid actions with one trailing invalid action
 - one leading invalid action followed by several valid actions
 - batch of several invalid actions and no valid actions
 - batch of intermixed valid and invalid actions
 - batch of exactly one action that is invalid

lib/logstash/plugin_mixins/elasticsearch/common.rb
yaauie: method naming nitpick:

The word `filter` is ambiguous; lean on ruby/smalltalk's `select`/`reject` to be more clear about the intended effect. 

NOTE: suggestion requires rename propagation

```suggestion
    def reject_unsupported_actions(actions)
```

lib/logstash/plugin_mixins/elasticsearch/common.rb
```diff
@@ -192,6 +192,27 @@ def retrying_submit(actions)
       end
     end
 
+    # we shouldn't send unsupported actions to ES
+    # method filters out unsupported actions by warning, writes event to DQL if enabled
+    # @returns valid actions
+    def filter_unsupported_actions(actions)
+      return if actions.nil? || actions.size < 1
```
yaauie, line 199: It is impossible for `retrying_submit` to receive `nil` for its `actions`, and therefore it is impossible for this method to receive `nil`. We can count on having an `Enumerable` (result of an upstream `Enumerable#map`) and can avoid the excess nil-checks.


Since we are relying on the return value being the actions that we should submit, this guard clause changes a given `Array` that is _empty_ into a `nil`. This is effectively handled somewhat accidentally in a guard clause in `retrying_submit`, but we can avoid the complexity because `Enumerable#partition` on an empty input array is _very_ fast so we really don't need to guard at all.

```suggestion
```

lib/logstash/plugin_mixins/elasticsearch/common.rb
```diff
@@ -165,11 +165,11 @@ def discover_cluster_uuid
...
     def retrying_submit(actions)
-      # Initially we submit the full list of actions
-      submit_actions = actions
-
+      # filter out unsupported ES actions
+      submit_actions = filter_unsupported_actions(actions)
       sleep_interval = @retry_initial_interval
 
+      # Initially we submit the full list of valid actions
       while submit_actions && submit_actions.size > 0
```
yaauie, line 173: out-of-scope note: `submit_actions` is either an array that was given to this method, or the return value of `submit` which is also guaranteed to be an array (as long as our new `filter_unsupported_actions` does not return `nil`, which it can do as of cf9a834).

If we avoid returning `nil` from `filter_unsupported_actions`, this guard-clause (and several others in this method) can be simplified. _Feel free to exclude this suggestion from the scope of this PR, as it is merely meant to provide for context for other suggestions._

```suggestion
      while submit_actions.any?
```

lib/logstash/plugin_mixins/elasticsearch/common.rb
```diff
@@ -192,6 +192,27 @@ def retrying_submit(actions)
...
+      supported_actions
+    end
+
+    def handle_dlq_or_drop(unsupported_actions)
+      return if @dlq_writer.nil?
+
+      unsupported_actions.each do |action|
+        @dlq_writer.write(action.event, "Unsupported event.")
+      end
+    end
```
yaauie, line 214: Method-naming aside, since:
 - we _should_ log each and every instance of data-loss,
 - we _should_ increment the `non_retryable_failures` metric, AND
 - we will _always_ get an array,

this method becomes something like:

```suggestion
    def handle_dlq_or_drop(unsupported_actions)
      unsupported_actions.each do |action|
        if @dlq_writer
          @dlq_writer.write(action.event, "Unsupported action.")
        else
          @document_level_metrics.increment(:non_retryable_failures)
          @logger.warn("Could not index event to Elasticsearch because its action is not supported.", action: action)
        end
      end
    end
```

spec/unit/outputs/elasticsearch/http_client_spec.rb
```diff
@@ -266,6 +266,22 @@
...
+          let(:actions) { [
+            ["index", {:_id=>nil, :_index=>"logstash"}, {"message"=> "hello"}],
+            ["unsupported_action", {:_id=>nil, :_index=>"logstash"}, {"message"=> "world!"}],
+          ]}
+          it "executes one bulk_send operation" do
+            allow(subject).to receive(:join_bulk_responses)
+            expect(subject).to receive(:bulk_send).once do |body_stream, batch_actions|
```
yaauie, line 277: we can likely specify receiving `submit` instead of `bulk_send`, because `bulk_send` has its own optimizations around payload size that are out of scope.

lib/logstash/plugin_mixins/elasticsearch/common.rb
mashhurs: Thanks for the awesome feedback!
I usually avoid giving a naming includes `reject` (which sounds to me a sense of ending the process) but from ruby's `select` point of view, it seems `reject` is appropriative.

mashhurs: In Java, lambda expressions use `filter` mostly.

lib/logstash/plugin_mixins/elasticsearch/common.rb
mashhurs: I agree with the parts for `unsupported actions` but in `retrying_submit` method using `.any?` will not be appropriative since `.any` doesn't work on `NilClass` (as my understanding). Here is my quick test:

```
[2022-08-29T11:33:56,621][ERROR][logstash.outputs.elasticsearch][main][ca10f590b...] Encountered an unexpected error submitting a bulk request, 
will retry {:message=>"undefined method `any?' for nil:NilClass", :exception=>NoMethodError, :backtrace=>["/Users/mashhur/Dev/elastic/logstash/logstash-plugins/logstash-output-elasticsearch/lib/logstash/plugin_mixins/elasticsearch/
common.rb:178:in `retrying_submit'", "/Users/mashhur/Dev/elastic/logstash/logstash-plugins/logstash-output-elasticsearch/lib/logstash/outputs/elasticsearch.rb:342:in `multi_receive'", "org/logstash/config/ir/compiler/AbstractOutputDelegatorExt.java:121:in `multi_receive'", "/Users/mashhur/Dev/elastic/logstash/logstash-core/lib/logstash/java_pipeline.rb:300:in `block in start_workers'"]}
```

mashhurs:
@yaauie thanks for the review!
Since we had alot discussions and suggestions in the conversation, wrapping up here.
- applied majority of suggestions (especially ruby related) and did overall test
- tests: 
  - unit tests covers the recently introduced `reject_unsupported_actions` method where rejects unsupported actions.
  - added integration tests to be sure we are not sending unsupported action to the ES.

Let me know your opinion if I can still improve. Thanks.

spec/integration/outputs/unsupported_actions_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+      subject.register
+      subject.multi_receive(events)
+      events.each do | event |
+        action = event.get("action_field")
+        if action.eql?("index") || action.eql?("update")
+          response = @es.get(:index => INDEX, :type => doc_type, :id => event.get("doc_id"), :refresh => true)
+          expect(response['_source']['message']).to eq(event.get("message"))
+        else
+          expect {@es.get(:index => INDEX, :type => doc_type, :id => event.get("doc_id"), :refresh => true)}.to raise_error(Elasticsearch::Transport::Transport::Errors::NotFound)
+        end
```
andsel, line 63: Having `if`s statements in test that verify different conditions could be problematic to reproduce, and they tend to ring me a bell.

Maybe something more declarative could help exposing also the intentions:
```ruby
index_or_update = proc do |x| 
  action = event.get("action_field")
  return action.eql?("index") || action.eql?("update")
end

good = events.filter(&index_or_update)      
bad = events.filter { |event| !index_or_update.call(event) }

good.each do |event|
  response = @es.get(:index => INDEX, :type => doc_type, :id => event.get("doc_id"), :refresh => true)
  expect(response['_source']['message']).to eq(event.get("message"))
end

bad.each do |event|
  expect {@es.get(:index => INDEX, :type => doc_type, :id => event.get("doc_id"), :refresh => true)}.to raise_error(Elasticsearch::Transport::Transport::Errors::NotFound)
end
```

andsel (commented):
Sorry to chime in, I saw this while working on #1084, and had similar problem in pre-validate a condition (in my case the `sprintf` of the `index` is fully resolved) and in case send to DLQ.

lib/logstash/plugin_mixins/elasticsearch/common.rb
```diff
@@ -192,6 +192,26 @@ def retrying_submit(actions)
...
+
+    def dlq_or_drop_unsupported_actions(unsupported_actions)
+      unsupported_actions.each do |action|
+        if @dlq_writer
+          @dlq_writer.write(action.event, "Unsupported action.")
+        else
+          @document_level_metrics.increment(:non_retryable_failures)
+          @logger.warn("Could not index event to Elasticsearch because its action is not supported.", action: action)
+        end
```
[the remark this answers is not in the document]
mashhurs: We discussed this [here](https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/1080#discussion_r950718369).

lib/logstash/plugin_mixins/elasticsearch/common.rb
```diff
@@ -206,19 +219,23 @@ def next_sleep_interval(current_interval)
...
-        # TODO: Change this to send a map with { :status => status, :action => action } in the future
-        @dlq_writer.write(event, "#{message} status: #{status}, action: #{action}, response: #{response}")
+        @dlq_writer.write(action.event, "#{message}")
       else
-        if dig_value(response, 'index', 'error', 'type') == 'invalid_index_name_exception'
-          level = :error
-        else
-          level = :warn
-        end
-        @logger.send level, message, status: status, action: action, response: response
```
mashhurs, line 221: @andsel Now we can globally use `handle_dlq_status` without relying on `response` (3-rd party data).
And this particular line is changed from map to ordinary string in order to standardize, rest of structures and logics are same. Any ideas if this bring issue?

andsel: I don't foresee issues, it doesn't instantiate more new objects than before, so the pressure in GC should remain the same.
This new version of `handle_dlq_status` could be easily reused also in PS #1084 (https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/1084/files#diff-940291b716938425b715566d94fceb6a35ac05e51f901276637846a9abc77737R361-R365)

andsel (approved):
LGTM

andsel:
@mashhurs please check [this error](https://app.travis-ci.com/github/logstash-plugins/logstash-output-elasticsearch/jobs/582413630#L5130) because it's strange that `Array` doesn't have `#filter` method.
[Ruby `2.5.0` Array ](https://ruby-doc.org/core-2.5.0/Array.html)seems to haven't the `#filter` method, while [Ruby `2.6` does](https://ruby-doc.org/core-2.6/Array.html).
Logstash `7.x`  ships with Ruby `2.5`.

mashhurs:
> @mashhurs please check [this error](https://app.travis-ci.com/github/logstash-plugins/logstash-output-elasticsearch/jobs/582413630#L5130) because it's strange that `Array` doesn't have `#filter` method. [Ruby `2.5.0` Array ](https://ruby-doc.org/core-2.5.0/Array.html)seems to haven't the `#filter` method, while [Ruby `2.6` does](https://ruby-doc.org/core-2.6/Array.html). Logstash `7.x` ships with Ruby `2.5`.

Ah it is because of ruby version. We are using `jruby 9.2.20.1 (2.5.8)` and `.filter` seems introduced in `2.6.0`.

lib/logstash/plugin_mixins/elasticsearch/common.rb
mashhurs: Thanks for the confirmation!

lib/logstash/plugin_mixins/elasticsearch/common.rb
mashhurs: `handle_dlq_status` method is now used as a common to handle such cases.

spec/unit/outputs/elasticsearch/http_client_spec.rb
mashhurs: unit tests cover all these mentioned cases and integration tests guarantee unsupported actions did reach ES with CRUD operation.

andsel (approved):
LGTM

lib/logstash/outputs/elasticsearch.rb
```diff
@@ -367,36 +367,37 @@ def multi_receive(events)
...
+      begin
         successful_events << @event_mapper.call(event)
-       rescue IndexInterpolationError, e
-         action = event.sprintf(@action || 'index')
-         event_action_tuple = EventActionTuple.new(action, [], event)
-         failed_events << event_action_tuple
+      rescue EventMappingError => ie
+        action = event.sprintf(@action || 'index')
+        event_action_tuple = EventActionTuple.new(action, [], event)
+        event_mapping_errors << [event_action_tuple, ie.message]
```
mashhurs, line 397: @andsel , I have unified badly interpolated index name case with unsupported action. They both seem to me and Ry (as we quickly talked) common `EventMappingError` with different messages.
Also, as my understanding you used the `EventActionTuple` to send entire event body to DLQ. It seems we don't necessarily `sprintf` here and return `EventActionTuple` list, rather use the `Logstash::Event`.
I have tried `Logstash::Event` but I wasn't able to create detail message in L379. The reason `event.to_s` runs `Event.java -> toString()` and instead of inspecting entire data, it only considers `host` and `message` field. So for now, I have also aligned with `EventActionTuple` to properly send entire event data to DLQ (or drop by logging detail message). Let me know your opinion if we can still improve. Huge thanks for clear logic placement, BTW.

cc: @yaauie

andsel: The motivation why I originally used `EventActionTuple` also for failed events, was that the original `#handle_dlq_status` required that as input argument, plus was kept also for type symmetry  in the two lists that composes `MapEventsResult(:successful_events, :failed_events)`.
I like the rephrase of `IndexInterpolationError` to inherit from `EventMappingError`.

spec/unit/outputs/elasticsearch_spec.rb
```diff
@@ -381,6 +381,86 @@
...
+        let(:events) {[
+          LogStash::Event.new("action_field" => "index", "id" => 1, "message"=> "world!"),
+          LogStash::Event.new("action_field" => "index", "id" => 2, "message"=> "hello"),
+          LogStash::Event.new("action_field" => "unsupported_action3", "id" => 3, "message"=> "hi"),
+          LogStash::Event.new("action_field" => "index", "id" => 4, "message"=> "bye")
+        ]}
+        it "rejects unsupported action" do
+          event_result = subject.send(:safe_interpolation_map_events, events)
+          expect(event_result.successful_events.size).to be == 3
+          # expect(logger_stub).to have_received(:warn).with(a_string_including "Could not index event to Elasticsearch because its action is not supported.")
```
andsel, line 460: Should be re-enabled to check also the failed ones?

spec/unit/outputs/elasticsearch_spec.rb
```diff
@@ -381,6 +381,86 @@
...
+        let(:events) {[
+          LogStash::Event.new("action_field" => "index", "id" => 1, "message"=> "world!"),
+          LogStash::Event.new("action_field" => "unsupported_action2", "id" => 2, "message"=> "hello"),
+          LogStash::Event.new("action_field" => "unsupported_action3", "id" => 3, "message"=> "hi"),
+          LogStash::Event.new("action_field" => "index", "id" => 4, "message"=> "bye")
+        ]}
+        it "rejects unsupported actions" do
+          event_result = subject.send(:safe_interpolation_map_events, events)
+          expect(event_result.successful_events.size).to be == 2
+          # expect(logger_stub).to have_received(:warn).with(a_string_including "Could not index event to Elasticsearch because its action is not supported.")
```
andsel, line 446: Should be re-enabled to check also the failed ones?

lib/logstash/outputs/elasticsearch.rb
```diff
@@ -367,36 +367,37 @@ def multi_receive(events)
     wait_for_successful_connection if @after_successful_connection_done
     events_mapped = safe_interpolation_map_events(events)
     retrying_submit(events_mapped.successful_events)
-    unless events_mapped.failed_events.empty?
-      @logger.error("Can't map some events, needs to be handled by DLQ #{events_mapped.failed_events}")
-      send_failed_resolutions_to_dlq(events_mapped.failed_events)
+    unless events_mapped.event_mapping_errors.empty?
+      handle_event_mapping_errors(events_mapped.event_mapping_errors)
```
andsel, line 371: If dlq is enabled then no log lines are logged. I think that  any mapping error, that route events to DLQ, should be clearly highlighted into the logs, so a `warn` log line should be present to eventually understand what's happening.

andsel (changes requested):
The changes you proposed seems good to me.
Left a couple of notes on test's expectations, if should be re-enabled plus a concern about the eventuality to log something in case a mapping error happens.

lib/logstash/outputs/elasticsearch.rb
```diff
@@ -367,36 +367,37 @@ def multi_receive(events)
...
-  # @param: Arrays of EventActionTuple
+  # @param: Arrays of EventActionTuple with error messages
   private
-  def send_failed_resolutions_to_dlq(failed_action_tuples)
-    failed_action_tuples.each do |action|
-      handle_dlq_status(action, "warn", "Could not resolve dynamic index")
+  def handle_event_mapping_errors(event_mapping_errors)
+    event_mapping_errors.each do |action, message|
+      detail_message = message + ", " + action.to_s
+      handle_dlq_status(action.event, :warn, detail_message)
```
yaauie, line 380: I would prefer a rephrase of this, so that `event_mapping_errors` was _either_ an array of _`event`_/`message` tuples or an array of objects that have `event`/`message` properties, since `action` in this code is somewhat ambiguous and can mean either a String action or an action-tuple as prepared for a bulk request.

When we are building this array, we already have access to the event and an object that _has_ the message (the exception), and when we are raising the exception we have access to the `event`. We could either create a specialized struct, or can embed the event into the exception that we are already handling.

Additionally, we have everything we need to produce our `detail_message` when populating the array (or even when throwing the relevant exception that leads to an entry in the array), so I would rather pull that up.

lib/logstash/outputs/elasticsearch.rb
mashhurs: FYI: I have changed to Event as we don't need extra `sprintf`.

spec/unit/outputs/elasticsearch_spec.rb
mashhurs: Yup, added in new commits.

spec/unit/outputs/elasticsearch_spec.rb
```diff
@@ -381,6 +381,100 @@
...
+      context "with multiple valid actions with one trailing invalid action" do
+        let(:events) {[
+          LogStash::Event.new("action_field" => "index", "id" => 1, "message"=> "hello"),
+          LogStash::Event.new("action_field" => "index", "id" => 2, "message"=> "hi"),
+          LogStash::Event.new("action_field" => "index", "id" => 3, "message"=> "bye"),
+          LogStash::Event.new("action_field" => "unsupported_action", "id" => 4, "message"=> "world!")
+        ]}
+        it "rejects unsupported actions" do
+          event_result = subject.send(:safe_interpolation_map_events, events)
+          expect(event_result.successful_events.size).to be == 3
```
mashhurs, line 397: @yaauie , I really liked the idea of having `rspec-collection_matchers` when unit tests check agains collections sizes. I understood more about use cases after taking a look at [this thread](https://github.com/rspec/rspec-expectations/issues/785).
However, in order to use `collection_matchers`, we need to onboard it with LS-core (add `gem "rspec-collection_matchers", :group => :development` in `Gemfile.template` file).

Let me know your thoughts if I can add it to plugin level dependency.

yaauie: I could go either way -- the failure scenario reads "clear enough" for these specs.

The normal way of adding the dependency would be to declare a development dependency in this project's gemspec, not in Logstash core.

yaauie (approved):
Left one follow-up on the collections expectations. Feel free to either chase it down or merge as-is.

yaauie (approved):
LGTM 🎉

PR #1081, opened 2022-08-24
fix: avoid `message` in manticore exception logging context

yaauie:
When Logstash is configured to use JSON-formatted logs, the Log4j2 JSONLayout
will serialize both the provided message and the context map's `message` in
the same key/value collection. This is _syntactically_ valid, because JSON
does not impose constraints on the uniqueness of keys in a key/value
collection, but considered _semantically_ invalid by many parsers including
Beats that attempt to deserialize into a structure that _do_ impose
uniqueness constraints.

Here we ensure that we provide the _contents_ of the exception's message
inside a single message, without also providing a `message` key to the
logging context.


robbavey:
FYI, this is a broader issue than just the elasticsearch output - https://github.com/elastic/logstash/issues/14335

PR #1084, opened 2022-09-07
Route badly interpolated dynamic index to DLQ

### LINKED ISSUE #718: DLQ when sprintf references a missing field

andsel:
## Release notes

Modified the resolution of dynamic index name so that in case of interpolation errors the event is sent to the DLQ. This feature could be explicitly disabled configuring properly the `dlq_on_failed_indexname_interpolation`.

## What does this PR do?
Update the the pre-indexing steps, so to that in case of an `IndexInterpolationError` is raised during the index name resolution, the event is sent to DLQ. It leverarges the refactored `#handle_dlq_status` from PR #1080 to reuse the enqueuing logic.

Let the user to handle specifically the events which once indexed would create an ugly formatted name like  `%{[not_existing_field]}_index`.

## Checklist

## Author's Checklist

- [x] run with 
```
.ci/docker-setup.sh && INTEGRATION=true .ci/docker-run.sh
```
- [x] adds feature flag to disable this behavior. Track telemetry data to know how many uses this explicit disabling.

- enable the DLQ,  set `dead_letter_queue.enable: true` in `config/logstash.yml`.
- launch an Elasticsearch instance
- execute a pipeline which wouldn't  resolve the index name like:
```
input {
	generator {
		message => '{"name": "John", "surname": "Doe"}'
		count => 1
		codec => json
	}
}

output {
	elasticsearch {
		index => "%{[not_existing_field]}"
		hosts => "http://localhost:9200"
		user => "elastic"
		password => "changeme"
	}
}
```
- then drain the DQL with another pipeline and verify it contains the previous failed event:
```
input {
  dead_letter_queue {
  	path => "/path_to/data/dead_letter_queue/"
  }
}

output {
  stdout {
    codec => rubydebug
  }

```
## Related issues

- Fixes #718
- Copies a refactoring done in #1080

## Use cases
As a user I don't want that unresolved index names generates unexpected indexes and I would to reprocess such failed events with another DLQ pipeline.

### DIFF
--- a/lib/logstash/outputs/elasticsearch.rb
+++ b/lib/logstash/outputs/elasticsearch.rb
@@ -479,7 +479,7 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
     raise IndexInterpolationError, sprintf_index if sprintf_index.match(/%{.*?}/)
     params = {
         :_id => @document_id ? event.sprintf(@document_id) : nil,
-         :_index => sprintf_index,
+        :_index => sprintf_index,
         routing_field_name => @routing ? event.sprintf(@routing) : nil
     }
 
--- a/lib/logstash/outputs/elasticsearch.rb
+++ b/lib/logstash/outputs/elasticsearch.rb
@@ -261,6 +261,9 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
   # The option accepts a list of natural numbers corresponding to HTTP errors codes.
   config :dlq_custom_codes, :validate => :number, :list => true, :default => []
 
+  # if enabled, failed index name interpolation events go into dead letter queue.
+  config :dlq_on_failed_indexname_interpolation, :validate => :boolean, :default => true
+
   attr_reader :client
   attr_reader :default_index
   attr_reader :default_ilm_rollover_alias
@@ -487,7 +490,7 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
   # @private shared event params factory between index and data_stream mode
   def common_event_params(event)
     sprintf_index = @event_target.call(event)
-    raise IndexInterpolationError, sprintf_index if sprintf_index.match(/%{.*?}/)
+    raise IndexInterpolationError, sprintf_index if sprintf_index.match(/%{.*?}/) && dlq_on_failed_indexname_interpolation
     params = {
         :_id => @document_id ? event.sprintf(@document_id) : nil,
         :_index => sprintf_index,
--- a/lib/logstash/outputs/elasticsearch.rb
+++ b/lib/logstash/outputs/elasticsearch.rb
@@ -374,9 +374,10 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
   private
   def send_failed_resolutions_to_dlq(failed_action_tuples)
     failed_action_tuples.each do |action|
-      status = nil
-      response = {}
-      handle_dlq_status("Could not resolve dynamic index", action, status, response)
+#       status = nil
+#       response = {}
+#       handle_dlq_status("Could not resolve dynamic index", action, status, response)
+      handle_dlq_status(action, "warn", "Could not resolve dynamic index")
     end
   end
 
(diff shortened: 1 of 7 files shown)


lib/logstash/outputs/elasticsearch.rb
```diff
@@ -362,11 +365,43 @@ def config_init(params)
   # Receive an array of events and immediately attempt to index them (no buffering)
   def multi_receive(events)
     wait_for_successful_connection if @after_successful_connection_done
-    retrying_submit map_events(events)
+    events_mapped = safe_interpolation_map_events(events)
```
mashhurs, line 368: How do you think if we reject index name unresolved events and only bring back successful events?
In my [unmerged PR](https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/1080/files#diff-221ddb62bde6d62966fab1051387ded7513fffd575cf0706a5f3a94ad2d31307R198), we had a great discussion with Ry and agreed to work with successful/submittable events here, send to DLQ rejected events wen filtering.

lib/logstash/outputs/elasticsearch.rb
```diff
@@ -362,11 +365,43 @@ def config_init(params)
...
+    events.each do |event|
+       begin
+        successful_events << @event_mapper.call(event)
+       rescue IndexInterpolationError, e
+         action = event.sprintf(@action || 'index')
+         event_action_tuple = EventActionTuple.new(action, [], event)
+         failed_events << event_action_tuple
+       end
+    end
+    MapEventsResult.new(successful_events, failed_events)
```
mashhurs, line 399:
```suggestion
    unless failed_events.empty?
      @logger.error("Can't map some events, needs to be handled by DLQ #{failed_events}")
      send_failed_resolutions_to_dlq(failed_events)
    end
    successful_events
```

lib/logstash/plugin_mixins/elasticsearch/common.rb
```diff
@@ -206,19 +206,23 @@ def next_sleep_interval(current_interval)
...
-    def handle_dlq_status(message, action, status, response)
+    def handle_dlq_response(message, action, status, response)
+      _, action_params = action.event, [action[0], action[1], action[2]]
+      
+      # TODO: Change this to send a map with { :status => status, :action => action } in the future
+      detailed_message = "#{message} status: #{status}, action: #{action_params}, response: #{response}"
+      
+      log_level = dig_value(response, 'index', 'error', 'type') == 'invalid_index_name_exception' ? :error : :warn
+      
+      handle_dlq_status(action, log_level, detailed_message)
```
mashhurs, line 217: Special thanks for including `handle_dlq_response` & `handle_dlq_status`!

mashhurs (changes requested):

yaauie (approved):
Changes LGTM.

I'm not terribly fond of raising exceptions for flow control, but feel that the structure as-proposed here is simpler and _much_ more clear than alternatives.

My only concern is conflicts with the also-in-flight #1080, which similarly pre-filters events whose _action_ cannot be interpolated.

mashhurs:
> Changes LGTM.
> 
> I'm not terribly fond of raising exceptions for flow control, but feel that the structure as-proposed here is simpler and _much_ more clear than alternatives.
> 
> My only concern is conflicts with the also-in-flight #1080, which similarly pre-filters events whose _action_ cannot be interpolated.

Initially I was thinking that seemed to me easily include my #1080 changes but it takes sometime to make an adjustment (especially unit test cases). Feel free to go first @andsel !

Issue #1088, opened 2022-09-29
request of using legacy template API in Elasticsearch 8.x


kaisecheng:
The composable [index templates](https://www.elastic.co/guide/en/elasticsearch/reference/8.4/index-templates.html) are introduced in Elasticsearch 7.8. The plan was to deprecate the [legacy template API](https://www.elastic.co/guide/en/elasticsearch/reference/8.4/indices-templates-v1.html) in Elasticsearch 8.0, but as of now, the [deprecation](https://github.com/elastic/elasticsearch/issues/53101) defers to 9.0.

It is valid to use both APIs in ES 8.x. Currently, the plugin uses legacy API for ES 7.x and forces to use index template API for ES 8.x. This issue is to request backward compatibility for users to choose which API to use.

One suggestion is to provide a boolean flag `template_legacy` => auto/true/false (default auto) to use legacy API

PR #1089, opened 2022-09-29
[DOC] Add index template restriction

kaisecheng:
Relates: #1088

The composable [index templates](https://www.elastic.co/guide/en/elasticsearch/reference/8.4/index-templates.html) are introduced in Elasticsearch 7.8. The plan was to deprecate the [legacy template API](https://www.elastic.co/guide/en/elasticsearch/reference/8.4/indices-templates-v1.html) in Elasticsearch 8.0, but as of now, the [deprecation](https://github.com/elastic/elasticsearch/issues/53101) defers to 9.0.

This document explains the expectation of the plugin that the template used for ES 8.x is different to ES 7.x

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 11.9.3
+  - [DOC] Add index template restriction
+
 ## 11.9.2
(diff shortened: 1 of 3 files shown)


kaisecheng:
implement template_api instead of documenting

PR #1085, opened 2022-09-22
Fix broken link to Logstash Reference 

karenzone:

PR #1091, opened 2022-09-30
Doc: Clarify wording on action option

karenzone:
We fixed a broken link in #1085 during a live demo, and introduced some unclear wording.  This PR fixes that.

No version bump/publish required for this change.


karenzone (commented):
Tested locally. Builds and renders as expected

PR #1094, opened 2022-10-21
Update index.asciidoc

Gunnerva:

I have added a section to clarify outputting to Winlogbeat created Data Stream.

### DIFF
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -93,6 +93,21 @@ output {
 }
 -----
 
+**Example: Configuration for Winlogbeats to Output to Winlogbeat created Data Stream**
+
+[source,sh]
+-----
+output {
+    elasticsearch {
+        hosts => "hostname"
+        index => "winlogbeat-8.4.3"
+        pipeline => "winlogbeat-8.4.3-routing"
+        action => "create"
+    }
+}
+-----
+
+
 This example shows the minimal settings for processing data streams. Events
 with `data_stream.*`` fields are routed to the appropriate data streams. If the
 fields are missing, routing defaults to `logs-generic-logstash`.


docs/index.asciidoc
```diff
@@ -93,6 +93,21 @@ output {
 }
 -----
 
+**Example: Configuration for Winlogbeats to Output to Winlogbeat created Data Stream**
```
jsvd, line 96: This example is being inserted between the code block and explanation text of the previous example. Please move it to after the  "Customize data stream name" example.

docs/index.asciidoc
```diff
@@ -93,6 +93,21 @@ output {
...
+output {
+    elasticsearch {
+        hosts => "hostname"
+        index => "winlogbeat-8.4.3"
+        pipeline => "winlogbeat-8.4.3-routing"
+        action => "create"
+    }
+}
+-----
+
```
jsvd, line 109:
```suggestion

This example shows how to handle writing to data streams that don't conform to the "logs-generic-default" format.
For that purpose, you can disable datastreams, configure the name of the data stream to be written to in the "index =>" setting, and set "action" to "create" as it is the only action supported in data streams.
```

docs/index.asciidoc
```diff
@@ -93,6 +93,21 @@ output {
...
+**Example: Configuration for Winlogbeats to Output to Winlogbeat created Data Stream**
+
+[source,sh]
+-----
+output {
+    elasticsearch {
+        hosts => "hostname"
+        index => "winlogbeat-8.4.3"
+        pipeline => "winlogbeat-8.4.3-routing"
+        action => "create"
```
jsvd, line 105:
```suggestion
        action => "create"
        data_stream => "false"
```

Issue #999, opened 2021-03-17
When bulk request rejected with HTTP 400, we should log cause at higher level


yaauie:
When an entire bulk request is rejected by Elasticsearch, we should include the response body from Elasticsearch regardless of `debug`-level logging being enabled, since this is not a normal condition and is the error message should be actionable.

Without `debug`-level logging:
~~~
[2021-03-16T15:40:05,355][ERROR][logstash.outputs.elasticsearch][main][es-output] Encountered a retryable error. Will Retry with exponential backoff {:code=>400, :url=>"https://REDACTED:443/_bulk"}
~~~

With `debug`-level logging enabled:
~~~
[2021-03-16T16:59:50,322][ERROR][logstash.outputs.elasticsearch][main][es-output] Encountered a retryable error. Will Retry with exponential backoff {:code=>400, :url=>“https://REDACTED:443/_bulk”, :body=>“{\“error\“:{\“root_cause\“:[{\“type\“:\“illegal_argument_exception\“,\“reason\“:\“The bulk request must be terminated by a newline [\\\\n]\“}],\“type\“:\“illegal_argument_exception\“,\“reason\“:\“The bulk request must be terminated by a newline [\\\\n]\“},\“status\“:400}“}
~~~

PR #1096, opened 2022-11-14
Log bulk request response body on error, not just when debug logging …

robbavey:
…is set

Prior to this commit, an error response to a bulk request would only log the response body when debug logging is enabled. This removes an important piece of information when tracking down issues, which is particularly important in this case, as an error response to a bulk request may result in requests being retried indefinitely, causing pipeline stalls.

This commit adds the response body to the `log_hash` regardless of the log level,so 429 responses will remain unchanged, and will only log when debug is enabled, but other error responses will include the response body to add this additional context to help track down issues

Closes: #999


andsel (approved):
LGTM.

Please, remember to add 
```
closes: #999
```
to the comment to autoclose the issue and bump a new version.

PR #1098, opened 2022-11-18
Update changelog entry with more details

karenzone:
Updates changelog to point to fix rather than to only the issue


andsel (approved):
LGTM

PR #1102, opened 2022-12-16
Resolve ILM settings based loaded template or/and `template_api`

### LINKED ISSUE #1040: Manage template (ILM) fails if settings section is missing
Issue elastic/logstash#11893 was opened by @tmanninger but I think needs to be in this repo instead.
I've done some rewording and provided details more relevant to our environment.

**Logstash information**:

1. Logstash version **7.4.0**
2. Logstash installation source: **RPM package from release repo**
3. How is Logstash being run: **systemd**
4. How was the Logstash Plugin installed: **bundled**

**JVM** :

1. JVM version (`java -version`)
```
openjdk version "11.0.11" 2021-04-20 LTS
OpenJDK Runtime Environment 18.9 (build 11.0.11+9-LTS)
OpenJDK 64-Bit Server VM 18.9 (build 11.0.11+9-LTS, mixed mode, sharing)
```

2. JVM installation source: **RPM package from release repo**
3. Value of the `JAVA_HOME` environment variable: **not set**

**OS version** : 
```
Linux xxx-logstash-staging 3.10.0-1160.31.1.el7.x86_64 #1 SMP Thu Jun 10 13:32:12 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
```

If the plugin includes configuration for ILM, adding these settings to the template **fails** if the template has no "settings" section. The log file shows an opaque `NoMethodError` message from `add_ilm_settings_to_template`:

```
Failed to install template. {:message=>"undefined method `update' for nil:NilClass", :class=>"NoMethodError", :backtrace=>["/usr/share/logstash/vendor/bundle/jruby/2.5.0/gems/logstash-output-elasticsearch-10.3.1-java/lib/logstash/outputs/elasticsearch/template_manager.rb:39:in `add_ilm_settings_to_template'", "/usr/share/logstash/vendor/bundle/jruby/2.5.0/gems/logstash-output-elasticsearch-10.3.1-java/lib/logstash/outputs/elasticsearch/template_manager.rb:14:in `install_template'", "/usr/share/logstash/vendor/bundle/jruby/2.5.0/gems/logstash-output-elasticsearch-10.3.1-java/lib/logstash/outputs/elasticsearch/common.rb:197:in `install_template'", "/usr/share/logstash/vendor/bundle/jruby/2.5.0/gems/logstash-output-elasticsearch-10.3.1-java/lib/logstash/outputs/elasticsearch/common.rb:53:in `block in setup_after_successful_connection'"]}
```
 
**Expected**: Logstash would apply ILM settings even when a minimal template has no static settings.
**Actual**: it doesn't. Further, the pipeline loads and runs, obscuring the error.

[template_manager.rb](https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/master/lib/logstash/outputs/elasticsearch/template_manager.rb) should create the `settings` field if it does not exist, before attempting to populate it.

Further, the error message should provide the context of which pipeline and template were being processed - I encountered this while adding ILM to a system which already had many pipelines, and had to do a fair bit of digging to unearth the problem.

**Workaround**: add an empty `settings` field to all templates to be used with ILM.

### LINKED ISSUE #1108: Plugin fails to install default template when using legacy API with Elasticsearch 8.x (or higher) versions.
### Issue description

When we upgrade Elasticsearch to 8.x and use legacy template (`template_api`=>`legacy`), the plugin resolves the composable style template as a default template and send to legacy API, obviously it doesn't work (see the produced error below). 
The discussion with team brought to the idea of using legacy template api with custom template only. However, if ILM (ilm_* settings) is enabled, the plugin crashes. When plugin tries to set ILM settings to provided custom template, the plugin aligns on Elasticsearch version (pseudo: `template_settings  = version < 8 ? template['settings']: template['template']['settings']`). Obviously, with custom _flexible_ template it is hard to guess where ILM settings located.

**Steps to reproduce**:

Please include a *minimal* but *complete* recreation of the problem,
including (e.g.) pipeline definition(s), settings, locale, etc.  The easier
you make for us to reproduce it, the more likely that somebody will take the
time to look at it.

 1. set `template_api` to `legacy`
 2. connect to any (including cloud) Elasticsearch where its version is 8.x+
 3. run Logstash and check the log

```
[ERROR] 2022-12-13 23:42:54.690 [Ruby-0-Thread-12: :1] elasticsearch - Failed to install template 
{:message=>"Got response code '400' contacting Elasticsearch at URL 
'https://my-es-host.com:9200/_template/logstashb'", :exception=>LogStash::Outputs::ElasticSearch::HttpClient::Pool::BadResponseCodeError, 
:backtrace=>["/grid01/apps/PIPELINES/logstash-8.2.3/vendor/bundle/jruby/2.5.0/gems/logstash-output-elasticsearch-11.12.1-
java/lib/logstash/outputs/elasticsearch/http_client/manticore_adapter.rb:84:in `perform_request'", 
"/grid01/apps/PIPELINES/logstash-8.2.3/vendor/bundle/jruby/2.5.0/gems/logstash-output-elasticsearch-11.12.1-
java/lib/logstash/outputs/elasticsearch/http_client/pool.rb:324:in `perform_request_to_url'", "/grid01/apps/PIPELINES/logstash-8.2.3/vendor/bundle/jruby/2.5.0/gems/logstash-output-elasticsearch-11.12.1-
java/lib/logstash/outputs/elasticsearch/http_client/pool.rb:311:in `block in perform_request'", "/grid01/apps/PIPELINES/logstash-8.2.3/vendor/bundle/jruby/2.5.0/gems/logstash-output-elasticsearch-11.12.1-
java/lib/logstash/outputs/elasticsearch/http_client/pool.rb:398:in `with_connection'", "/grid01/apps/PIPELINES/logstash-8.2.3/vendor/bundle/jruby/2.5.0/gems/logstash-output-elasticsearch-11.12.1-
java/lib/logstash/outputs/elasticsearch/http_client/pool.rb:310:in `perform_request'", "/grid01/apps/PIPELINES/logstash-8.2.3/vendor/bundle/jruby/2.5.0/gems/logstash-output-elasticsearch-11.12.1-java/lib/logstash/outputs/elasticsearch/http_client/pool.rb:318:in `block in Pool'", "/grid01/apps/PIPELINES/logstash-8.2.3/vendor/bundle/jruby/2.5.0/gems/logstash-output-elasticsearch-11.12.1-java/lib/logstash/outputs/elasticsearch/http_client.rb:412:in `template_put'", "/grid01/apps/PIPELINES/logstash-8.2.3/vendor/bundle/jruby/2.5.0/gems/logstash-output-elasticsearch-11.12.1-java/lib/logstash/outputs/elasticsearch/http_client.rb:85:in `template_install'", "/grid01/apps/PIPELINES/logstash-8.2.3/vendor/bundle/jruby/2.5.0/gems/logstash-output-elasticsearch-11.12.1-java/lib/logstash/outputs/elasticsearch/template_manager.rb:41:in `install'", "/grid01/apps/PIPELINES/logstash-8.2.3/vendor/bundle/jruby/2.5.0/gems/logstash-output-elasticsearch-11.12.1-
java/lib/logstash/outputs/elasticsearch/template_manager.rb:29:in `install_template'", "/grid01/apps/PIPELINES/logstash-
8.2.3/vendor/bundle/jruby/2.5.0/gems/logstash-output-elasticsearch-11.12.1-java/lib/logstash/outputs/elasticsearch.rb:583:in 
`install_template'", "/grid01/apps/PIPELINES/logstash-8.2.3/vendor/bundle/jruby/2.5.0/gems/logstash-output-elasticsearch-11.12.1-
java/lib/logstash/outputs/elasticsearch.rb:349:in `finish_register'", "/grid01/apps/PIPELINES/logstash-
8.2.3/vendor/bundle/jruby/2.5.0/gems/logstash-output-elasticsearch-11.12.1-java/lib/logstash/outputs/elasticsearch.rb:305:in 
`block in register'", "/grid01/apps/PIPELINES/logstash-8.2.3/vendor/bundle/jruby/2.5.0/gems/logstash-output-elasticsearch-11.12.1-
java/lib/logstash/plugin_mixins/elasticsearch/common.rb:167:in `block in after_successful_connection'"]}
```

### Acceptance Criteria
- Legacy template API is used with only custom templates
- Plugin should not crash, instead provide a clear message/guide what to do next

mashhurs:
### Issue description
When installing the template (either custom or auto resolved), sometimes there is a mismatch. For example, without this change, the case with `template_api=>'legacy'` connecting to ES 8+ resolves the wrong ILM setting field and crashes.

### What this change introduces?
This change introduces ILM setting resolution based on custom/loaded template or/and `template_api` settings.

### Acceptance criteria
For the given any settings, including custom template or/and `template_api`, plugin needs to resolve ILM setting instead crashing.

### Testing
- Unit tests
```
Finished in 1 minute 22.92 seconds (files took 7.1 seconds to load)
269 examples, 0 failures

Randomized with seed 22550
```

- Closes #1108
- Closes #1040

--- 
_History_
The change mainly solves two problems:
1. the plugin crashes when resolving and installing default template using legacy template (`legacy` with ES 8.x versions, `composable` with ES 7.x versions behaves same). This change encourages customers using custom their _**own custom template when using legacy template API (`legacy` or `composable`) with any ES versions**_.
2. the plugin crashes when using custom template with ILM enabled. This change opts out the ILM setting processes when using custom template and discourages users using custom template when ILM enabled, informs to set ILM settings in the template.

### DIFF
--- a/lib/logstash/outputs/elasticsearch/template_manager.rb
+++ b/lib/logstash/outputs/elasticsearch/template_manager.rb
@@ -54,19 +54,27 @@ module LogStash; module Outputs; class ElasticSearch
     end
 
     def self.resolve_template_settings(plugin, template)
-      return composable_index_template_settings(template) if template.key?('template')
-      return legacy_index_template_settings(template) if template.key?('settings')
-      return template_endpoint(plugin) == INDEX_TEMPLATE_ENDPOINT ?
-               composable_index_template_settings(template) :
-               legacy_index_template_settings(template)
+      if template.key?('template')
+        plugin.logger.trace("Resolving ILM template settings: under 'template' key", :template => template, :template_api => plugin.template_api, :es_version => plugin.maximum_seen_major_version)
+        composable_index_template_settings(template)
+      elsif template.key?('settings')
+        plugin.logger.trace("Resolving ILM template settings: under 'settings' key", :template => template, :template_api => plugin.template_api, :es_version => plugin.maximum_seen_major_version)
+        legacy_index_template_settings(template)
+      else
+        template_endpoint = template_endpoint(plugin)
+        plugin.logger.trace("Resolving ILM template settings: template doesn't have 'settings' or 'template' fields, falling back to auto detection", :template => template, :template_api => plugin.template_api, :es_version => plugin.maximum_seen_major_version, :template_endpoint => template_endpoint)
+        template_endpoint == INDEX_TEMPLATE_ENDPOINT ?
+          composable_index_template_settings(template) :
+          legacy_index_template_settings(template)
+      end
     end
 
-    # Returns (if exists) or creates _template API compatible template settings
+    # Sets ['settings'] field to be compatible with _template API structure
     def self.legacy_index_template_settings(template)
       template['settings'] ||= {}
     end
 
-    # Returns (if exists) or creates _index_template API compatible template settings
+    # Sets the ['template']['settings'] fields if not exist to be compatible with _index_template API structure
     def self.composable_index_template_settings(template)
       template['template'] ||= {}
       template['template']['settings'] ||= {}
--- a/lib/logstash/outputs/elasticsearch/template_manager.rb
+++ b/lib/logstash/outputs/elasticsearch/template_manager.rb
@@ -46,15 +46,30 @@ module LogStash; module Outputs; class ElasticSearch
       # definition - remove any existing definition of 'template'
       template.delete('template') if template.include?('template') if plugin.maximum_seen_major_version < 8
       template['index_patterns'] = "#{plugin.ilm_rollover_alias}-*"
-      settings = template_settings(plugin, template)
+      settings = resolve_template_settings(plugin, template)
       if settings && (settings['index.lifecycle.name'] || settings['index.lifecycle.rollover_alias'])
         plugin.logger.info("Overwriting index lifecycle name and rollover alias as ILM is enabled")
       end
       settings.update({ 'index.lifecycle.name' => plugin.ilm_policy, 'index.lifecycle.rollover_alias' => plugin.ilm_rollover_alias})
     end
 
-    def self.template_settings(plugin, template)
-      plugin.maximum_seen_major_version < 8 ? template['settings']: template['template']['settings']
+    def self.resolve_template_settings(plugin, template)
+      return composable_index_template_settings(template) if template.key?('template')
+      return legacy_index_template_settings(template) if template.key?('settings')
+      return template_endpoint(plugin) == INDEX_TEMPLATE_ENDPOINT ?
+               composable_index_template_settings(template) :
+               legacy_index_template_settings(template)
+    end
+
+    # Returns (if exists) or creates _template API compatible template settings
+    def self.legacy_index_template_settings(template)
+      template['settings'] ||= {}
+    end
+
+    # Returns (if exists) or creates _index_template API compatible template settings
+    def self.composable_index_template_settings(template)
+      template['template'] ||= {}
+      template['template']['settings'] ||= {}
     end
 
     # Template name - if template_name set, use it
--- a/spec/unit/outputs/elasticsearch/template_manager_spec.rb
+++ b/spec/unit/outputs/elasticsearch/template_manager_spec.rb
@@ -73,6 +73,7 @@ describe LogStash::Outputs::ElasticSearch::TemplateManager do
         let(:template_api) { "composable" }
 
         it 'resolves composable index template API compatible setting' do
+          expect(plugin).to receive(:maximum_seen_major_version).at_least(:once).and_return(8) # required to log
           template = {}
           described_class.resolve_template_settings(plugin, template)
           expect(template["template"]["settings"]).not_to eq(nil)
@@ -83,6 +84,7 @@ describe LogStash::Outputs::ElasticSearch::TemplateManager do
         let(:template_api) { "legacy" }
 
         it 'resolves legacy index template API compatible setting' do
+          expect(plugin).to receive(:maximum_seen_major_version).at_least(:once).and_return(7) # required to log
           template = {}
           described_class.resolve_template_settings(plugin, template)
           expect(template["settings"]).not_to eq(nil)
--- a/spec/unit/outputs/elasticsearch/template_manager_spec.rb
+++ b/spec/unit/outputs/elasticsearch/template_manager_spec.rb
@@ -1,4 +1,4 @@
-require "logstash/devutils/rspec/spec_helper"
+require_relative "../../../../spec/spec_helper"
 require "logstash/outputs/elasticsearch/template_manager"
 
 describe LogStash::Outputs::ElasticSearch::TemplateManager do
@@ -33,33 +33,83 @@ describe LogStash::Outputs::ElasticSearch::TemplateManager do
     end
   end
 
-  describe "index template with ilm settings" do
+  context "index template with ilm settings" do
     let(:plugin_settings) { {"manage_template" => true, "template_overwrite" => true} }
     let(:plugin) { LogStash::Outputs::ElasticSearch.new(plugin_settings) }
 
-    describe "in version 8+" do
-      let(:file_path) { described_class.default_template_path(8) }
-      let(:template) { described_class.read_template_file(file_path)}
+    describe "with custom template" do
 
-      it "should update settings" do
-        expect(plugin).to receive(:maximum_seen_major_version).at_least(:once).and_return(8)
-        described_class.add_ilm_settings_to_template(plugin, template)
-        expect(template['template']['settings']['index.lifecycle.name']).not_to eq(nil)
-        expect(template['template']['settings']['index.lifecycle.rollover_alias']).not_to eq(nil)
-        expect(template.include?('settings')).to be_falsey
+      describe "in version 8+" do
+        let(:file_path) { described_class.default_template_path(8) }
+        let(:template) { described_class.read_template_file(file_path)}
+
+        it "should update settings" do
+          expect(plugin).to receive(:maximum_seen_major_version).at_least(:once).and_return(8)
+          described_class.add_ilm_settings_to_template(plugin, template)
+          expect(template['template']['settings']['index.lifecycle.name']).not_to eq(nil)
+          expect(template['template']['settings']['index.lifecycle.rollover_alias']).not_to eq(nil)
+          expect(template.include?('settings')).to be_falsey
+        end
+      end
+
+      describe "in version < 8" do
+        let(:file_path) { described_class.default_template_path(7) }
+        let(:template) { described_class.read_template_file(file_path)}
+
+        it "should update settings" do
+          expect(plugin).to receive(:maximum_seen_major_version).at_least(:once).and_return(7)
+          described_class.add_ilm_settings_to_template(plugin, template)
+          expect(template['settings']['index.lifecycle.name']).not_to eq(nil)
+          expect(template['settings']['index.lifecycle.rollover_alias']).not_to eq(nil)
+          expect(template.include?('template')).to be_falsey
+        end
       end
     end
 
-    describe "in version < 8" do
-      let(:file_path) { described_class.default_template_path(7) }
-      let(:template) { described_class.read_template_file(file_path)}
+    context "resolve template setting" do
+      let(:plugin_settings) { super().merge({"template_api" => template_api}) }
 
-      it "should update settings" do
-        expect(plugin).to receive(:maximum_seen_major_version).at_least(:once).and_return(7)
-        described_class.add_ilm_settings_to_template(plugin, template)
-        expect(template['settings']['index.lifecycle.name']).not_to eq(nil)
-        expect(template['settings']['index.lifecycle.rollover_alias']).not_to eq(nil)
-        expect(template.include?('template')).to be_falsey
+      describe "with composable template API" do
+        let(:template_api) { "composable" }
+
+        it 'resolves composable index template API compatible setting' do
+          template = {}
+          described_class.resolve_template_settings(plugin, template)
+          expect(template["template"]["settings"]).not_to eq(nil)
+        end
+      end
+
+      describe "with legacy template API" do
+        let(:template_api) { "legacy" }
+
+        it 'resolves legacy index template API compatible setting' do
+          template = {}
+          described_class.resolve_template_settings(plugin, template)
+          expect(template["settings"]).not_to eq(nil)
+        end
+      end
+
+      describe "with `template_api => 'auto'`" do
+        let(:template_api) { "auto" }
+
+        describe "with ES < 8 versions" do
+
+          it 'resolves legacy index template API compatible setting' do
+            expect(plugin).to receive(:maximum_seen_major_version).at_least(:once).and_return(7)
+            template = {}
+            described_class.resolve_template_settings(plugin, template)
+            expect(template["settings"]).not_to eq(nil)
+          end
+        end
+
+        describe "with ES >= 8 versions" do
+          it 'resolves composable index template API compatible setting' do
+            expect(plugin).to receive(:maximum_seen_major_version).at_least(:once).and_return(8)
+            template = {}
+            described_class.resolve_template_settings(plugin, template)
+            expect(template["template"]["settings"]).not_to eq(nil)
+          end
+        end
       end
     end
   end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 11.13.1
+ - Fixes the failure when legacy template (`template_api=>'legacy'`) API is used with custom template, the plugin doesn't resolve ILM settings and crash. [#1102](https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/1102)
+
 ## 11.13.0
  - add technology preview support for allowing events to individually encode a default pipeline with `[@metadata][target_ingest_pipeline]` (as part of a technology preview, this feature may change without notice) [#1113](https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/1113)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,5 @@
 ## 11.13.1
- - Fixes the failure when legacy template (`template_api=>'legacy'`) API is used with custom template, the plugin doesn't resolve ILM settings and crash. [#1102](https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/1102)
+ - Avoid crash by ensuring ILM settings are injected in the correct location depending on the default (or custom) template format, template_api setting and ES version [#1102](https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/1102)
 
 ## 11.13.0
  - add technology preview support for allowing events to individually encode a default pipeline with `[@metadata][target_ingest_pipeline]` (as part of a technology preview, this feature may change without notice) [#1113](https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/1113)
--- a/logstash-output-elasticsearch.gemspec
+++ b/logstash-output-elasticsearch.gemspec
@@ -1,6 +1,6 @@
 Gem::Specification.new do |s|
   s.name            = 'logstash-output-elasticsearch'
-  s.version         = '11.13.0'
+  s.version         = '11.13.1'
   s.licenses        = ['apache-2.0']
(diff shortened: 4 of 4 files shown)


lib/logstash/outputs/elasticsearch/template_manager.rb
```diff
@@ -5,6 +5,7 @@ class TemplateManager
 
     # To be mixed into the elasticsearch plugin base
     def self.install_template(plugin)
+      puts "[MIK] install_template"
```
mashhurs, line 8: Oops!
```suggestion
```

lib/logstash/outputs/elasticsearch/template_manager.rb
```diff
@@ -19,9 +20,19 @@ def self.install_template(plugin)
...
         template = read_template_file(plugin.template)
       else
-        plugin.logger.info("Using a default mapping template", :es_version => plugin.maximum_seen_major_version,
+        es_major_version = plugin.maximum_seen_major_version
+
+        # for the v8+, LS shapes the template install request in a shape of _index_template
+        # however, when using `temaplte_api` == `legacy`, LS should shape the request
+        # as `_template` legacy structure requires
+        if plugin.template_api == 'legacy'
+          es_major_version = plugin.maximum_seen_major_version >= 8 ?
```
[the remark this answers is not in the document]
andsel: I would  simplify this as
```ruby
es_major_version = 7 if plugin.template_api == 'legacy' && plugin.maximum_seen_major_version >= 8
```

lib/logstash/outputs/elasticsearch/template_manager.rb
```diff
@@ -19,9 +20,19 @@ def self.install_template(plugin)
         plugin.logger.info("Using mapping template from", :path => plugin.template)
         template = read_template_file(plugin.template)
       else
-        plugin.logger.info("Using a default mapping template", :es_version => plugin.maximum_seen_major_version,
+        es_major_version = plugin.maximum_seen_major_version
+
+        # for the v8+, LS shapes the template install request in a shape of _index_template
+        # however, when using `temaplte_api` == `legacy`, LS should shape the request
```
andsel, line 26:
```suggestion
        # however, when using `template_api` == `legacy`, LS should shape the request
```

andsel:
I don't know why this change impact also the test with datastream and alias, maybe are not related.
However I think we should add an integration test to cover this use case, maybe a test that's enabled when running on `7.x` stack that check the template installation.

yaauie:
I have giant warning bells about exploding complexity with this change.

We already have [significant complexity in the matrix of templates we need to maintain](https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/main/Rakefile), and this change as-proposed only solves one possible edge-case (e.g., `template_api => composable` when connected to ES7 would attempt to push a non-composable template to the compostable template API). Extending this change-request to add a third dimension to the matrix of templates we maintain is costly from a maintenance/complexity standpoint.

I would much rather we make using the `template_api` _require_ that the user also provides a `template`, because doing so is less surprising than secretly using an ES7 template when we also provide ES8 templates.

mashhurs:
> I have giant warning bells about exploding complexity with this change.
> 
> We already have [significant complexity in the matrix of templates we need to maintain](https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/main/Rakefile), and this change as-proposed only solves one possible edge-case (e.g., `template_api => composable` when connected to ES7 would attempt to push a non-composable template to the compostable template API). Extending this change-request to add a third dimension to the matrix of templates we maintain is costly from a maintenance/complexity standpoint.
> 
> I would much rather we make using the `template_api` _require_ that the user also provides a `template`, because doing so is less surprising than secretly using an ES7 template when we also provide ES8 templates.

Thanks for @yaauie for good insights. I think we have to make a _clear_ behavior for _default template + ES version_ combination with our plugin since following edge cases are open:
1. `template_api` is `legacy` with ES8 (current behavior is to use legacy template API with ES8+`ecs_compatibility` template)
2. `template_api` is `composable` with ES7 (current behavior is to use index template API with ES7+`ecs_compatibility` template)

For the 1st case, as per [ES doc](https://www.elastic.co/guide/en/elasticsearch/reference/current/index-templates.html), it sounds to me we need to support default template as _legacy template_ is not yet deprecated.
For the 2nd case, I think we need to simply fail since ES7 doesn't even have a _index template_ (`_index_template` index)

Your opinion is great that simplifies and encourages users to know template details what they are doing and future changes. But my concern is, we seem not aligning with ES8. It seems ES8 (the attachment below with 8.5.3) has a legacy template API and default templates. Let's have a quick discussion when you get available to be on same page.

image

lib/logstash/outputs/elasticsearch/template_manager.rb
```diff
@@ -15,16 +15,24 @@ def self.install_template(plugin)
...
       if plugin.template
         plugin.logger.info("Using mapping template from", :path => plugin.template)
+        # we let customers to place ILM setting in their custom templates
         template = read_template_file(plugin.template)
       else
         plugin.logger.info("Using a default mapping template", :es_version => plugin.maximum_seen_major_version,
                                                                :ecs_compatibility => plugin.ecs_compatibility)
         template = load_default_template(plugin.maximum_seen_major_version, plugin.ecs_compatibility)
+        # we set ILM setting when we automatically resolve template
+        add_ilm_settings_to_template(plugin, template) if plugin.ilm_in_use?
```
mashhurs, line 33: We have [another logic to set ILM (elasticsearch.rb#L350)](https://github.com/mashhurs/logstash-output-elasticsearch/blob/0eff5a8b2bce2cf3f585a9fb508219b252b5be5c/lib/logstash/outputs/elasticsearch.rb#L350) and I wonder if we need to change this behavior as well. It seems to me we need to ignore ILM set process if custom template is provided.
```
setup_ilm if ilm_in_use? && !@template
```

lib/logstash/outputs/elasticsearch/template_manager.rb
```diff
@@ -15,16 +15,24 @@ def self.install_template(plugin)
                            "The legacy template API is slated for removal in Elasticsearch 9.")
       end
 
+      # we should avoid installing defaul legacy templates for ES8+ versions
+      # we need to guide customers using custom template or composable template API
+      if !plugin.template && plugin.template_api == 'legacy' && plugin.maximum_seen_major_version >= 8
+        raise LogStash::ConfigurationError, "Plugin cannot automatically resolve the default template when using legacy template API. Please consider using custom template or composable template API."
```
jsvd, line 21:
```suggestion
        raise LogStash::ConfigurationError, "Plugin cannot automatically resolve the default template when using legacy template API. Please consider using a custom template or 'template_api => composable'"
```

lib/logstash/outputs/elasticsearch/template_manager.rb
```diff
@@ -15,16 +15,24 @@ def self.install_template(plugin)
...
+
       if plugin.template
         plugin.logger.info("Using mapping template from", :path => plugin.template)
+        # we let customers to place ILM setting in their custom templates
         template = read_template_file(plugin.template)
       else
         plugin.logger.info("Using a default mapping template", :es_version => plugin.maximum_seen_major_version,
                                                                :ecs_compatibility => plugin.ecs_compatibility)
         template = load_default_template(plugin.maximum_seen_major_version, plugin.ecs_compatibility)
+        # we set ILM setting when we automatically resolve template
```
jsvd, line 32:
```suggestion
        # we set ILM setting ONLY when we automatically resolve template
```

mashhurs: accepted with new commit.

yaauie (changes requested):
- Breaks real-world configs by silently ignoring `ilm_*` settings when `template` is provided.
 - Guards only against specific combination of (`template_api=>legacy`) + (_with_ default template) + ( _when_ connected to ES8), when other combinations cause the same root issue (`template_api` is a _hint_ about a _provided_ `template`, not a _directive_)

lib/logstash/outputs/elasticsearch/template_manager.rb
```diff
@@ -15,16 +15,24 @@ def self.install_template(plugin)
                            "The legacy template API is slated for removal in Elasticsearch 9.")
       end
 
+      # we need to guide customers using their own (custom) template when they want to use legacy template API with Elasticsearch 8.x (or up) versions,
+      # otherwise `load_default_template` method loads the template based on ES 8.x versions which fits the `_index_template` index mapping but not `_template`.
+      if !plugin.template && plugin.template_api == 'legacy' && plugin.maximum_seen_major_version >= 8
```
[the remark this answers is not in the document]
mashhurs: Thanks Ry for the clarification!
Indeed the case is same when `template_api` is `composable` with ES 7.x versions. And we have cases where customer is using the ES 7.x versions and continuously upgrading es-output plugin.
I have applied the change which abstracts overall case. Please review it one more time.
Thanks.

lib/logstash/outputs/elasticsearch/template_manager.rb
```diff
@@ -15,16 +15,24 @@ def self.install_template(plugin)
...
       if plugin.template
         plugin.logger.info("Using mapping template from", :path => plugin.template)
+        # we let customers to place ILM setting in their custom templates
         template = read_template_file(plugin.template)
       else
         plugin.logger.info("Using a default mapping template", :es_version => plugin.maximum_seen_major_version,
                                                                :ecs_compatibility => plugin.ecs_compatibility)
         template = load_default_template(plugin.maximum_seen_major_version, plugin.ecs_compatibility)
+        # we set ILM setting ONLY when we automatically resolve template
+        add_ilm_settings_to_template(plugin, template) if plugin.ilm_in_use?
```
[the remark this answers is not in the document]
mashhurs: For the 1st and 2nd points, I was thinking to warn 
Related lines: [`elasticsearch.rb`:L350](https://github.com/mashhurs/logstash-output-elasticsearch/blob/c8f2089a5be68e464414c4a26663fb236d16200a/lib/logstash/outputs/elasticsearch.rb#L350) and  [`template_manager.rb`:L31~L34](https://github.com/mashhurs/logstash-output-elasticsearch/blob/c8f2089a5be68e464414c4a26663fb236d16200a/lib/logstash/outputs/elasticsearch/template_manager.rb#L31-L34)

Can you please provide more info about the 3rd point, what would be an ideal suggestion _manage their entire template in one place_?!

yaauie (changes requested):
We _cannot_ remove support for combining `ilm_*` settings with `template` in a patch release, silently or otherwise. While doing so technically prevents the specific situation that leads up to a crash, it _also breaks real-world configurations_ and violates our promise to our users to provide them with adequate deprecation windows so that they can resolve pending breaking changes independently of upgrades to our stack components.

Half-managed ILM/templates historically have worked, and our docs talk about ILM injection in a manner that makes no caveats for the _source_ of the template that they are injecting the ILM settings into. When ILM injection guesses the format of the template correctly, it works even if the user provided their own base template (e.g., `template_api => legacy` with a legacy-formatted template connected to ES 7; OR `template_api => composable` with a composable-formatted template connected to ES 8).

There are two issues at play, both with the newly-introduced `template_api` _hint_ about the format of the provided `template` (which allows the user to provide a template whose format doesn't match the favored format for the version of ES we are connected to):

1. `template_api` does not make sense in the absence of a user-provided `template` (it does not change how a _default_ template is resolved or generated, and making it do so was never part of its scope).
2. Even when a `template_api` hint is available, it is not used by the ILM injection code when it is guessing how to apply the ILM settings. Instead, it continues to switch based solely on which version of Elasticsearch it is connected to, and when it guesses incorrectly it crashes _opaquely_.

Therefore:
 - Providing the new `template_api` setting without _also_ providing a `template` should be a configuration error (e.g., ``a `template` is required when providing a `template_api` hint about how to handle a user-provided `template`; If you wish to use the #{template_api} template API, you need to provide your own #{template_api}-formatted template``).
 - ILM should _use_ the `template_api` hint when deciding how to apply the ILM settings.

Additionally, we know that half-managed templates with ILM are unnecessarily risky for our users, and the complexity is only going to continue to grow; therefore:
 - we should _deprecate_ (but continue to support) providing `ilm_*` settings along-side an explicit `template`, and should guide users to resolve the deprecation by adding their ILM settings to the template that they are providing (e.g., ``"Injecting ILM configuration into a custom template is deprecated and support for doing so will be removed in a future release; when managing your own template, please include your ILM configuration directly in the template you provide instead of using this plugin's `ilm_*` settings"``)
 - ILM injection should be audited to ensure we are as resilient as we can possibly be (e.g., tolerate missing intermediate nestings by creating them)
 - When ILM injection _does_ crash (e.g., if we need to set a key on a key/value map, but the value at that address is a scalar), we should provide clear messaging and helpful guidance (e.g., ``"failed to apply ILM settings to the provided template: #{e.message}; when managing your own template, we advise including your ILM configuration directly in the template you provide and avoiding this plugin's `ilm_*` settings"``)

lib/logstash/outputs/elasticsearch/template_manager.rb
```diff
@@ -46,15 +46,30 @@ def self.add_ilm_settings_to_template(plugin, template)
...
     end
 
-    def self.template_settings(plugin, template)
-      plugin.maximum_seen_major_version < 8 ? template['settings']: template['template']['settings']
+    def self.resolve_template_settings(plugin, template)
+      return composable_index_template_settings(template) if template.key?('template')
+      return legacy_index_template_settings(template) if template.key?('settings')
+      return template_endpoint(plugin) == INDEX_TEMPLATE_ENDPOINT ?
+               composable_index_template_settings(template) :
+               legacy_index_template_settings(template)
```
jsvd, line 61: There is a lot of logic and guesswork here and I'm a bit concerned with the lack of logging
```suggestion
      if template.key?('template')
        logger.trace("template_settings: settings location for template is under 'template' key", :template => template, :template_api => plugin.template_api, :es_version => plugin.maximum_seen_major_version)
        composable_index_template_settings(template)
      elsif template.key?('settings')
        logger.trace("template_settings: settings location for template is at object's root", :template => template, :template_api => plugin.template_api, :es_version => plugin.maximum_seen_major_version)
        legacy_index_template_settings(template) 
      else
        template_endpoint = template_endpoint(plugin)
        logger.trace("template_settings: template doesn't have 'settings' or 'template' fields, falling back to auto detection, :template => template, :template_api => plugin.template_api, :es_version => plugin.maximum_seen_major_version, :detected_endpoint => template_endpoint)
        template_endpoint == INDEX_TEMPLATE_ENDPOINT ?
               composable_index_template_settings(template) :
               legacy_index_template_settings(template)
     end
```

lib/logstash/outputs/elasticsearch/template_manager.rb
```diff
@@ -46,15 +46,30 @@ def self.add_ilm_settings_to_template(plugin, template)
...
+    end
+
+    # Returns (if exists) or creates _template API compatible template settings
```
mashhurs, line 64:
```suggestion
    # Sets ['settings'] field to be compatible with _template API structure
```

lib/logstash/outputs/elasticsearch/template_manager.rb
```diff
@@ -46,15 +46,30 @@ def self.add_ilm_settings_to_template(plugin, template)
...
+    def self.legacy_index_template_settings(template)
+      template['settings'] ||= {}
+    end
+
+    # Returns (if exists) or creates _index_template API compatible template settings
```
mashhurs, line 69:
```suggestion
    # Sets the ['template']['settings'] fields if not exist to be compatible with _index_template API structure
```

spec/unit/outputs/elasticsearch/template_manager_spec.rb
```diff
@@ -33,33 +33,83 @@
...
+          expect(template['settings']['index.lifecycle.rollover_alias']).not_to eq(nil)
+          expect(template.include?('template')).to be_falsey
+        end
       end
     end
 
-    describe "in version < 8" do
-      let(:file_path) { described_class.default_template_path(7) }
-      let(:template) { described_class.read_template_file(file_path)}
+    context "resolve template setting" do
```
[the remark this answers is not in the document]
mashhurs: These two cases are covered by L49 (when `template['template']` exists) & L62 (when `template['settings']` exists) lines. So didn't make double.

lib/logstash/outputs/elasticsearch/template_manager.rb
mashhurs: Added the logs between as well.

jsvd (approved):
LGTM, I've tested the multiple scenarios and it all seems to work correctly.

Issue #1103, opened 2022-12-21
DLQ error messages squashed in string instead of keeping structure when logging in json


andsel:
**Logstash information**:

Please include the following information:

1. Logstash version (e.g. `bin/logstash --version`) `>= 8.3.0`
2. Logstash installation source (e.g. built from source, with a package manager: DEB/RPM, expanded from tar or zip archive, docker)
3. How is Logstash being run (e.g. as a service/service manager: systemd, upstart, etc. Via command line, docker/kubernetes)
4. How was the Logstash Plugin installed `>= 11.9.0`

**JVM** (e.g. `java -version`):

If the affected version of Logstash is 7.9 (or earlier), or if it is NOT using the bundled JDK or using the 'no-jdk' version in 7.10 (or higher), please provide the following information:

1. JVM version (`java -version`)
2. JVM installation source (e.g. from the Operating System's package manager, from source, etc).
3. Value of the `JAVA_HOME` environment variable if set.

**OS version** (`uname -a` if on a Unix-like system):

**Description of the problem including expected versus actual behavior**:

**Steps to reproduce**:

Please include a *minimal* but *complete* recreation of the problem,
including (e.g.) pipeline definition(s), settings, locale, etc.  The easier
you make for us to reproduce it, the more likely that somebody will take the
time to look at it.

 1. Start en Elasticsearch with a closed `test_index` to generate http error code `400`.
 ```
 PUT test_index/
 POST test_index/_close
 ```
 2. Create pipeline with following config
 ```
input {
  generator {
    message => '{"name": "Andrea"}'
    count => 1
  }
}

output {
  elasticsearch {
    index => "test_index"
    hosts => "http://localhost:9200"
    user => "elastic"
    password => "changeme"
  }
}
```
 3. run Logstash with json format with a version `8.3.3` and `8.5.0`
 ```bash
> logstash-8.5.3/bin/logstash -f `pwd/`pipeline.conf --log.format json
> logstash-8.3.3/bin/logstash -f `pwd/`pipeline.conf --log.format json
 ```

**Provide logs (if relevant)**:

### In Logstash `8.3.3`the log is still JSON also for hash  map carried with the log line
```json
[...]

PR #1105, opened 2022-12-21
dlq-logging: reintroduce structured payload where possible

yaauie:
Alternative to #1104.

I authored this before getting derailed by a more urgent task, and just saw that @andsel had replicated my work in my off-hours. 

In this variant, because the `Elasticsearch#handle_event_mapping_errors` _already_ was making decisions based on the value of `@dlq_writer`, I didn't see the need to constrain the already-leaked DLQ abstraction to `Elasticsearch::Common#handle_dlq_status` and felt that working with `@dlq_writer` directly was more readable and easier to make sense of.

----

Fixes: #1103

### DIFF
--- a/lib/logstash/outputs/elasticsearch.rb
+++ b/lib/logstash/outputs/elasticsearch.rb
@@ -405,7 +405,7 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
 
     event_mapping_errors.each do |event_mapping_error|
       detailed_message = "#{event_mapping_error.message}; event: `#{event_mapping_error.event.to_hash_with_metadata}`"
-      handle_dlq_status(event_mapping_error.event, :warn, detailed_message)
+      @dlq_writer ? @dlq_writer.write(event_mapping_error.event, detailed_message) : @logger.warn(detailed_message)
     end
     @document_level_metrics.increment(:non_retryable_failures, event_mapping_errors.size)
   end
@@ -422,7 +422,7 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
         successful_events << @event_mapper.call(event)
       rescue EventMappingError => ie
         event_mapping_errors << FailedEventMapping.new(event, ie.message)
-       end
+      end
     end
     MapEventsResult.new(successful_events, event_mapping_errors)
   end
--- a/spec/es_spec_helper.rb
+++ b/spec/es_spec_helper.rb
@@ -67,19 +67,24 @@ module ESHelper
   end
 
   RSpec::Matchers.define :have_hits do |expected|
+    hits_count_path = ESHelper.es_version_satisfies?(">=7") ? %w(hits total value) : %w(hits total)
+
     match do |actual|
-      if ESHelper.es_version_satisfies?(">=7")
-        expected == actual['hits']['total']['value']
-      else
-        expected == actual['hits']['total']
-      end
+      @actual_hits_count = actual&.dig(*hits_count_path)
+      values_match? expected, @actual_hits_count
+    end
+    failure_message do |actual|
+      "expected that #{actual} with #{@actual_hits_count || "UNKNOWN" } hits would have #{expected} hits"
     end
   end
 
   RSpec::Matchers.define :have_index_pattern do |expected|
     match do |actual|
-      test_against = Array(actual['index_patterns'].nil? ? actual['template'] : actual['index_patterns'])
-      test_against.include?(expected)
+      @actual_index_pattterns = Array(actual['index_patterns'].nil? ? actual['template'] : actual['index_patterns'])
+      @actual_index_pattterns.any? { |v| values_match? expected, v }
+    end
+    failure_message do |actual|
+      "expected that #{actual} with index patterns #{@actual_index_pattterns} would have included `#{expected}`"
     end
   end
 
--- a/spec/integration/outputs/templates_spec.rb
+++ b/spec/integration/outputs/templates_spec.rb
@@ -23,7 +23,9 @@ describe "index template expected behavior", :integration => true do
 
     elasticsearch_client.indices.delete_template(:name => '*')
     # This can fail if there are no indexes, ignore failure.
-    elasticsearch_client.indices.delete(:index => '*') rescue nil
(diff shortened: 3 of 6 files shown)


lib/logstash/outputs/elasticsearch.rb
```diff
@@ -387,7 +387,7 @@ def handle_event_mapping_errors(event_mapping_errors)
 
     event_mapping_errors.each do |event_mapping_error|
       detailed_message = "#{event_mapping_error.message}; event: `#{event_mapping_error.event.to_hash_with_metadata}`"
-      handle_dlq_status(event_mapping_error.event, :warn, detailed_message)
+      @dlq_writer ? @dlq_writer.write(event_mapping_error.event, detailed_message) : @logger.warn(detailed_message)
```
andsel, line 390: In case of json log format, it logs a json document with `message` containing the event map converted to a string.
Is that desired or do you think it should keep document structure, nested in a `event` field, like you suggested in [comment](https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/1104#discussion_r1054681500)?

I like to keep things simple, without many moving parts, but this reintroduces a "code duplication"
```ruby
if @dlq_writer
  @dlq_writer.write
else 
    @logger.log
end    
```  

that the original #1084 tried to avoid.

spec/unit/outputs/elasticsearch_spec.rb
```diff
@@ -870,41 +870,63 @@
...
+                                                                          :response => mock_response))
           end
         end
 
         context 'when getting any other exception' do
           it 'should log at WARN level' do
-            logger = double("logger").as_null_object
-            subject.instance_variable_set(:@logger, logger)
-            expect(logger).to receive(:warn).with(a_string_including "Could not index event to Elasticsearch. status: some_status, action: [:action, :params, {")
+            # logger = double("logger").as_null_object
```
andsel, line 903: Remove commented code

spec/unit/outputs/elasticsearch_spec.rb
```diff
@@ -870,41 +870,63 @@
...
+                                                                         :response => mock_response))
           end
         end
 
         context 'when the response does not include [error]' do
           it 'should not fail, but just log a warning' do
-            logger = double("logger").as_null_object
-            subject.instance_variable_set(:@logger, logger)
-            expect(logger).to receive(:warn).with(a_string_including "Could not index event to Elasticsearch. status: some_status, action: [:action, :params, {")
+            # logger = double("logger").as_null_object
```
andsel, line 918: Remove commented code

andsel (approved):
LGTM

Left a couple of doubts, but I like the idea to keep it simple (this is more direct than #1104).
Nice to have the tests that verifies the structure to avoid future regressions like this.

lib/logstash/outputs/elasticsearch.rb
yaauie: The original logging for _dropped-events-when-DLQ-was-unavailable_ included the action tuple that was the cause of a rejection from Elasticsearch, but these events are being routed to DLQ because they _cannot_ generate a valid action tuple -- we only have our own pre-request rejection reason and the event itself.

While #1084 tried to avoid this particular DLQ-or-LOG duplication, the related #1080 _also_ introduced paths in this method that differ based on the availability of DLQ -- the abstraction has leaked already.

Looking at this again, it also doesn't make much sense for our `detailed_message` that is being added to the DLQ as _metadata-about-why-the-event-was-invalid_ to contain a stringified version of the entire event, since the entry itself will contain the entire event.

What about this (pseudo-code):

~~~ ruby
error_message, event = event_mapping_error.message, event_mapping_error.event
@dlq_writer ? @dlq_writer.write(event, error_message)
            : @logger.warn("Could not index event to Elasticsearch: #{error_message}", event: event.to_hash_with_metadata)
~~~

In this way, the DLQ entry would have a message like:
> ~~~
> "Badly formatted index, after interpolation still contains placeholder: [%{indexprefix}-foo]"
> ~~~
> ~~~
> "Elasticsearch doesn't support [DANCE] action"
> ~~~

And when the DLQ is disabled the logger would have entries like:
> ~~~
> {"message":"Could not index event to Elasticsearch: Badly formatted index, after interpolation still contains placeholder: [%{indexprefix}-foo]", "event": { ... } }
> ~~~
> ~~~
> {"message":"Could not index event to Elasticsearch: Elasticsearch doesn't support [DANCE] action", "event": {...} }
> ~~~

andsel: I like it, go for it :+1:

spec/es_spec_helper.rb
```diff
@@ -67,19 +67,24 @@ def self.es_version
...
     match do |actual|
-      if ESHelper.es_version_satisfies?(">=7")
-        expected == actual['hits']['total']['value']
-      else
-        expected == actual['hits']['total']
-      end
+      @actual_hits_count = actual&.dig(*hits_count_path)
+      values_match? expected, @actual_hits_count
+    end
+    failure_message do |actual|
```
yaauie, line 76: review note: the changes to the es_spec_helper make previously-flaky specs have more understandable causes, but those specs are no longer flaky :)

spec/integration/outputs/templates_spec.rb
```diff
@@ -23,7 +23,9 @@
 
     elasticsearch_client.indices.delete_template(:name => '*')
     # This can fail if there are no indexes, ignore failure.
-    elasticsearch_client.indices.delete(:index => '*') rescue nil
+    elasticsearch_client.indices.delete(:index => '*') rescue puts("DELETE INDICES ERROR: #{$!}")
+    # Since we are pinned to ES client 7.x, we need to delete data streams the hard way...
+    elasticsearch_client.perform_request("DELETE", "/_data_stream/*") rescue puts("DELETE DATA STREAMS ERROR: #{$!}")
```
yaauie, line 28: In ES 8.x, deleting indices doesn't delete data streams, which was causing some of the specs here to pull in more results than they were intending if-and-only-if they were executed after specs that had created data streams without cleaning up after themselves.

yaauie (commented):
Because the effective code changes were previously approved and specs are still green, I have updated the changelog and will merge.

PR #1107, opened 2023-01-09
Forces register phase to fail if bad credentials are provided

andsel:


andsel:
Closed by #1119

PR #1109, opened 2023-01-12
Improve log messages for data stream support checks

### LINKED ISSUE #1070: 'action' doesn't default to 'create' (as datastreams is default)
The documentation for the 'action' setting says the following:

"Default value is create for data streams, and index for non-time series data."

Starting from 8.0, datastreams is now the default, however, it seems that the 'action' setting is not being set by whether datastreams is enabled or disabled.

**Logstash information**:

Logstash 8.1.2

I would have expected that because datastreams is enabled by default, that I wouldn't have to specify the 'action' setting.

**Steps to reproduce**:

 1. Install Logstash and metricbeat
 2. Configure Logstash with:
```
 input {
  beats {
    port => 5044
  }
}
output {
  if [@metadata][pipeline] {
    elasticsearch {
      hosts => "http://localhost:9200/"
      pipeline => "%{[@metadata][pipeline]}"
      user => "elastic"
      password => "password"
      manage_template => false
      index => "%{[@metadata][beat]}-%{[@metadata][version]}"
    }
  } else {
    elasticsearch {
      hosts => "http://localhost:9200/"
      user => "elastic"
      password => "password"
      manage_template => false
      index => "%{[@metadata][beat]}-%{[@metadata][version]}"
    }
  }
}
``` 
 
 
  
```
:action=>["index",
[2022-04-21T08:41:13,559][WARN ][logstash.outputs.elasticsearch][main][d8e9b92d5d708958ec030423d27ccf66fb111900efd5ff96754c69edee4d7d80] Could not index event to Elasticsearch. {:status=>400, :action=>["index", {:_id=>nil, :_index=>"metricbeat-8.1.2", :routing=>nil}, 
....
:response=>{"index"=>{"_index"=>"metricbeat-8.1.2", "_id"=>nil, "status"=>400, "error"=>{"type"=>"illegal_argument_exception", "reason"=>"only write ops with an op_type of create are allowed in data streams"}}}}

```

edmocosta:
What this PR does?
- Refactored the [check_data_stream_config!](https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/1109/files#diff-56fc556089a712f2ab141b29aa8363f8aaf2f3ee65360c538cc42245a0647ca9R53) method to reduce the cyclomatic complexity. 
- Changed the [data_stream_default](https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/1109/files#diff-56fc556089a712f2ab141b29aa8363f8aaf2f3ee65360c538cc42245a0647ca9R161) log levels from `debug` to `info`. Those messages are helpful to users understanding the data stream auto-configuration mechanism/default and troubleshooting possibly incompatible settings.
- Logs were being printed twice due to a race condition on the `data_stream_config?` method, caused by the [register](https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/1109/files#diff-940291b716938425b715566d94fceb6a35ac05e51f901276637846a9abc77737) function.

Closes #1070

### DIFF
--- a/lib/logstash/outputs/elasticsearch.rb
+++ b/lib/logstash/outputs/elasticsearch.rb
@@ -300,6 +300,10 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
     # to build_client down to the Pool class.
     @client = build_client(LicenseChecker.new(@logger))
 
+    # Avoids race conditions in the @data_stream_config initialization (invoking check_data_stream_config! twice).
+    # It's being concurrently invoked by this register method and by the finish_register on the @after_successful_connection_thread
+    data_stream_enabled = data_stream_config?
+
     @after_successful_connection_thread = after_successful_connection do
       begin
         finish_register
@@ -324,7 +328,7 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
       raise LogStash::ConfigurationError, "DLQ feature (dlq_custom_codes) is configured while DLQ is not enabled" unless dlq_custom_codes.empty?
     end
 
-    if data_stream_config?
+    if data_stream_enabled
       @event_mapper = -> (e) { data_stream_event_action_tuple(e) }
       @event_target = -> (e) { data_stream_name(e) }
       @index = "#{data_stream_type}-#{data_stream_dataset}-#{data_stream_namespace}".freeze # default name
--- a/lib/logstash/outputs/elasticsearch/data_stream_support.rb
+++ b/lib/logstash/outputs/elasticsearch/data_stream_support.rb
@@ -28,7 +28,6 @@ module LogStash module Outputs class ElasticSearch
       base.extend(Validator)
     end
 
-    # @note assumes to be running AFTER {after_successful_connection} completed, due ES version checks
     def data_stream_config?
       @data_stream_config.nil? ? @data_stream_config = check_data_stream_config! : @data_stream_config
     end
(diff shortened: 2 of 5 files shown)


lib/logstash/outputs/elasticsearch/data_stream_support.rb
```diff
@@ -28,7 +28,6 @@ def self.included(base)
       base.extend(Validator)
     end
 
-    # @note assumes to be running AFTER {after_successful_connection} completed, due ES version checks
```
edmocosta, line 31: This comment looks to be true only for the [assert_es_version_supports_data_streams](https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/1109/files#diff-56fc556089a712f2ab141b29aa8363f8aaf2f3ee65360c538cc42245a0647ca9R144) method.

lib/logstash/outputs/elasticsearch.rb
```diff
@@ -300,6 +300,10 @@ def register
     # to build_client down to the Pool class.
     @client = build_client(LicenseChecker.new(@logger))
 
+    # Avoids race conditions in the @data_stream_config initialization (invoking check_data_stream_config! twice).
+    # It's being concurrently invoked by this register method and by the finish_register on the @after_successful_connection_thread
+    data_stream_enabled = data_stream_config?
```
andsel, line 305: A note on `#data_stream_config?` report that:
```
# @note assumes to be running AFTER {after_successful_connection} completed, due ES version checks
```

https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/v11.12.1/lib/logstash/outputs/elasticsearch/data_stream_support.rb#L31-L34

So here can't be invoked

lib/logstash/outputs/elasticsearch/data_stream_support.rb
```diff
@@ -50,48 +49,59 @@ def data_stream_name(event)
...
+    def check_enabled_data_stream_config!(params)
+      invalid_data_stream_params = invalid_data_stream_params(params)
+
+      if invalid_data_stream_params.any?
+        @logger.error "Invalid data stream configuration, the following parameters are not supported:", invalid_data_stream_params
+        raise LogStash::ConfigurationError, "Invalid data stream configuration: #{invalid_data_stream_params.keys}"
+      end
+
+      if ecs_compatibility == :disabled
+        if ::Gem::Version.create(LOGSTASH_VERSION) < ::Gem::Version.create(DATA_STREAMS_REQUIRES_ECS_LS_VERSION)
```
andsel, line 80: Since we are moving to a more assertion like checks, like:
```ruby
if ! condition
  raise Error
end
if ! another_condition
  raise AnotherError
end  
```

maybe would be best to invert the `if` condition and "cut" the `else`:

```ruby
if ::Gem::Version.create(LOGSTASH_VERSION) >= ::Gem::Version.create(DATA_STREAMS_REQUIRES_ECS_LS_VERSION)
  @logger.error "Invalid data stream configuration; `ecs_compatibility` must not be `disabled`. " + ENABLING_ECS_GUIDANCE
  raise LogStash::ConfigurationError, "Invalid data stream configuration: `ecs_compatibility => disabled`"
end

@deprecation_logger.deprecated "In a future release of Logstash, the Elasticsearch output plugin's `data_stream => true` will require the plugin to be run in ECS compatibility mode. " + ENABLING_ECS_GUIDANCE
```

andsel: `DATA_STREAMS_ENABLED_BY_DEFAULT_LS_VERSION` and `DATA_STREAMS_REQUIRES_ECS_LS_VERSION` both ECS and Datastream on by default was introduced with version `8.0.0` so maybe could collapsed in one constant and 

```
::Gem::Version.create(LOGSTASH_VERSION) >= ::Gem::Version.create(DATA_STREAMS_REQUIRES_ECS_LS_VERSION)
```

```
ds_default = ::Gem::Version.create(LOGSTASH_VERSION) >= ::Gem::Version.create(DATA_STREAMS_ENABLED_BY_DEFAULT_LS_VERSION)
```

could be extracted in a meaningful method. WDYT?

andsel (changes requested):
Nice refactoring.
I've left a doubt on the usage of `#data_stream_config?` because on the method signature is explicitly cited that has to be used after `#after_successful_connection` is invoked.

Left a couple of suggestions I would like to know your ideas

lib/logstash/outputs/elasticsearch.rb
edmocosta: Thanks for reviewing @andsel, Not sure If I missed something, but It seems it actually can (and is currently randomly doing it). that's what I tried to explain in my comment here: https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/1109#discussion_r1068362267

lib/logstash/outputs/elasticsearch/data_stream_support.rb
edmocosta: Yes, it sounds good to me to keep only one constant and extract this logic into a meaningful method! I tried to not touch the existing code logic to keep the refactoring as small as possible. There's also this [duplicate logic](https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/main/lib/logstash/outputs/elasticsearch/data_stream_support.rb#L233) `validate_dataset_identifier` & `validate_namespace_identifier` that could be collapsed into one.

If you think it would be okay to address those changes on this PR, I'd be glad to do so.

lib/logstash/outputs/elasticsearch.rb
andsel: Missed that sorry :-(
A couple of thoughts, the code comment seems to refer to the ES version retrieved during HTTP connection, but as I can see in the code that stays nested in the method `#check_data_stream_config!` (and method invoked inside this method), doesn't seem that it ever checks the version retrieved from the remote ES, to decide if enable or disable the datastream feature. It just checks Logstash version against some static constants.

What' effectively checks the live ES version, is in the `#finish_register` method the `assert_es_version_supports_data_streams`

```ruby
def finish_register
  assert_es_version_supports_data_streams if data_stream_config?
  ...
end  
```

So from my perspective we can remove the [code comment](https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/25aedb6e448997c6831839b97221959028917583/lib/logstash/outputs/elasticsearch/data_stream_support.rb#L31)  and keep your fix to avoid race on the `@data_stream_config` field between the "open connector" thread and the one that's executing the `#register`

edmocosta: Exactly! I actually already moved the comment to the [assert_es_version_supports_data_streams](https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/1109/files#diff-56fc556089a712f2ab141b29aa8363f8aaf2f3ee65360c538cc42245a0647ca9R144) method.

lib/logstash/outputs/elasticsearch/data_stream_support.rb
andsel: Maybe it's better to do these other changes in a follow PR, to keep the scope of this close to the original problem.

andsel (approved):
LGTM

jsvd (commented):
Thanks for the changes here.
My major concern in this PR is the lack of testing for the new validation logic.
We should have a few tests describing the new permutations of settings and the validations being added here.

edmocosta:
> Thanks for the changes here. My major concern in this PR is the lack of testing for the new validation logic. We should have a few tests describing the new permutations of settings and the validations being added here.

Hey @jsvd, thanks for reviewing! :) 

I'm probably missing something, but this PR should not be adding/changing any new validation logic. 
Except for the log messages text, It should behave exactly the same and should be already covered by the existing tests.
We did some refactoring on the `check_data_stream_config!` method, moving code around, and a small change suggested on this PR [thread](https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/1109#discussion_r1069150557), which should not affect the existing logic.

Could you please provide more details on that?

jsvd:
Fair enough, I just happened to be playing around with the PR and removed the check for settings incompatible with explicitly enabling data streams ([here](https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/1109/files#diff-56fc556089a712f2ab141b29aa8363f8aaf2f3ee65360c538cc42245a0647ca9R74-R77)), and didn't cause any tests to fail in the data_stream_support_spec.rb. That said I realize it's not an issue with your refactoring, it was a missing test case already.

Overall, after some extra manual testing I'm good with merging.

jsvd (approved):
LGTM

PR #1111, opened 2023-01-24
Change manage_template default value for data streams

### LINKED ISSUE #1093: Data Stream Settings expects `manage_template` to be false (not true) to avoid creating the ecs-logstash template
**Logstash information**:

Please include the following information:

1. Logstash version: logstash 8.4.3
2. Logstash installation source: tar
3. How is Logstash being run: command line
4. How was the Logstash Plugin installed: the plugin was shipped with the product

**JVM** (e.g. `java -version`): Builtin JVM 
> openjdk 17.0.4 2022-07-19
OpenJDK Runtime Environment Temurin-17.0.4+8 (build 17.0.4+8)

**OS version** (`uname -a` if on a Unix-like system): Linux LAPTOP-LFCO0LFS 5.10.102.1-microsoft-standard-WSL2 #1 SMP Wed Mar 2 00:30:59 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux 

The manage_template setting is not compatible with the data_stream settings. When data_stream settings are active and the Logstash Output Plugin is configured with a user without the `manage_index_template` cluster privilege it encounters 403 errors, because Logstash cannot load the Index Template ecs-logstash.

In my expectation when data_stream is used this template should not be loaded.

I dig a little bit in the source code and I think this [line](https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/7c24cfa4c3ef76078a77f0d317038b2c7e0c44db/lib/logstash/outputs/elasticsearch/data_stream_support.rb#L111) may explain the unexpected behavior: the comment state that [manage_template](https://www.elastic.co/guide/en/logstash/current/plugins-outputs-elasticsearch.html#plugins-outputs-elasticsearch-manage_template) is false by default, but the documentation (and also my experience) shows that manage_template is true by default.

**Steps to reproduce**:

 1. Create a user in Elasticsearch with the monitor cluster privilege and the auto_configure,create index privileges for logs-*
 1. Make sure there is no ecs-logstash index template in the Elasticsearch Cluster (the monitor cluster privilege allows to read the index_template, thus if there is already an index_template named ecs-logstash, the plugin will not try to load it again)
 1. Configure the following pipeline:
 ```ruby
input { generator {} }

filter { sleep { time => 1 } }

output {
       elasticsearch {
         user => "logstash_writer"
         password => "XXXX"
         data_stream => "true"
         data_stream_dataset => "apache.access"
         data_stream_namespace => "default"
         data_stream_type => "logs"
         hosts => "https://XXXXX:9200"
       }
}
 ```
4. It should immediately fail with the error reported in the next section of the issue
 
```json
[2022-10-19T07:38:51,515][ERROR][logstash.outputs.elasticsearch][another_test] Failed to install template {:message=>"Got response code '403' contacting Elasticsearch at URL 'https://test-fleet-12d4a4.es.us-central1.gcp.cloud.es.io:443/_index_template/ecs-logstash'", :exception=>LogStash::Outputs::ElasticSearch::HttpClient::Pool::BadResponseCodeError, :backtrace=>["/home/mirko/logstash-8.4.3/vendor/bundle/jruby/2.6.0/gems/logstash-output-elasticsearch-11.6.0-java/lib/logstash/outputs/elasticsearch/http_client/manticore_adapter.rb:84:in `perform_request'", "/home/mirko/logstash-8.4.3/vendor/bundle/jruby/2.6.0/gems/logstash-output-elasticsearch-11.6.0-java/lib/logstash/outputs/elasticsearch/http_client/pool.rb:324:in `perform_request_to_url'", "/home/mirko/logstash-8.4.3/vendor/bundle/jruby/2.6.0/gems/logstash-output-elasticsearch-11.6.0-java/lib/logstash/outputs/elasticsearch/http_client/pool.rb:311:in `block in perform_request'", "/home/mirko/logstash-8.4.3/vendor/bundle/jruby/2.6.0/gems/logstash-output-elasticsearch-11.6.0-java/lib/logstash/outputs/elasticsearch/http_client/pool.rb:398:in `with_connection'", "/home/mirko/logstash-8.4.3/vendor/bundle/jruby/2.6.0/gems/logstash-output-elasticsearch-11.6.0-java/lib/logstash/outputs/elasticsearch/http_client/pool.rb:310:in `perform_request'", "/home/mirko/logstash-8.4.3/vendor/bundle/jruby/2.6.0/gems/logstash-output-elasticsearch-11.6.0-java/lib/logstash/outputs/elasticsearch/http_client/pool.rb:318:in `block in Pool'", "/home/mirko/logstash-8.4.3/vendor/bundle/jruby/2.6.0/gems/logstash-output-elasticsearch-11.6.0-java/lib/logstash/outputs/elasticsearch/http_client.rb:408:in `template_put'", "/home/mirko/logstash-8.4.3/vendor/bundle/jruby/2.6.0/gems/logstash-output-elasticsearch-11.6.0-java/lib/logstash/outputs/elasticsearch/http_client.rb:85:in `template_install'", "/home/mirko/logstash-8.4.3/vendor/bundle/jruby/2.6.0/gems/logstash-output-elasticsearch-11.6.0-java/lib/logstash/outputs/elasticsearch/template_manager.rb:29:in `install'", "/home/mirko/logstash-8.4.3/vendor/bundle/jruby/2.6.0/gems/logstash-output-elasticsearch-11.6.0-java/lib/logstash/outputs/elasticsearch/template_manager.rb:17:in `install_template'", "/home/mirko/logstash-8.4.3/vendor/bundle/jruby/2.6.0/gems/logstash-output-elasticsearch-11.6.0-java/lib/logstash/outputs/elasticsearch.rb:494:in `install_template'", "/home/mirko/logstash-8.4.3/vendor/bundle/jruby/2.6.0/gems/logstash-output-elasticsearch-11.6.0-java/lib/logstash/outputs/elasticsearch.rb:318:in `finish_register'", "/home/mirko/logstash-8.4.3/vendor/bundle/jruby/2.6.0/gems/logstash-output-elasticsearch-11.6.0-java/lib/logstash/outputs/elasticsearch.rb:283:in `block in register'", "/home/mirko/logstash-8.4.3/vendor/bundle/jruby/2.6.0/gems/logstash-output-elasticsearch-11.6.0-java/lib/logstash/plugin_mixins/elasticsearch/common.rb:154:in `block in after_successful_connection'"]}
```

**Possible Related Issues**
- #1071
- #1070

edmocosta:
What this PR does?

- Changed the `manage_template` default value to `false` when data stream is enabled or the `auto` mechanism resolves to true.
- Added the `manage_template => false` as a valid data stream option, that way, explicitly disabling the template manager will not consider the configuration as data stream incompatible.

---
Closes #1093

### DIFF
--- a/lib/logstash/outputs/elasticsearch.rb
+++ b/lib/logstash/outputs/elasticsearch.rb
@@ -158,7 +158,7 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
   # control over template creation, (e.g. creating indices dynamically based on
   # field names) you should set `manage_template` to false and use the REST
   # API to apply your templates manually.
-  config :manage_template, :validate => :boolean, :default => true
+  config :manage_template, :validate => :boolean # :default => true unless data_stream_config?
 
   # This configuration option defines how the template is named inside Elasticsearch.
   # Note that if you have used the template management features and subsequently
@@ -303,6 +303,8 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
     # Avoids race conditions in the @data_stream_config initialization (invoking check_data_stream_config! twice).
     # It's being concurrently invoked by this register method and by the finish_register on the @after_successful_connection_thread
     data_stream_enabled = data_stream_config?
+
+    setup_template_manager_defaults(data_stream_enabled)
 
     @after_successful_connection_thread = after_successful_connection do
       begin
@@ -606,6 +608,12 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
     @index ||= default_index
     @ilm_rollover_alias ||= default_ilm_rollover_alias
     @template_name ||= default_template_name
+  end
+
+  def setup_template_manager_defaults(data_stream_enabled)
+    if @manage_template.nil? # not set
+      @manage_template = !data_stream_enabled
+    end
   end
 
   # To be overidden by the -java version
--- a/spec/unit/outputs/elasticsearch/data_stream_support_spec.rb
+++ b/spec/unit/outputs/elasticsearch/data_stream_support_spec.rb
@@ -152,6 +152,42 @@ describe LogStash::Outputs::ElasticSearch::DataStreamSupport do
 
   end
 
+  context 'ds value-dependent configuration' do
+    # Valid settings values
+    let(:options) { super().merge(
+      'action' => 'create',
+      'routing' => 'any',
+      'pipeline' => 'any',
+      'manage_template' => "false",
+      'data_stream' => 'true')
+    }
+
+    context 'with valid values' do
+      let(:options) { super().merge(
+        'data_stream_type' => 'logs',
+        'data_stream_dataset' => 'any',
+        'data_stream_namespace' => 'any',
+        'data_stream_sync_fields' => true,
+        'data_stream_auto_routing' => true)
+      }
+
+      it 'should enable data-streams by default' do
+        expect ( subject.data_stream_config? ).to be true
+      end
+    end
+
+    context 'with invalid values' do
+      let(:options) { super().merge(
+        'action' => 'index',
+        'manage_template' => 'true')
+      }
+
+      it 'should raise a configuration error' do
+        expect { subject.data_stream_config? }.to raise_error(LogStash::ConfigurationError, 'Invalid data stream configuration: ["action", "manage_template"]')
+      end
+    end
+  end
+
   context "default (non data-stream) configuration (on 7.x)" do
 
     let(:options) do
--- a/spec/unit/outputs/elasticsearch_spec.rb
+++ b/spec/unit/outputs/elasticsearch_spec.rb
@@ -606,6 +606,26 @@ describe LogStash::Outputs::ElasticSearch do
     end
   end
 
+  describe "the manage_template option" do
+    context "with data stream enabled" do
+      let(:options) { {"data_stream" => "true", "data_stream_type" => "logs" } }
+      let(:do_register) { true }
+
+      it "should default to false" do
+        expect(subject).to have_attributes(manage_template: false)
+      end
+    end
+
+    context "with data stream disabled" do
+      let(:options) { {"data_stream" => "false", "index" => "logs" } }
+      let(:do_register) { true }
+
+      it "should default to true" do
+        expect(subject).to have_attributes(manage_template: true)
+      end
+    end
+  end
+
   describe "SSL end to end" do
     let(:do_register) { false } # skip the register in the global before block, as is called here.
 
@@ -870,7 +890,7 @@ describe LogStash::Outputs::ElasticSearch do
   end if LOGSTASH_VERSION > '6.0'
 
   context 'handling elasticsearch document-level status meant for the DLQ' do
-    let(:options) { { "manage_template" => false } }
+    let(:options) { { "manage_template" => false, "data_stream" => 'false' } }
     let(:action) { LogStash::Outputs::ElasticSearch::EventActionTuple.new(:action, :params, LogStash::Event.new("foo" => "bar")) }
 
     context 'when @dlq_writer is nil' do
@@ -1140,7 +1160,7 @@ describe LogStash::Outputs::ElasticSearch do
   describe "post-register ES setup" do
     let(:do_register) { false }
     let(:es_version) { '7.10.0' } # DS default on LS 8.x
-    let(:options) { { 'hosts' => '127.0.0.1:9999' } }
+    let(:options) { { 'hosts' => '127.0.0.1:9999', 'data_stream' => 'false' } }
     let(:logger) { subject.logger }
 
     before do
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 11.12.4
+ - Changed the `manage_template` default value to `false` when data streams is enabled [#1111](https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/1111)
+    -  Added the `manage_template => false` as a valid data stream option
+
 ## 11.12.3
 - Changed the log messages for data stream checks [#1109](https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/1109)
     - Added more details about incompatible data streams supplied configurations
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -790,7 +790,7 @@ Set the keystore password
 ===== `manage_template`
 
   * Value type is <<boolean,boolean>>
-  * Default value is `true`
+  * Default value is `true` for non-time series data, and `false` for data streams.
 
 From Logstash 1.3 onwards, a template is applied to Elasticsearch during
 Logstash's startup if one with the name <<plugins-{type}s-{plugin}-template_name>>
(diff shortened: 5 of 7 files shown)


lib/logstash/outputs/elasticsearch.rb
```diff
@@ -158,7 +158,7 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
   # control over template creation, (e.g. creating indices dynamically based on
   # field names) you should set `manage_template` to false and use the REST
   # API to apply your templates manually.
-  config :manage_template, :validate => :boolean, :default => true
+  config :manage_template, :validate => :boolean # :default => true unless data_stream_config?
```
andsel, line 161: I would add this information as part of the comment instead of commenting part of the line
```suggestion
# Is false for data streams else it's true
  config :manage_template, :validate => :boolean
```

spec/unit/outputs/elasticsearch/data_stream_support_spec.rb
```diff
@@ -152,6 +152,42 @@
 
   end
 
+  context 'ds value-dependent configuration' do
+    # Valid settings values
+    let(:options) { super().merge(
+      'action' => 'create',
+      'routing' => 'any',
+      'pipeline' => 'any',
```
andsel, line 160: Are `routing` and `pipeline` required for the the test?

spec/unit/outputs/elasticsearch/data_stream_support_spec.rb
```diff
@@ -152,6 +152,42 @@
...
+      'manage_template' => 'false',
+      'data_stream' => 'true')
+    }
+
+    context 'with valid values' do
+      let(:options) { super().merge(
+        'data_stream_type' => 'logs',
+        'data_stream_dataset' => 'any',
```
andsel, line 168: Are all of these necessary for the functionality under test?

spec/unit/outputs/elasticsearch_spec.rb
```diff
@@ -870,7 +890,7 @@
   end if LOGSTASH_VERSION > '6.0'
 
   context 'handling elasticsearch document-level status meant for the DLQ' do
-    let(:options) { { "manage_template" => false } }
```
andsel, line 873: Maybe is best to use the `false` instead of `'false'` to keep the same pattern used in `manage_template`

spec/unit/outputs/elasticsearch_spec.rb
```diff
@@ -1140,7 +1160,7 @@
   describe "post-register ES setup" do
     let(:do_register) { false }
     let(:es_version) { '7.10.0' } # DS default on LS 8.x
-    let(:options) { { 'hosts' => '127.0.0.1:9999' } }
+    let(:options) { { 'hosts' => '127.0.0.1:9999', 'data_stream' => 'false' } }
```
andsel, line 1163:
```suggestion
    let(:options) { { 'hosts' => '127.0.0.1:9999', 'data_stream' => false } }
```

andsel (changes requested):
Left a couple of notes on the tests

spec/unit/outputs/elasticsearch/data_stream_support_spec.rb
edmocosta: I'd say yes, this test is checking the `invalid_data_stream_params` method conditions (which is currently uncovered),
Having `routing` and `pipeline` here will ensure we do not disable the data stream if any of them is set on the plugin configuration.

spec/unit/outputs/elasticsearch/data_stream_support_spec.rb
edmocosta: Not really, testing only one would be enough as the `invalid_data_stream_params` method is only checking if the name prefix starts with `data_stream_*`. I ended up adding all valid `data_stream_` options here as a safeguard. Do you think we should keep only one?

spec/unit/outputs/elasticsearch_spec.rb
edmocosta: For some reason, several tests fail on this spec if I set it as a boolean instead of a string! It looks the other tests are doing the same! 🤷

spec/unit/outputs/elasticsearch/data_stream_support_spec.rb
andsel: oks

spec/unit/outputs/elasticsearch/data_stream_support_spec.rb
andsel: So maybe the name of the `context` could be `when any data_stream option has a valid value` instead of `with valid values`

spec/unit/outputs/elasticsearch_spec.rb
andsel: Uhhh, ok would be nice to understand why, but however is out of context of this PR

spec/unit/outputs/elasticsearch/data_stream_support_spec.rb
edmocosta: It's testing all possible valid values, not only the `data_stream` ones. It's merging the `data_stream_*` values with the parent context `let(:options) { 'action' => ..}` values.
Giving it a second thought, it's actually confusing, maybe it would be better to remove this parent `let(:options)` and move everything into the `with valid values` context. WDYT?

spec/unit/outputs/elasticsearch_spec.rb
edmocosta: Same as https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/1111#discussion_r1086724547

andsel: oks

andsel (approved):
LGTM

spec/unit/outputs/elasticsearch_spec.rb
jsvd: the setting is a string that can have values of 'true', 'false', and 'auto', so it makes sense to pass a string here too.

lib/logstash/outputs/elasticsearch.rb
```diff
@@ -608,6 +612,12 @@ def setup_ecs_compatibility_related_defaults
     @template_name ||= default_template_name
   end
 
+  def setup_template_manager_defaults(data_stream_enabled)
+    if @manage_template.nil? # not set
+      @manage_template = !data_stream_enabled
```
jsvd, line 617:
```suggestion
+    if original_params["manage_template"].nil? && data_stream_enabled
+      logger.debug("Disabling template management since data streams are enabled")
+      @manage_template = false
```

lib/logstash/outputs/elasticsearch.rb
```diff
@@ -158,7 +158,9 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
   # control over template creation, (e.g. creating indices dynamically based on
   # field names) you should set `manage_template` to false and use the REST
   # API to apply your templates manually.
-  config :manage_template, :validate => :boolean, :default => true
+  #
+  # Default value is `true` for non-time series data, and `false` for data streams.
+  config :manage_template, :validate => :boolean
```
jsvd, line 163:
```suggestion
  #
  # Default value is `true` unless data streams is enabled
  config :manage_template, :validate => :boolean, :default => true
```

jsvd (changes requested):
I've suggested an alternative means to implement disabling template management when data streams are enabled.
In practice the result is the same, but constraining the change of `@manage_template` only when "user not setting a manage_template value + data streams being enabling", makes it easier to reason about the impact.
Also gives us an opportunity to log an entry to inform of this change in the default.

roaksoax (approved):
Given that the final review changes were simply to adopt suggestions, now that they are adopted, I'm approving this since Joao is out today!

Issue #1112, opened 2023-01-30
Allow events to directly encode the downstream pipeline to use.


yaauie:
When `pipeline` parameter for this plugin is unset and an event has a `[@metadata][target_ingest_pipeline]` field containing a string value, use that value in the event tuple's pipeline.

Note: we _cannot_ implement this as a default value for the `pipeline` parameter because the field's _absence_ would cause us to send the uninterpolated `%{[@metadata][target_ingest_pipeline]}` _as_ the pipeline name.

Instead, we must interpolate the _result_ of a non-nil (`@pipeline || event.get("[@metadata][target_ingest_pipeline]")&.to_s)

Likely solution:

~~~ diff
diff --git a/lib/logstash/outputs/elasticsearch.rb b/lib/logstash/outputs/elasticsearch.rb
index 0da3c90..93e486e 100644
--- a/lib/logstash/outputs/elasticsearch.rb
+++ b/lib/logstash/outputs/elasticsearch.rb
@@ -516,19 +516,23 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
         routing_field_name => @routing ? event.sprintf(@routing) : nil
     }
 
-    if @pipeline
-      value = event.sprintf(@pipeline)
-      # convention: empty string equates to not using a pipeline
-      # this is useful when using a field reference in the pipeline setting, e.g.
-      #      elasticsearch {
-      #        pipeline => "%{[@metadata][pipeline]}"
-      #      }
-      params[:pipeline] = value unless value.empty?
-    end
+    target_pipeline = resolve_pipeline(event)
+    # convention: empty string equates to not using a pipeline
+    # this is useful when using a field reference in the pipeline setting, e.g.
+    #      elasticsearch {
+    #        pipeline => "%{[@metadata][pipeline]}"
+    #      }
+    params[:pipeline] = target_pipeline unless target_pipeline.empty?
 
     params
   end
 
+  def resolve_pipeline(event)
+    pipeline_template = @pipeline || event.get("[@metadata][target_ingest_pipeline]")&.to_s
+
+    pipeline_template && event.sprintf(pipeline_template)
+  end
+
   @@plugins = Gem::Specification.find_all{|spec| spec.name =~ /logstash-output-elasticsearch-/ }
 
   @@plugins.each do |plugin|

~~~

PR #1113, opened 2023-01-31
Add support for pipeline to decode from metadata target_ingest_pipeline

kaisecheng:
Fixes: #1112

This PR allows `pipeline` option to take value from `[@metadata][target_ingest_pipeline]` if `pipeline` is not set.
if `[@metadata][target_ingest_pipeline]` gives empty string, the `pipeline` won't be set

### DIFF
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -849,7 +849,7 @@ not also set this field. That will raise an error at startup
 ===== `pipeline`
 
   * Value type is <<string,string>>
-  * Default value is `nil`
+  * There is no default value.
 
 Set which ingest pipeline you wish to execute for an event. You can also use
 event dependent configuration here like `pipeline => "%{[@metadata][pipeline]}"`.
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -853,7 +853,6 @@ not also set this field. That will raise an error at startup
 
 Set which ingest pipeline you wish to execute for an event. You can also use
 event dependent configuration here like `pipeline => "%{[@metadata][pipeline]}"`.
-If `pipeline` is not set, the plugin takes the value from `[@metadata][target_ingest_pipeline]`.
 The pipeline parameter won't be set if the value resolves to empty string ("").
 
 [id="plugins-{type}s-{plugin}-pool_max"]
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -852,9 +852,9 @@ not also set this field. That will raise an error at startup
   * Default value is `nil`
 
 Set which ingest pipeline you wish to execute for an event. You can also use
-event dependent configuration here like `pipeline =>
-"%{[@metadata][pipeline]}"`. The pipeline parameter won't be set if the value
-resolves to empty string ("").
+event dependent configuration here like `pipeline => "%{[@metadata][pipeline]}"`.
+If `pipeline` is not set, the plugin takes the value from `[@metadata][target_ingest_pipeline]`.
+The pipeline parameter won't be set if the value resolves to empty string ("").
 
 [id="plugins-{type}s-{plugin}-pool_max"]
 ===== `pool_max`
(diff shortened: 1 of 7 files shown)


spec/unit/outputs/elasticsearch_spec.rb
```diff
@@ -604,6 +604,48 @@
...
       end
     end
+
+    context "with both pipeline and target_ingest_pipeline" do
+      let(:options) { {"pipeline" => "%{pipeline}" } }
+      let(:event) { LogStash::Event.new({"pipeline" => "my-ingest-pipeline", "[@metadata][target_ingest_pipeline]" => "meta-ingest-pipeline"}) }
+
+      it "should interpolate the pipeline value and set it" do
+        expect(subject.send(:event_action_tuple, event)[1]).to include(:pipeline => "my-ingest-pipeline")
+      end
```
yaauie, line 614: nit: we certainly have a mix, but since this is a spec, using _declarative_ language about what this plugin _does_ is more clear than using language that says what it _should do_ ([more in the ruby style guide](https://github.com/rubocop/rspec-style-guide#should-in-example-docstrings)).

also, we should catch the case where the plugin's `pipeline` is specified but it doesn't get sprintf expanded, something like:

```suggestion
      it "interpolates the plugin's pipeline value" do
        expect(subject.send(:event_action_tuple, event)[1]).to include(:pipeline => "my-ingest-pipeline")
      end
      context "when the plugin's `pipeline` is constant" do
        let(:options) { super().merge("pipeline" => "my-constant-pipeline") }
         it "uses plugin's pipeline value" do
          expect(subject.send(:event_action_tuple, event)[1]).to include(:pipeline => "my-constant-pipeline")
        end
      end
      context "when the plugin's `pipeline` includes an unresolvable sprintf placeholder" do
        let(:options) { super().merge("pipeline" => "reference-%{unset}-field") }
        it "does not use the target_ingest_pipeline" do
          # when sprintf doesn't resolve a placeholder, the behaviour of our `pipeline` is UNSPECIFIED.
          # here we only validate that the presence of the magic field does not
          # override an explicitly-configured pipeline.
          expect(subject.send(:event_action_tuple, event)[1]).to_not include(:pipeline => "my-ingest-pipeline")
        end
      end
```

docs/index.asciidoc
```diff
@@ -852,9 +852,9 @@ not also set this field. That will raise an error at startup
   * Default value is `nil`
 
 Set which ingest pipeline you wish to execute for an event. You can also use
-event dependent configuration here like `pipeline =>
-"%{[@metadata][pipeline]}"`. The pipeline parameter won't be set if the value
-resolves to empty string ("").
+event dependent configuration here like `pipeline => "%{[@metadata][pipeline]}"`.
+If `pipeline` is not set, the plugin takes the value from `[@metadata][target_ingest_pipeline]`.
+The pipeline parameter won't be set if the value resolves to empty string ("").
```
yaauie, line 857: Since this is Tech Preview, I am wary of injecting it directly into the docs but I like your other improvements.

```suggestion
event dependent configuration here like `pipeline => "%{[@metadata][pipeline]}"`.
The pipeline parameter won't be set if the value resolves to empty string ("").
```

docs/index.asciidoc
```diff
@@ -852,9 +852,9 @@ not also set this field. That will raise an error at startup
   * Default value is `nil`
```
yaauie, line 852: `nil` is an implementation detail that we don't need to leak to the users.

```suggestion
  * There is no default value.
```

kaisecheng:
All test cases run successfully but one of the unit test [stages](https://app.travis-ci.com/github/logstash-plugins/logstash-output-elasticsearch/jobs/594742778) is frozen after tests are done.

yaauie (approved):
Looks to work as-advertised. Unsure about the failing spec and can't trigger Travis builds for some reason, will attempt to replicate when I get home in a few hours.

PR #1117, opened 2023-02-10
Fix termination deadlock when the output is retrying whole failed bulk requests

pfcoperez:
Closes: https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/1116
Detailed explanation of the problem: https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/1116#issuecomment-1424343607

Break the termination deadlock when whole bulk requests are failing in a retry loop during pipeline terminations

### DIFF
--- a/lib/logstash/plugin_mixins/elasticsearch/common.rb
+++ b/lib/logstash/plugin_mixins/elasticsearch/common.rb
@@ -155,7 +155,7 @@ module LogStash; module PluginMixins; module ElasticSearch
     end
 
     def shutting_down?
-      @stopping.true? || (!execution_context.pipeline.nil? && execution_context.pipeline.shutdown_requested? && !execution_context.pipeline.worker_threads_draining?)
+      @stopping.true? || (!execution_context.nil? && !execution_context.pipeline.nil? && execution_context.pipeline.shutdown_requested? && !execution_context.pipeline.worker_threads_draining?)
     end
 
     # launch a thread that waits for an initial successful connection to the ES cluster to call the given block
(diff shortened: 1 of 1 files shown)


yaauie (commented):
If we return without raising (and crashing the plugin/worker),  the subsequent normal-completion of a batch that was sourced from the PQ will cause it to be acknowledged and marked eligible for immediate deletion, causing a restarted pipeline to NOT include the batch. I consider to be a data-loss regression, even though it is less-bad for the memory-queue case, or for cases where some amount of data-loss is preferred if it is the cost of keeping things moving.

The current behaviour is an unfortunate deadlock, true, but a force-shutdown followed by a restart currently will re-emit the unacknowledged batch for reprocessing. 

Should we instead raise an exception that crashes the worker and prevents the batch from being acknowledged?

pfcoperez:
Thanks a ton for chiming in @yaauie ! 

> Should we instead raise an exception that crashes the worker and prevents the batch from being acknowledged?

Do you mean that we could replace the changed `retry unless shutting_down?` lines with an if that would raise an exception when `shutting_down?` is true?

lib/logstash/plugin_mixins/elasticsearch/common.rb
```diff
@@ -154,12 +154,17 @@ def successful_connection?
       !!maximum_seen_major_version && alive_urls_count > 0
     end
 
+    def shutting_down?
+      @stopping.true? || (!execution_context.nil? && !execution_context.pipeline.nil? && execution_context.pipeline.shutdown_requested? && !execution_context.pipeline.worker_threads_draining?)
```
andsel, line 158: To avoid the cascading `nil` checks we could use the `&.` safe navigation operator. That would simplify as:
```suggestion
      @stopping.true? || (execution_context&.pipeline&.shutdown_requested? && !execution_context&.pipeline&.worker_threads_draining?)
```

pfcoperez:
Superseded by https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/1119

Issue #676, opened 2017-09-20
Enhancement: support PEM format for client certificates


hex2a:
For the certificate authority the `cacert` (PEM) or `truststore` parameter can be used,
for client certificates there is only the `keystore` is available, which only accepts .jks or .p12 files.
Please implement support for PEM format for client certificates. (similar to the CA config parameter)

wkjun:
a long time has pass,this issue can be solved?

Issue #672, opened 2017-09-13
Logstash “Elasticsearch” output plugin does not support SSL client certificates


bozzerreaneaux:
Plugin version: v7.4.0
Released on: 2017-08-21

Hello team,

This is more of a feature request and not an issue..

I've noticed the Elasticsearch output plugin does not offer configuration for client certificate so that the Logstash pipeline can also authenticate with a client certificate.

Here's my case: I am evaluating X-Pack and PKI realms for a centralized logging POC solution for our company. My intention was to implement mutual authentication between beats and Logstash and between Logstash and Elasticsearch .

I would like to have separate Beats and Logstash instance per each application server so that I can later configure role based access for each Logstash pipeline and isolate the data from different applications(projects)

I was surprised to learn that the elasticsearch output configuration doesn't offer client certificate options but only options to configure certification authority (either cacert path or in java truststore) to verify the Elasticsearch server cert.

Is this something that I can possibly implement by editing  \vendor\bundle\jruby\1.9\gems\logstash-output-elasticsearch-7.3.8-java\lib\logstash\outputs\elasticsearch\**http_client_builder.rb** or any other file?

Nrezinorn:
+1 for this features please.  I am fumbling trying to get this working on a dev machine myself.

jordansissel:
We will work on this feature. No ETA is available. No further +1s are
necessary.

hex2a:
according to the documentation, this should be possible using the `keystore` parameter?

https://www.elastic.co/guide/en/logstash/current/plugins-outputs-elasticsearch.html#plugins-outputs-elasticsearch-keystore

sadok-f:
+1

jsvd:
closing as duplicate of https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/672

carldunham:
Wait, this IS #672

berndvdveen:
Is it possible to give an ETA for the implementation of this feature?

Issue #404, opened 2016-04-04
Use consistent naming for ca_file/cacert parameters across Elasticsearch input/output plugins


joshuar:
In this output, we use the parameter `cacert` for specifying a path to a PEM encoded file holding the CA cert for an SSL-secured ES cluster while we use `ca_file` in the output plugin.  As these are the same parameters, they should have the same name.

See also https://github.com/logstash-plugins/logstash-input-elasticsearch/issues/42

PR #1118, opened 2023-02-28
Standardize and add SSL settings

edmocosta:
### What this PR does?
Added the following SSL settings: 
`ssl_truststore_type`: The format of the truststore file
`ssl_keystore_type`: The format of the keystore file
`ssl_certificate`: OpenSSL-style X.509 certificate file to authenticate the client
`ssl_key`: OpenSSL-style RSA private key that corresponds to the `ssl_certificate`
`ssl_cipher_suites`: The list of cipher suites
   
Reviewed and deprecated the following SSL settings to comply with Logstash's naming convention:
`ssl` in favor of `ssl_enabled`
`cacert` in favor of `ssl_certificate_authorities`
`keystore` in favor of `ssl_keystore_path`
`keystore_password` in favor of `ssl_keystore_password`
`truststore` in favor of `ssl_truststore_path`
`truststore_password` in favor of `ssl_truststore_password`
`ssl_certificate_verification` in favor of `ssl_verification_mode`

The behavior standardization across plugins, such as the accepted certificate formats, default values, etc will be tackled in future PRs.

---
Closes https://github.com/elastic/logstash/issues/14924
Closes https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/404
Closes https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/672
Closes https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/676

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,13 @@
+## 11.14.0
+ - Reviewed and deprecated SSL settings to comply with Logstash's naming convention [#1115](https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/1115)
+   - Deprecated `ssl` in favor of `ssl_enabled`
+   - Deprecated `cacert` in favor of `ssl_certificate_authorities`
+   - Deprecated `keystore` in favor of `ssl_keystore_path`
+   - Deprecated `keystore_password` in favor of `ssl_keystore_password`
+   - Deprecated `truststore` in favor of `ssl_truststore_path`
+   - Deprecated `truststore_password` in favor of `ssl_truststore_password`
+   - Deprecated `ssl_certificate_verification` in favor of `ssl_verification_mode`
+
 ## 11.13.1
  - Avoid crash by ensuring ILM settings are injected in the correct location depending on the default (or custom) template format, template_api setting and ES version [#1102](https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/1102)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 ## 11.14.0
- - Reviewed and deprecated SSL settings to comply with Logstash's naming convention [#1115](https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/1115)
+ - Added SSL settings for: [#1115](https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/1115)
+   - `ssl_truststore_type`: The format of the truststore file
+   - `ssl_keystore_type`: The format of the keystore file
+   - `ssl_certificate`: OpenSSL-style X.509 certificate file to authenticate the client
+   - `ssl_key`: OpenSSL-style RSA private key that corresponds to the `ssl_certificate`
+   - `ssl_cipher_suites`: The list of cipher suites
+ - Reviewed and deprecated SSL settings to comply with Logstash's naming convention
    - Deprecated `ssl` in favor of `ssl_enabled`
    - Deprecated `cacert` in favor of `ssl_certificate_authorities`
    - Deprecated `keystore` in favor of `ssl_keystore_path`
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -299,7 +299,7 @@ checks.
 ==== Elasticsearch Output Configuration Options
 
 This plugin supports the following configuration options plus the
-<<plugins-{type}s-{plugin}-common-options>> described later.
+<<plugins-{type}s-{plugin}-common-options>> and the <<plugins-{type}s-{plugin}-deprecated-options>> described later.
 
 [cols="<,<,<",options="header",]
 |=======================================================================
@@ -307,7 +307,6 @@ This plugin supports the following configuration options plus the
 | <<plugins-{type}s-{plugin}-action>> |<<string,string>>|No
 | <<plugins-{type}s-{plugin}-api_key>> |<<password,password>>|No
 | <<plugins-{type}s-{plugin}-bulk_path>> |<<string,string>>|No
-| <<plugins-{type}s-{plugin}-cacert>> |a valid filesystem path|__Deprecated__
 | <<plugins-{type}s-{plugin}-ca_trusted_fingerprint>> |<<string,string>>|No
 | <<plugins-{type}s-{plugin}-cloud_auth>> |<<password,password>>|No
 | <<plugins-{type}s-{plugin}-cloud_id>> |<<string,string>>|No
@@ -333,8 +332,6 @@ This plugin supports the following configuration options plus the
 | <<plugins-{type}s-{plugin}-ilm_policy>> |<<string,string>>|No
 | <<plugins-{type}s-{plugin}-ilm_rollover_alias>> |<<string,string>>|No
 | <<plugins-{type}s-{plugin}-index>> |<<string,string>>|No
-| <<plugins-{type}s-{plugin}-keystore>> |a valid filesystem path|__Deprecated__
-| <<plugins-{type}s-{plugin}-keystore_password>> |<<password,password>>|__Deprecated__
 | <<plugins-{type}s-{plugin}-silence_errors_in_log>> |<<array,array>>|No
 | <<plugins-{type}s-{plugin}-manage_template>> |<<boolean,boolean>>|No
 | <<plugins-{type}s-{plugin}-parameters>> |<<hash,hash>>|No
@@ -358,10 +355,8 @@ This plugin supports the following configuration options plus the
 | <<plugins-{type}s-{plugin}-sniffing>> |<<boolean,boolean>>|No
 | <<plugins-{type}s-{plugin}-sniffing_delay>> |<<number,number>>|No
 | <<plugins-{type}s-{plugin}-sniffing_path>> |<<string,string>>|No
-| <<plugins-{type}s-{plugin}-ssl>> |<<boolean,boolean>>|__Deprecated__
 | <<plugins-{type}s-{plugin}-ssl_certificate>> |<<path,path>>|No
 | <<plugins-{type}s-{plugin}-ssl_certificate_authorities>> |list of <<path,path>>|No
-| <<plugins-{type}s-{plugin}-ssl_certificate_verification>> |<<boolean,boolean>>|__Deprecated__
 | <<plugins-{type}s-{plugin}-ssl_cipher_suites>> |list of <<string,string>>|No
 | <<plugins-{type}s-{plugin}-ssl_enabled>> |<<boolean,boolean>>|No
 | <<plugins-{type}s-{plugin}-ssl_key>> |<<path,path>>|No
@@ -378,8 +373,6 @@ This plugin supports the following configuration options plus the
 | <<plugins-{type}s-{plugin}-template_name>> |<<string,string>>|No
 | <<plugins-{type}s-{plugin}-template_overwrite>> |<<boolean,boolean>>|No
 | <<plugins-{type}s-{plugin}-timeout>> |<<number,number>>|No
-| <<plugins-{type}s-{plugin}-truststore>> |a valid filesystem path|__Deprecated__
-| <<plugins-{type}s-{plugin}-truststore_password>> |<<password,password>>|__Deprecated__
 | <<plugins-{type}s-{plugin}-upsert>> |<<string,string>>|No
 | <<plugins-{type}s-{plugin}-user>> |<<string,string>>|No
 | <<plugins-{type}s-{plugin}-validate_after_inactivity>> |<<number,number>>|No
@@ -433,15 +426,6 @@ Elasticsearch {ref}/security-api-create-api-key.html[Create API key API].
 
 HTTP Path to perform the _bulk requests to
 this defaults to a concatenation of the path parameter and "_bulk"
-
-[id="plugins-{type}s-{plugin}-cacert"]
-===== `cacert`
-deprecated[11.14.0, Replaced by <<plugins-{type}s-{plugin}-ssl_certificate_authorities>>]
-
-  * Value type is a list of <<path,path>>
-  * There is no default value for this setting.
-
-The .cer or .pem file to validate the server's certificate.
 
 [id="plugins-{type}s-{plugin}-ca_trusted_fingerprint"]
 ===== `ca_trusted_fingerprint`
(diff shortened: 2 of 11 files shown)


docs/index.asciidoc
```diff
@@ -771,15 +784,19 @@ formats] and the `@timestamp` field of each event is being used as source for th
...
 ===== `keystore`
+deprecated[8.8.0, Replaced by <<plugins-{type}s-{plugin}-ssl_keystore_path>>]
 
   * Value type is <<path,path>>
   * There is no default value for this setting.
 
 The keystore used to present a certificate to the server.
 It can be either .jks or .p12
 
+NOTE: You cannot use this setting and `ssl_certificate` at the same time.
```
andsel, line 795:
```suggestion
NOTE: You cannot use this setting and <<plugins-{type}s-{plugin}-ssl_certificate>> at the same time.
```

docs/index.asciidoc
```diff
@@ -1044,8 +1062,28 @@ Enable SSL/TLS secured communication to Elasticsearch cluster.
...
 If no explicit protocol is specified plain HTTP will be used.
 
+[id="plugins-{type}s-{plugin}-ssl_certificate"]
+===== `ssl_certificate`
+  * Value type is <<path,path>>
+  * There is no default value for this setting.
+
+SSL certificate to use to authenticate the client. This certificate should be an OpenSSL-style X.509 certificate file.
+
+NOTE: This setting can be used only if `ssl_key` is set.
```
andsel, line 1072:
```suggestion
NOTE: This setting can be used only if <<plugins-{type}s-{plugin}-ssl_key>> is set.
```

docs/index.asciidoc
```diff
@@ -1044,8 +1062,28 @@ Enable SSL/TLS secured communication to Elasticsearch cluster.
...
+
+[id="plugins-{type}s-{plugin}-ssl_certificate_authorities"]
+===== `ssl_certificate_authorities`
+
+  * Value type is a list of <<path,path>>
+  * There is no default value for this setting
+
+The .cer or .pem files to validate the server's certificate.
```
andsel, line 1080:
```suggestion
The `.cer` or `.pem` files to validate the server's certificate.
```

docs/index.asciidoc
```diff
@@ -1044,8 +1062,28 @@ Enable SSL/TLS secured communication to Elasticsearch cluster.
...
+
+NOTE: You cannot use this setting and `ssl_truststore_path` at the same time.
```
andsel, line 1082:
```suggestion
NOTE: You cannot use this setting and <<plugins-{type}s-{plugin}-ssl_truststore_path>> at the same time.
```

docs/index.asciidoc
```diff
@@ -1054,6 +1092,60 @@ Option to validate the server's certificate. Disabling this severely compromises
...
+Enable SSL/TLS secured communication to Elasticsearch cluster.
+Leaving this unspecified will use whatever scheme is specified in the URLs listed in <<plugins-{type}s-{plugin}-hosts>> or extracted from the <<plugins-{type}s-{plugin}-cloud_id>>.
+If no explicit protocol is specified plain HTTP will be used.
+
+[id="plugins-{type}s-{plugin}-ssl_key"]
+===== `ssl_key`
+  * Value type is <<path,path>>
+  * There is no default value for this setting.
+
+OpenSSL-style RSA private key that corresponds to the `ssl_certificate`.
```
andsel, line 1118:
```suggestion
OpenSSL-style RSA private key that corresponds to the <<plugins-{type}s-{plugin}-ssl_certificate>>.
```

docs/index.asciidoc
```diff
@@ -1054,6 +1092,60 @@ Option to validate the server's certificate. Disabling this severely compromises
...
+
+NOTE: This setting can be used only if `ssl_certificate` is set.
```
andsel, line 1120:
```suggestion
NOTE: This setting can be used only if <<plugins-{type}s-{plugin}-ssl_certificate>> is set.
```

docs/index.asciidoc
```diff
@@ -1054,6 +1092,60 @@ Option to validate the server's certificate. Disabling this severely compromises
...
+[id="plugins-{type}s-{plugin}-ssl_keystore_path"]
+===== `ssl_keystore_path`
+
+  * Value type is <<path,path>>
+  * There is no default value for this setting.
+
+The keystore used to present a certificate to the server.
+It can be either .jks or .p12
+
+NOTE: You cannot use this setting and `ssl_certificate` at the same time.
```
andsel, line 1139:
```suggestion
NOTE: You cannot use this setting and <<plugins-{type}s-{plugin}-ssl_certificate>> at the same time.
```

docs/index.asciidoc
```diff
@@ -1071,6 +1163,49 @@ NOTE: If you configure the plugin to use `'TLSv1.1'` on any recent JVM, such as
...
+[id="plugins-{type}s-{plugin}-ssl_truststore_path"]
+===== `ssl_truststore_path`
+
+  * Value type is <<path,path>>
+  * There is no default value for this setting.
+
+The truststore to validate the server's certificate.
+It can be either .jks or .p12.
+
+NOTE: You cannot use this setting and `ssl_certificate_authorities` at the same time.
```
andsel, line 1183:
```suggestion
NOTE: You cannot use this setting and <<plugins-{type}s-{plugin}-ssl_certificate_authorities>> at the same time.
```

lib/logstash/outputs/elasticsearch/http_client_builder.rb
```diff
@@ -107,38 +107,61 @@ def self.create_http_client(options)
...
+      if ssl_certificate_authorities&.any?
+        raise(LogStash::ConfigurationError, 'Multiple values on "ssl_certificate_authorities" are not supported by this plugin') if ssl_certificate_authorities.size > 1
+        ssl_options[:ca_file] = ssl_certificate_authorities.first
+      end
+
+      if ssl_truststore_path
+        ssl_options[:truststore] = ssl_truststore_path
+        ssl_options[:truststore_type] = params["ssl_truststore_type"] if params.include?("ssl_truststore_type")
+        ssl_options[:truststore_password] = params["ssl_truststore_password"].value if params.include?("ssl_truststore_password")
+      end
```
andsel, line 136: Maybe an overkill, but these two keystore and trustore setup could extracted like:

```ruby
# kind is a string [trustore|keystore]
def setup_store(ssl_options, kind, store_path)
    if store_path
        ssl_options[kind.to_sym] = store_path
        ssl_options["#{kind}_type".to_sym] = params["ssl_#{kind}_type"] if params.include?("ssl_#{kind}_type")
        ssl_options["#{kind}_password".to_sym] = params["ssl_#{kind}_password"].value if params.include?("ssl_#{kind}_password")
   end
end


# usage is 
setup_store(ssl_options, 'trustore', ssl_truststore_path)
setup_store(ssl_options, 'keystore', ssl_keystore_path)
```

lib/logstash/outputs/elasticsearch/http_client_builder.rb
```diff
@@ -107,38 +107,61 @@ def self.create_http_client(options)
...
 
-      ssl_options[:truststore] = truststore if truststore
-      if keystore
-        ssl_options[:keystore] = keystore
-        ssl_options[:keystore_password] = keystore_password.value if keystore_password
+      ssl_key = params["ssl_key"]
+      if ssl_certificate && ssl_key
+        ssl_options[:client_cert] = ssl_certificate
+        ssl_options[:client_key] = ssl_key
+      elsif !!ssl_certificate ^ !!ssl_key
```
andsel, line 148: Given that the other branch of the if cover the `true` && `true` case we are sure that here is never happens so we could use the widely used logic `||` operator. The semantic remains the same.
I'm saying just because the xor `^` bitwise operator it's little bit unusual for control logic.

spec/unit/outputs/elasticsearch_spec.rb
```diff
@@ -712,6 +711,81 @@
...
 
+  describe "SSL deprecated settings" do
+    context "with client certificate" do
+      let(:do_register) { true }
+      let(:cacert) { Stud::Temporary.file.path }
+      let(:options) { {
+        "ssl" => "true",
+        "cacert" => cacert,
+        "ssl_certificate_verification" => false
+      } }
```
andsel, line 722: Given that this fixture part is almost equally value in both contexts then it could be lifted in the outer `describe` for example:
```ruby
context "SSL deprecated settings" do
  let(:do_register) { true }
  let(:cacert) { Stud::Temporary.file.path }
  let(:base_options) do  # in case of multiline blocks the do..end should be used instead of {...}
   {
        "ssl" => true,
      }
  end
  
  context "with client certificate" do
    let(:options) do 
      base_options.merge({ 
        "ssl_certificate_verification" => false, 
        "cacert" => cacert,
      }) 
    end   
     # test code here
  end
  
  context "with java stores" do
    let(:options) do base_options.merge({
        "keystore" => keystore,
        "keystore_password" => "keystore",
        "truststore" => truststore,
        "truststore_password" => "truststore",
        "ssl_certificate_verification" => true
      }
    end
  end
end
```

In this case it makes evident what changes in the fixture setup between the two cases.

andsel (changes requested):
First review run.
Left some comments to evaluate.

docs/index.asciidoc
```diff
@@ -771,15 +784,19 @@ formats] and the `@timestamp` field of each event is being used as source for th
 
 [id="plugins-{type}s-{plugin}-keystore"]
 ===== `keystore`
+deprecated[8.8.0, Replaced by <<plugins-{type}s-{plugin}-ssl_keystore_path>>]
```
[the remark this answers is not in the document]
yaauie: Backwards compatible changes do _no_t require a major bump. This is additive, and all configurations that work under 11.13.x will continue to work without modification in 11.14.0.

lib/logstash/outputs/elasticsearch/http_client_builder.rb
```diff
@@ -107,38 +107,52 @@ def self.create_http_client(options)
...
+      setup_ssl_store(ssl_options, 'truststore', params)
+      setup_ssl_store(ssl_options, 'keystore', params)
+
+      ssl_key = params["ssl_key"]
+      if ssl_certificate && ssl_key
+        ssl_options[:client_cert] = ssl_certificate
+        ssl_options[:client_key] = ssl_key
+      elsif !!ssl_certificate || !!ssl_key
+        raise(LogStash::ConfigurationError, 'You must set both "ssl_certificate" and "ssl_key" for client authentication')
       end
```
yaauie, line 141: Message is a bit unclear (you don't actually have to set either, because you can use a keystore), and catches two distinct cases:
 - using `ssl_certificate` requires `ssl_key` AND
 - using `ssl_key` requires `ssl_certificate`

I prefer declaring it in a way that centers the single config that "activates" the feature, in this case `ssl_certificate`:
 - using `ssl_certificate` requires an `ssl_key`
- an `ssl_certificate` is required when using an `ssl_key`

docs/index.asciidoc
```diff
@@ -358,16 +358,28 @@ This plugin supports the following configuration options plus the
 | <<plugins-{type}s-{plugin}-sniffing>> |<<boolean,boolean>>|No
 | <<plugins-{type}s-{plugin}-sniffing_delay>> |<<number,number>>|No
 | <<plugins-{type}s-{plugin}-sniffing_path>> |<<string,string>>|No
-| <<plugins-{type}s-{plugin}-ssl>> |<<boolean,boolean>>|No
-| <<plugins-{type}s-{plugin}-ssl_certificate_verification>> |<<boolean,boolean>>|No
+| <<plugins-{type}s-{plugin}-ssl>> |<<boolean,boolean>>|__Deprecated__
```
yaauie, line 361: I would prefer not to leave the deprecated settings in this table, and perhaps to have a separate section with them and their detailed blocks, so that we can reduce the noise new users of the plugin encounter while still ensuring that the deprecated options are still discoverable to those who are looking for documentation for options currently in their pipeline configs.

andsel (approved):
LGTM

CHANGELOG.md
```diff
@@ -1,3 +1,19 @@
+## 11.14.0
+ - Added SSL settings for: [#1115](https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/1115)
+   - `ssl_truststore_type`: The format of the truststore file
```
karenzone, line 3: Is this PR in the changelog correct?  Shouldn't it be a link to this PR (#1118)?

docs/index.asciidoc
```diff
@@ -1213,6 +1271,97 @@ https://www.elastic.co/blog/elasticsearch-versioning-support[versioning support
 blog] and {ref}/docs-index_.html#_version_types[Version types] in the
 Elasticsearch documentation.
 
+[id="plugins-{type}s-{plugin}-deprecated-options"]
+==== Elasticsearch Output Deprecated Configuration Options
```
karenzone, line 1275: Deviation from the way we typically handle deprecations in plugins.  (See https://www.elastic.co/guide/en/logstash/current/plugins-inputs-beats.html#plugins-inputs-beats-options.) 

Standards can change over time, but the change needs to be deliberate and communicated clearly.

karenzone: Nevermind. I see where this came from: https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/1118#discussion_r1131892391. 

This is definitely better for new users, but isn't very nice to our existing users.

CHANGELOG.md
edmocosta: Yes, you're right, it should be #1118. I'll submit a fix! Thanks for the heads up!

karenzone: Thanks, @edmo.  Remember that the link will be wrong for any release notes that include 11.14.0, and will require a manual fix.   :-)

whyyouwannaknow:
Hello,

I might not post in the correct issue but we saw a deprecation on the setting `ssl_certificate_verification ` which we replaced by `ssl_verification_mode`.
The thing is that this new setting `ssl_verification_mode` only supports the value `full` or `none` but not the value `certificate`.

This value `certificate` should be implemented in the output ElasticSearch of a Logstash pipeline as it is more secured then putting `none` instead.

Could this be implemented in a future update? Shoud we open a separated issue for this?

Thank you.

Regards,

edmocosta:
Hi @whyyouwannaknow!

Thank you for reaching out! Unfortunately, the underline HTTP library this plugin uses does not fully support the `certificate` mode, that's why it only accepts the `full` & `none` options. Please note that the previous configuration also didn't support it, being `true` => `full` and `false` => `none`.

We're working on standardising the SSL/TLS configs and features across plugins (https://github.com/elastic/logstash/issues/14905), and that project's second phase also includes adding support to the `certificate` mode. 

Thank you!

PR #1119, opened 2023-03-08
Check pipeline shutdown status while waiting for valid connection or while issuing a bulk to ES

andsel:
## Release notes

Introduce the ability to negatively acknowledge the batch under processing if the plugin is blocked in a retry-error-loop and a shutdown is requested.

## What does this PR do?
Updates the wait_for_successful_connection method and safe_bulk to react to shutdown requests.
When the plugin is in a retry loop and Logstash it's running into provide the ability to negatively ACK the batch under processing and a pipeline shutdown was requested then the plugin terminate signalling that the batch under processing shouldn't be acknowledged, raising an AbortedBatchException.

Let the user to update the configuration of the plugin, when managed by CPM, resuming from a retry-loop due to bad configuration values (or expired credentials);

## Checklist

## Author's Checklist

- [ ] Test with cloud instance

This PR has to be tested in lockstep with the changes on Logstash core applied with https://github.com/elastic/logstash/pull/14940
Checkout a branch which contains that change or a release if already published.
The test plan needs to create an Elasticsearch instance in Elastic cloud, recording the following data:
- `cloud_id`
- `cloud_auth`
- `api_key`

### Test plan
After created the Elastic deployment execute the following steps:
-  checkout this  branch
- install this plugin in your Logstash instance
```sh
gem "logstash-output-elasticsearch", :path => "/path/to/logstash-output-elasticsearch"
bin/logstash-plugin install --no-verify
```
- configure your Logstash's `config/logstash.yml` file to connect to central pipeline management (CPM) in Elastic cloud
```yaml
xpack.management.enabled: true
xpack.management.pipeline.id: ["test*"]
xpack.management.elasticsearch.cloud_id: "<your cloud_id>"
xpack.management.elasticsearch.cloud_auth: "<your cloud_auth>"
```
- in CPM create a new pipeline like the following:
```
input {
  tcp {
    port => 1234
  }
}

output {
  elasticsearch {
    cloud_id => "<your cloud_id>"
    api_key => "<your api_key>"
    index => "andsel_test"
    ssl => true
  }
  stdout{}
}
```
- run Logstash (`bin/logstash`)
- verify that it can process some input
```sh
echo "test input line" | netcat localhost 1234
```
- now update the the pipeline definition on CPM, corrupting the `api_key` such as removing last char of the key (remember somewhere the correct key)
- the output should start blaming a connectivity problem with log lines such as:
```
[2023-03-22T14:38:23,927][WARN ][logstash.outputs.elasticsearch][test_remote_1] Attempted to resurrect connection to dead ES instance, but got an error {:url=>"https://blabla.gcp.cloud.es.io:443/", :exception=>LogStash::Outputs::ElasticSearch::HttpClient::Pool::BadResponseCodeError, :message=>"Got response code '401' contacting Elasticsearch at URL 'https://blabla.gcp.cloud.es.io:443/'"}
```
- in CPM restore the correct `api_key`. Without this fix the previous lines continues and no shutdown and restart of the pipeline happens, With this change you should fine a line like
```
[INFO ][org.logstash.execution.WorkerLoop][test_remote_1] Worker loop notified of aborting a batch
```
and the pipeline effectively reloads and is back to normal functioning.

## Related issues

- Superseed #1107
- Superseed #1117
- Relates https://github.com/elastic/logstash/pull/14940
- Relates https://github.com/elastic/logstash/pull/13811
- Fix https://github.com/elastic/logstash/issues/14739

## Use cases
As a user using the central pipeline management I want that once a credential is updated the pipeline running the plugin is effectively restarted with the updated values without manual intervention on the Logstash instance.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 11.15.0
+ - Added the ability to negatively acknowledge the batch under processing if the plugin is blocked in a retry-error-loop and a shutdown is requested. [#1119](https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/1119)
+
 ## 11.14.1
  - [DOC] Fixed incorrect pull request link on the CHANGELOG `11.14.0` entry [#1122](https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/1122)
 
--- a/lib/logstash/outputs/elasticsearch.rb
+++ b/lib/logstash/outputs/elasticsearch.rb
@@ -431,7 +431,12 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
   def wait_for_successful_connection
     after_successful_connection_done = @after_successful_connection_done
     return unless after_successful_connection_done
-    stoppable_sleep 1 until after_successful_connection_done.true?
+    stoppable_sleep 1 until (after_successful_connection_done.true? || execution_context&.pipeline&.shutdown_requested?)
+
+    if execution_context&.pipeline&.shutdown_requested?
+      @logger.info "Aborting the batch due to shutdown request while waiting for connections to become live"
+      raise org.logstash.execution.AbortedBatchException.new
+    end
 
     status = @after_successful_connection_thread && @after_successful_connection_thread.value
     if status.is_a?(Exception) # check if thread 'halted' with an error
--- a/lib/logstash/outputs/elasticsearch.rb
+++ b/lib/logstash/outputs/elasticsearch.rb
@@ -435,7 +435,7 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
 
     if execution_context&.pipeline&.shutdown_requested?
       @logger.info "Aborting the batch due to shutdown request while waiting for connections to become live"
-      raise org.logstash.execution.AbortedBatchException.new
+      abort_batch_if_available!
     end
 
     status = @after_successful_connection_thread && @after_successful_connection_thread.value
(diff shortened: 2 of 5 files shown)


andsel:
The tests for version `8.x` are green because the selected version results to be `8.7.0-SNAPSHOT`.
The testing against `8.8.0-SNAPSHOT` will happen after the `8.7.0` release, when selected [snapshot version change](https://github.com/elastic/logstash/blob/56e626afc55cc9360c0d8c8e1c63e743900ea0a0/ci/logstash_releases.json#L10).

jsvd (changes requested):
I tested a few scenarios:
A) Writing to ES when the API key is revoked and the Pipeline is updated with new API Key: pipeline restarts correctly ✅ 
B) Writing to ES when ES goes down, binds to a different port, Pipeline is updated with new port: the pipeline never terminates 🛑 

B) seems to fail as the ressureccionist thread never terminates, likely due to https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/main/lib/logstash/outputs/elasticsearch/http_client/pool.rb#L129C11-L146

andsel:
The `stop` method is invoked as a notification after all the pipeline workers has terminated. The  `@stopping` at 
https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/a0b4832bd4d4e045cdc2c485e79c5949379c96b3/lib/logstash/outputs/elasticsearch/http_client/pool.rb#L91
is invoked _after_ the plugin has terminated his `#multi_process` from the LS's pipeline during [shutdown_workers](url) execution.

A solution to exit the resurrectionist loop would be to propagate the `shutdown_requested ` down to the ES pool so that the `@stopping` could be flipped both during `#close` and also when shutdown is requested.

andsel:
With commit 5e57c3ab7854fbbd7a8f2d708b1b09e9e343bb17 was introduced the abort of a batch in case no connections are available in the pool and the pipeline is shutting down.

jsvd (approved):
LGTM, tested manually a few scenarios, ran the unit tests without the necessary patches, everything is working as expected.  :shipit:

PR #1121, opened 2023-03-21
Fix #1120: Get rid of warning message about unreleased ECS schema v8

### LINKED ISSUE #1120: Get rid of startup warning message about UNRELEASED ECS Schema v8
Each time a Logstash instance in version 8 configured with the Elasticsearch Plugin starts, it logs this Warning message:

```
[2023-03-21T15:56:51,383][WARN ][logstash.outputs.elasticsearch][another_test] Elasticsearch Output configured with `ecs_compatibility => v8`, which resolved to an UNRELEASED preview of version 8.0.0 of the Elastic Common Schema. Once ECS v8 and an updated release of this plugin are publicly available, you will need to update this plugin to resolve this warning.
```

This is caused by this [if block](https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/cfe44f3655e7cafa967d3a57d73e2287354f057d/lib/logstash/outputs/elasticsearch.rb#L353) (reported here for convenience):
```ruby
if ecs_compatibility == :v8
      @logger.warn("Elasticsearch Output configured with `ecs_compatibility => v8`, which resolved to an UNRELEASED preview of version 8.0.0 of the Elastic Common Schema. " +
                   "Once ECS v8 and an updated release of this plugin are publicly available, you will need to update this plugin to resolve this warning.")
end
````

Given that ECS 8 is a long standing reality, we can (IMO) safely deactivate this warning message (i.e., remove the if-block) to reduce the noise in Logstash Logs.

herrBez:
Fix #1120 by deleting the if-blocks that cause the log lines to be logged.

### DIFF
--- a/lib/logstash/outputs/elasticsearch.rb
+++ b/lib/logstash/outputs/elasticsearch.rb
@@ -349,11 +349,6 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
 
     @bulk_request_metrics = metric.namespace(:bulk_requests)
     @document_level_metrics = metric.namespace(:documents)
-
-    if ecs_compatibility == :v8
-      @logger.warn("Elasticsearch Output configured with `ecs_compatibility => v8`, which resolved to an UNRELEASED preview of version 8.0.0 of the Elastic Common Schema. " +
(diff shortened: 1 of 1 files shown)


mashhurs:
This was addressed by this change: https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/1131, closing.

PR #1125, opened 2023-04-14
move async finish_register to bottom of register to avoid race condition

### LINKED ISSUE #1126: Race condition between register and finish_register lands data in ecs_logstash instead of data stream
The `register` method has a sequence of actions:

1. [setup_template_manager_defaults(data_stream_enabled)](https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/v11.15.0/lib/logstash/outputs/elasticsearch.rb#L315)
2. [after_successful_connection w/ finish_register](https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/v11.15.0/lib/logstash/outputs/elasticsearch.rb#L317-L327)
3. [sets up `@index` according to data stream](https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/v11.15.0/lib/logstash/outputs/elasticsearch.rb#L341-L348)

During 2. a new thread is launched to execute [`finish_register`](https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/v11.15.0/lib/logstash/outputs/elasticsearch.rb#L360-L366) once a connection is established.

A race condition can happen if the order is:

1. [thread is launched in register to run `finish_register`](https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/v11.15.0/lib/logstash/outputs/elasticsearch.rb#L317)
2. [the first set of instructions from `finish_register`](https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/v11.15.0/lib/logstash/outputs/elasticsearch.rb#L330-L356) execute
4. [the rest of `register` executes](https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/v11.15.0/lib/logstash/outputs/elasticsearch.rb#L330-L356)
5. [ilm_setup](https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/v11.15.0/lib/logstash/outputs/elasticsearch.rb#L364-L365) executes.

This sequence causes [`install_template`](https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/v11.15.0/lib/logstash/outputs/elasticsearch.rb#L363) to be called before `@index` is set to a data stream [here](https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/v11.15.0/lib/logstash/outputs/elasticsearch.rb#L344).
During `install_template`, [`ilm_in_use?` is called](https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/v11.15.0/lib/logstash/outputs/elasticsearch/template_manager.rb#L27), and [`ilm_actually_enabled`](https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/v11.15.0/lib/logstash/outputs/elasticsearch/ilm.rb#L14) becomes true.
Once `finish_register` gets to [`setup_ilm if ilm_in_use?`](https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/v11.15.0/lib/logstash/outputs/elasticsearch.rb#L364), `ilm_in_use?` is true and so `@index` is overwritten [here](https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/v11.15.0/lib/logstash/outputs/elasticsearch/ilm.rb#L7-L8).

jsvd:
This PR fixes a race condition between the thread running `register` and the one running `finish_register`.

The issue can be observed by having 20-30 simple pipelines in `pipelines.yml` like:

```
input {
 java_generator { eps => 1.1 }
}
output {
  elasticsearch {
    user => elastic
    password => ".."
    data_stream => true
  }
```

```yml
- pipeline.id: test_0
  pipeline.workers: 1
  pipeline.batch.size: 1
  path.config: "/tmp/logstash-8.7.0/cfg"
```

```
ruby -ryaml -e '30.times.each {|i| puts "- pipeline.id: test_#{i}\n  pipeline.workers: 1\n  pipeline.batch.size: 1\n  path.config: \"/tmp/logstash-8.7.0/cfg\""}' > config/pipelines.yml
```
Run Logstash w/ pipeline reloading:
```
bin/logstash -r
```

Changing the pipeline config a few times and letting Logstash reload will eventually show one or more pipelines overwriting the index with "ecs-logstash":

```
[2023-04-20T12:08:33,181][WARN ][logstash.outputs.elasticsearch][test_0] Overwriting supplied index logs-generic-default with rollover alias ecs-logstash
```

Putting `finish_register` at the end of register ensures there are no concurrent instructions manipulating `@index` and other internal state variables.

fixes #1126

### DIFF
--- a/spec/unit/outputs/elasticsearch_spec.rb
+++ b/spec/unit/outputs/elasticsearch_spec.rb
@@ -1459,7 +1459,7 @@ describe LogStash::Outputs::ElasticSearch do
 
     context 'during register/finish_register' do
 
-      let(:options) { { 'hosts' => '127.0.0.1:9999' } }
+      let(:options) { { 'hosts' => '127.0.0.1:9999', 'data_stream' => 'true' } }
       let(:es_version) { '8.7.0' } # DS default on LS 8.x
       let(:latch) { Concurrent::CountDownLatch.new }
 
@@ -1477,7 +1477,7 @@ describe LogStash::Outputs::ElasticSearch do
       # after `ilm_in_use?` is called but before `setup_ilm`
       it 'doesn\'t have a race condition leading to resetting back to ILM' do
         ilm_in_use = subject.method(:ilm_in_use?)
-        expect(subject).to receive(:ilm_in_use?).once do |params|
+        expect(subject).to receive(:ilm_in_use?) do |params|
           ret = ilm_in_use.call
           sleep 3
           ret
--- a/spec/unit/outputs/elasticsearch_spec.rb
+++ b/spec/unit/outputs/elasticsearch_spec.rb
@@ -1461,7 +1461,6 @@ describe LogStash::Outputs::ElasticSearch do
 
       let(:options) { { 'hosts' => '127.0.0.1:9999', 'data_stream' => 'true' } }
       let(:es_version) { '8.7.0' } # DS default on LS 8.x
-      let(:latch) { Concurrent::CountDownLatch.new }
 
       before do
         allow(subject).to receive(:discover_cluster_uuid)
--- a/spec/unit/outputs/elasticsearch_spec.rb
+++ b/spec/unit/outputs/elasticsearch_spec.rb
@@ -1456,6 +1456,43 @@ describe LogStash::Outputs::ElasticSearch do
       end
 
     end
+
+    context 'during register/finish_register' do
+
+      let(:options) { { 'hosts' => '127.0.0.1:9999' } }
+      let(:es_version) { '8.7.0' } # DS default on LS 8.x
+      let(:latch) { Concurrent::CountDownLatch.new }
+
+      before do
+        allow(subject).to receive(:discover_cluster_uuid)
+        allow(subject).to receive(:maybe_create_rollover_alias)
+        allow(subject).to receive(:maybe_create_ilm_policy)
+        allow(subject).to receive(:build_client)
+      end
+
+      # this test could not have been done using latches as the fix to the race
+      # condition alters the order of the instructions in elasticsearch.rb
+      #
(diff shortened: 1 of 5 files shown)


spec/unit/outputs/elasticsearch_spec.rb
```diff
@@ -1456,6 +1456,43 @@
       end
 
     end
+
+    context 'during register/finish_register' do
+
+      let(:options) { { 'hosts' => '127.0.0.1:9999', 'data_stream' => 'true' } }
+      let(:es_version) { '8.7.0' } # DS default on LS 8.x
+      let(:latch) { Concurrent::CountDownLatch.new }
```
kaisecheng, line 1464: latch is not in use and can be removed?

kaisecheng (approved):
LGTM

jsvd:
For reliable reproduceability:

```
/tmp/test/logstash-8.7.0
> bin/logstash-plugin update logstash-output-elasticsearch
> vim ./vendor/bundle/jruby/2.6.0/gems/logstash-output-elasticsearch-11.15.0-java/lib/logstash/outputs/elasticsearch.rb
```

1. Add "sleep 1" before the `if` on line https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/main/lib/logstash/outputs/elasticsearch.rb#L341
2. Change line https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/main/lib/logstash/outputs/elasticsearch.rb#L364 to `    (sleep 3; setup_ilm) if ilm_in_use?`

andsel:
@jsvd tested as you suggested and it was reproducible, confirming that the PR fixes it.

jsvd:
@logstashmachine bump patch

PR #1127, opened 2023-04-20
fix integration test stability

jsvd:
data may be in other indices, let's focus the tests on the index we write to


kaisecheng (approved):
LGTM

PR #1128, opened 2023-04-27
actual version bump :facepalm:

yaauie:
CHANGELOG on main already includes the entry for 11.15.2 from #1105 -> https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/main/CHANGELOG.md#11152


mashhurs (approved):
LGTM!

PR #1132, opened 2023-05-16
Fixed abort condition in case of shutdown is requested during connection wait

andsel:
## Release notes
Fix the abort of batch also when the ES connection is alive and the processing should continue.

## What does this PR do?

Fixes the exit condition while waiting for valid connection. It could happen that while waiting both conditions becomes true:
- a connection was established
- the shutdown was requested

If the shutdown is requested *and* a valid ES connection is created the processing must take place and shouldn't abort the batch.

#### Technical observation
While waiting for connection to be established, the busy wait loop that also check for shutdown request. The abort of the batch should happen if no connection was successful established and there is a shutdown request; else it could be that the loop exit condition is trigger with both parts satisfied and it's not correct to abort the batch if a connection was created.

In this condition is important that at least one time Elasticsearch is reached before the flag `after_successful_connection_done`  reaches a final status, so it means that at least one `health_check!` is executed before the `#multi_receive` enter the wait loop.
This is granted by the fact that `health_check!` is executed during the `#register`.

It exposed in https://github.com/elastic/logstash/issues/15040 essentially running the command:
```sh
echo "foo" | bin/logstash -e 'input { stdin { } } output { elasticsearch {} }'
```

The `foo` message is not processed also if the ES is up and running.
Due to scheduling could be that `stdin` reached an EOF and request the shutdown while the ES output is waiting for connection. In this case, given that ES is up and running it should process the batch till the memory queue is completely consumed.

## Checklist

## Author's Checklist

- [x] test that Logstash's 37K integration test is green with this plugin's fix.

- in a local clone of Logstash set the elasticsearch output plugin to this PR checkout:
```
gem "logstash-output-elasticsearch", :path => "/path/to/logstash-output-elasticsearch"
bin/logstash-plugin install --no-verify
```
- run the integration test with
```sh
./gradlew runIntegrationTests -PrubyIntegrationSpecs="specs/es_output_how_spec.rb"
```

## Related issues

- Fixes https://github.com/elastic/logstash/issues/15040
- Closes https://github.com/elastic/logstash/issues/15009

### DIFF
--- a/spec/unit/outputs/elasticsearch_spec.rb
+++ b/spec/unit/outputs/elasticsearch_spec.rb
@@ -96,16 +96,13 @@ describe LogStash::Outputs::ElasticSearch do
     context "on a reachable ES instance" do
       let(:events) { [ ::LogStash::Event.new("foo" => "bar1"), ::LogStash::Event.new("foo" => "bar2") ] }
 
-      let(:shutdown_value) { true }
-
       let(:logger) { double("logger") }
+
       before(:each) do
         allow(subject).to receive(:logger).and_return(logger)
         allow(logger).to receive(:info)
 
-        allow(subject).to receive(:pipeline_shutdown_requested?) do
-          shutdown_value
-        end
+        allow(subject).to receive(:pipeline_shutdown_requested?).and_return(true)
         allow(subject).to receive(:retrying_submit)
       end
 
--- a/spec/unit/outputs/elasticsearch_spec.rb
+++ b/spec/unit/outputs/elasticsearch_spec.rb
@@ -107,6 +107,7 @@ describe LogStash::Outputs::ElasticSearch do
       end
 
       it "the #multi_receive doesn't abort when waiting for a connection on alive ES and a shutdown is requested" do
+        subject.multi_receive(events)
         expect(logger).to_not have_received(:info).with(/Aborting the batch due to shutdown request while waiting for connections to become live/i)
       end
     end
--- a/spec/unit/outputs/elasticsearch_spec.rb
+++ b/spec/unit/outputs/elasticsearch_spec.rb
@@ -155,6 +155,9 @@ describe LogStash::Outputs::ElasticSearch do
         end
 
         it "should exit the retry with an abort exception if shutdown is requested" do
+          # trigger the shutdown signal
+          allow(subject).to receive(:pipeline_shutdown_requested?) { true }
+
           # execute in another thread because it blocks in a retry loop until the shutdown is triggered
           th = Thread.new do
             subject.multi_receive([event])
@@ -162,9 +165,6 @@ describe LogStash::Outputs::ElasticSearch do
             # return exception's class so that it can be verified when retrieving the thread's value
             e.class
           end
-
-          # trigger the shutdown signal
-          allow(subject).to receive(:pipeline_shutdown_requested?) { true }
 
           expect(th.value).to eq(org.logstash.execution.AbortedBatchException)
         end
--- a/spec/unit/outputs/elasticsearch_spec.rb
+++ b/spec/unit/outputs/elasticsearch_spec.rb
@@ -58,7 +58,27 @@ describe LogStash::Outputs::ElasticSearch do
 
       let(:logger) { double("logger") }
 
+      let(:never_ending) { Thread.new { sleep 1 while true } }
+
+      let(:do_register) { false }
+
       before(:each) do
+        spy_http_client_builder!
+        stub_http_client_pool!
+
+        allow(subject).to receive(:finish_register) # stub-out thread completion (to avoid error log entries)
+
+        # emulate 'failed' ES connection, which sleeps forever
+        allow(subject).to receive(:after_successful_connection) { |&block| never_ending }
+        allow(subject).to receive(:stop_after_successful_connection_thread)
+
+        subject.register
+
+        allow(subject.client).to receive(:maximum_seen_major_version).at_least(:once).and_return(maximum_seen_major_version)
+        allow(subject.client).to receive(:get_xpack_info)
+
+        subject.client.pool.adapter.manticore.respond_with(:body => "{}")
+
         allow(subject).to receive(:logger).and_return(logger)
         allow(logger).to receive(:info)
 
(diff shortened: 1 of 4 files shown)


spec/unit/outputs/elasticsearch_spec.rb
```diff
@@ -73,6 +93,27 @@
...
+      let(:shutdown_value) { true }
+
+      let(:logger) { double("logger") }
+      before(:each) do
+        allow(subject).to receive(:logger).and_return(logger)
+        allow(logger).to receive(:info)
+
+        allow(subject).to receive(:pipeline_shutdown_requested?) do
+          shutdown_value
+        end
```
jsvd, line 108: Until it's necessary I prefer to be explicit to facilitate reading the code:

```suggestion
      let(:logger) { double("logger") }

      before(:each) do
        allow(subject).to receive(:logger).and_return(logger)
        allow(logger).to receive(:info)

        allow(subject).to receive(:pipeline_shutdown_requested?).and_return(true)
```

spec/unit/outputs/elasticsearch_spec.rb
```diff
@@ -73,6 +93,27 @@
...
+        allow(subject).to receive(:retrying_submit)
+      end
+
+      it "the #multi_receive doesn't abort when waiting for a connection on alive ES and a shutdown is requested" do
+        expect(logger).to_not have_received(:info).with(/Aborting the batch due to shutdown request while waiting for connections to become live/i)
+      end
+    end
```
jsvd, line 112:
```suggestion
      context "when waiting for a connection on alive ES and a shutdown is requested" do
        describe "#multi_receive" do
          it "doesn't abort" do
            expect(logger).to_not have_received(:info).with(/Aborting the batch due to shutdown request while waiting for connections to become live/i)
          end 
        end
      end
    end
```

jsvd (changes requested):
Unfortunately the test doesn't confirm that the changes fix an actual issue:

```
~/elastic/logstash-plugins/logstash-output-elasticsearch pr/1132*
❯ bundle exec rspec spec/unit/outputs/elasticsearch_spec.rb -e "ES and a shutdown is requested"
Sending Logstash logs to null which is now configured via log4j2.properties
Run options:
  include {:full_description=>/ES\ and\ a\ shutdown\ is\ requested/}
  exclude {:integration=>true, :redis=>true, :socket=>true, :performance=>true, :couchdb=>true, :elasticsearch=>true, :elasticsearch_secure=>true, :export_cypher=>true, :windows=>true}

Randomized with seed 44819
.

Finished in 0.29414 seconds (files took 1.82 seconds to load)
1 example, 0 failures

Randomized with seed 44819
```
Currently the test passes.
Looking at the changes in `lib/` compared to main:

```diff
~/elastic/logstash-plugins/logstash-output-elasticsearch pr/1132* 6s
❯ git diff -w main lib                                                                         
diff --git a/lib/logstash/outputs/elasticsearch.rb b/lib/logstash/outputs/elasticsearch.rb
index f384926..992db40 100644
--- a/lib/logstash/outputs/elasticsearch.rb
+++ b/lib/logstash/outputs/elasticsearch.rb
@@ -432,7 +432,7 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
     return unless after_successful_connection_done
     stoppable_sleep 1 until (after_successful_connection_done.true? || pipeline_shutdown_requested?)
 
-    if pipeline_shutdown_requested?
+    if pipeline_shutdown_requested? && !after_successful_connection_done.true?
       logger.info "Aborting the batch due to shutdown request while waiting for connections to become live"
       abort_batch_if_available!
     end
diff --git a/lib/logstash/plugin_mixins/elasticsearch/common.rb b/lib/logstash/plugin_mixins/elasticsearch/common.rb
index e5dd17e..f088c87 100644
--- a/lib/logstash/plugin_mixins/elasticsearch/common.rb
+++ b/lib/logstash/plugin_mixins/elasticsearch/common.rb
@@ -377,7 +377,7 @@ module LogStash; module PluginMixins; module ElasticSearch
 
     def pipeline_shutdown_requested?
         return super if defined?(super) # since LS 8.1.0
-      execution_context&.pipeline&.shutdown_requested?
+        execution_context&.pipeline&.shutdown_requested
       end
 
     def abort_batch_if_available!
```

Let's pick them up from main:

```
~/elastic/logstash-plugins/logstash-output-elasticsearch pr/1132*
❯ git co main lib     
Updated 2 paths from 7f18b32
```
And run the test again:

```~/elastic/logstash-plugins/logstash-output-elasticsearch pr/1132*
❯ bundle exec rspec spec/unit/outputs/elasticsearch_spec.rb -e "ES and a shutdown is requested"
Sending Logstash logs to null which is now configured via log4j2.properties
Run options:
  include {:full_description=>/ES\ and\ a\ shutdown\ is\ requested/}
  exclude {:integration=>true, :redis=>true, :socket=>true, :performance=>true, :couchdb=>true, :elasticsearch=>true, :elasticsearch_secure=>true, :export_cypher=>true, :windows=>true}

Randomized with seed 56520
.

Finished in 0.29204 seconds (files took 1.84 seconds to load)
1 example, 0 failures
```

Still passes. Does the test need to call `multi_receive` ?

andsel:
The errors that happens in CI
```
ci-elasticsearch-1  | {"@timestamp":"2023-05-30T12:37:16.799Z", "log.level": "WARN", "message":"http client did not trust this server's certificate, closing connection Netty4HttpChannel{localAddress=/172.20.0.3:9200, remoteAddress=/172.20.0.2:40224}", "ecs.version": "1.2.0","service.name":"ES_ECS","event.dataset":"elasticsearch.server","process.thread.name":"elasticsearch[feff64e10004][transport_worker][T#6]","log.logger":"org.elasticsearch.http.netty4.Netty4HttpServerTransport","elasticsearch.cluster.uuid":"cFI62nwWTTmI58pPVpOrsQ","elasticsearch.node.id":"ovWDynXKQ3e0eVOxpoqvjQ","elasticsearch.node.name":"feff64e10004","elasticsearch.cluster.name":"docker-cluster"}
ci-logstash-1       | [2023-05-30T12:37:16,800][INFO ][logstash.outputs.elasticsearch] Failed to perform request {:message=>"NotAfter: Tue May 23 17:28:55 GMT 2023", :exception=>Manticore::ClientProtocolException, :cause=>#<Java::JavaxNetSsl::SSLHandshakeException: NotAfter: Tue May 23 17:28:55 GMT 2023>}
ci-logstash-1       | [2023-05-30T12:37:16,801][WARN ][logstash.outputs.elasticsearch] Attempted to resurrect connection to dead ES instance, but got an error {:url=>"https://simpleuser:xxxxxx@elasticsearch:9200/", :exception=>LogStash::Outputs::ElasticSearch::HttpClient::Pool::HostUnreachableError, :message=>"Elasticsearch Unreachable: [https://elasticsearch:9200/][Manticore::ClientProtocolException] NotAfter: Tue May 23 17:28:55 GMT 2023"}```

spec/unit/outputs/elasticsearch_spec.rb
andsel: Hi @jsvd with commit ece4c94fb5a5121faf7dd0a8c319c815ff23b0ff I've updated the test, adding the invocation to `multi_receive` and commented the fix to prove that the test fails.
With commit 7c4e477ad34398cf81a7a5793cb99e1472d1d2af I've reenabled the fix to check the tests becomes green, however the CI remains stuck in some jobs as described in [comment](https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/1132#issuecomment-1568367270).
The same error appears also on `main` so should be related to latest changes in `8.8.1`on the ES side, I think.
The PR is ready for another review.

jsvd (approved):
LGTM

Issue #1133, opened 2023-05-17
`pipeline_shutdown_requested?` not found error with Logstash v7.17.x versions.


mashhurs:
**Logstash information**:

Please include the following information:

1. Logstash version (e.g. `bin/logstash --version`): `7.17.10`
2. Logstash installation source (e.g. built from source, with a package manager: DEB/RPM, expanded from tar or zip archive, docker)
  Downloaded `7.17.10` (Linux X86_64) from [Logstash download page](https://www.elastic.co/downloads/past-releases/logstash-7-17-10)
3. How is Logstash being run (e.g. as a service/service manager: systemd, upstart, etc. Via command line, docker/kubernetes)
  Running via command line: `./bin/logstash -f config/simple.conf` and `simple.conf` uses `es-output`.

4. How was the Logstash Plugin installed
`./bin/logstash-plugin install --version 11.15.4 logstash-output-elasticsearch`
**JVM** (e.g. `java -version`):

**OS version** (`uname -a` if on a Unix-like system): Ubuntu 22.04

**Description of the problem including expected versus actual behavior**:
We introduced an `expose shutdown to plugin` feature in `v8.1.1` version and `es-output` plugin started using this feature from its `v11.15.0` versions. When running 11.15.0+ plugin versions with Logstash `7.17.x` versions, plugin crashes. Please look at the below error.

**Steps to reproduce**:

Please include a *minimal* but *complete* recreation of the problem,
including (e.g.) pipeline definition(s), settings, locale, etc.  The easier
you make for us to reproduce it, the more likely that somebody will take the
time to look at it.

 1. Download Logstash `7.17.10`
 2. Install `logstash-output-elasticsearch-v11.15.4`
 3. Use simple config which uses Elastsearch output plugin and run Logstash with `./bin/logstash -f {your-config-path}`

**Provide logs (if relevant)**:
```
[2023-05-17T14:35:42,009][ERROR][logstash.javapipeline    ][main] Pipeline worker error, the pipeline will be stopped
 {:pipeline_id=>"main", :error=>"(NoMethodError) undefined method `shutdown_requested' for #
<LogStash::JavaPipeline:0x302ed65>", :exception=>Java::OrgJrubyExceptions::NoMethodError, :backtrace=>
["home.ubuntu.logstash_minus_7_dot_17_dot_10.vendor.bundle.jruby.$2_dot_5_dot_0.gems.logstash_minus_output_minus_
[...]

PR #1134, opened 2023-05-17
Fixes the undefined `shutdown_requested` method error.

mashhurs:
### Description
When plugin check if shutdown is received, it tries to call [`shutdown_requested?` java API](https://github.com/mashhurs/logstash/blob/6ab14e4e8a8c25ab8b6229b19923564e735a6fa0/logstash-core/src/main/java/org/logstash/execution/AbstractPipelineExt.java#L819) but in ruby code we are missing `?` mark after method. Since there is no `shutdown_requested` in core, plugin stops with the `(NoMethodError) undefined method 'shutdown_requested'` error.

- Closes: #1133

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 11.15.5
+  - Fixes `undefined 'shutdown_requested' method` error when plugin receives shutdown request [#1134](https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/1134)
+
 ## 11.15.4
   - Improved connection handling under several partial-failure scenarios [#1130](https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/1130)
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## 11.15.5
+  - Fixes `undefined 'shutdown_requested' method` error when plugin receives shutdown request [#1134](https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/1134)
```
mashhurs, line 2:
```suggestion
  - Fixes `undefined 'shutdown_requested' method` error when plugin checks if shutdown request is received [#1134](https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/1134)
```

andsel (approved):
LGTM

PR #1137, opened 2023-05-30
Renewed fixture certs

andsel:
Re-newed fixture certificates running the 
```sh
spec/fixtures/test_certs/renew.sh
```


kaisecheng (approved):
LGTM

PR #1140, opened 2023-06-09
Docfix: typo in default data stream name

jeanfabrice:
Fixing default datastream name to `logs-generic-default` rather than `logs-generic-logstash` as `data_stream_namespace` defaults to `default`: https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/main/lib/logstash/outputs/elasticsearch/data_stream_support.rb#L23


karenzone (approved):
Docs look good.

mashhurs (approved):
Thanks for the fix! 💯

karenzone (approved):
Content correct and conflicts resolved

PR #1142, opened 2023-06-12
fix: move ssl standardization to #initialize

### LINKED ISSUE #1141: normalized SSL config not applied when LS core uses private-API build_client directly
Logstash core uses this plugin `build_client` to build an Elasticsearch client that it can use for communications with Elasticsearch in core features like Monitoring and Central Management.

Unfortunately, the SSL standardization work that was completed for 11.14.0 of this plugin in #1118 was not aware of Logstash-core's use of this private API, which bypasses the plugin's normal `#register` method. Because `#register` is not sent by Logstash core, the ls-core-provided path-to-trust directives like `cacert` and `truststore` are not available to the client that it builds, leading to obtuse PKIX-path building failures.

yaauie:
Resolves #1141 by moving ssl standardization to `#initialize`


roaksoax (approved):
LGTM!

PR #1143, opened 2023-07-05
skip processing of bulk response on a 200 OK

jsvd:


lib/logstash/outputs/elasticsearch/http_client.rb
```diff
@@ -179,7 +179,15 @@ def bulk_send(body_stream, batch_actions)
 
       case response.code
       when 200 # OK
-        LogStash::Json.load(response.body)
+        body = response.body
+        if body[0..40].match?(/"errors":false/) && ENV["SHORT_CIRCUIT"] == "1"
```
jsvd, line 183:
```suggestion
        if body[0..40].match?(/"errors":false/)
```

Issue #1135, opened 2023-05-22
Invalid data stream configuration when using dlq_custom_codes option


ckauf:
**Logstash information**:

Please include the following information:

1. Logstash version (e.g. `bin/logstash --version`)
8.7.1

3. Logstash installation source (e.g. built from source, with a package manager: DEB/RPM, expanded from tar or zip archive, docker)
tar.gz

5. How is Logstash being run (e.g. as a service/service manager: systemd, upstart, etc. Via command line, docker/kubernetes)
via CLI

7. How was the Logstash Plugin installed
Bundled

**JVM** (e.g. `java -version`):
Bundled

**OS version** (`uname -a` if on a Unix-like system):
MacOS 13.4

**Description of the problem including expected versus actual behavior**:
Actual behavior: When adding `dlq_custom_codes` to an output with `data_stream => true` the pipeline does not start but instead displays the following error message:
```
Invalid data stream configuration, the following parameters are not supported: {"dlq_custom_codes"=>[401]}
```
Expected behavior: Configuring custom codes works with Data Streams

**Steps to reproduce**:
Create a pipeline config that looks like this:
```
input {
  generator {}
}

output {
  elasticsearch {
    hosts => ["https://localhost:9200/"]
    user => "elastic"
    password => "test"
    dlq_custom_codes => [401]
    data_stream => "true"
  }
}
```

**Provide logs (if relevant)**:
```
[2023-05-22T16:10:21,618][ERROR][logstash.outputs.elasticsearch][main] Invalid data stream configuration, the following parameters are not supported: {"dlq_custom_codes"=>[401]}
[...]

PR #1144, opened 2023-07-18
allow dlq_ settings when using data streams

jsvd:
move dlq_ setting definition to shared parameters (could be used by other ES outputs) so it is allowed to work with data streams

fixes https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/1135


roaksoax (approved):
lgtm!

jsvd:
@logstashmachine bump patch

PR #1148, opened 2023-09-06
Add http compression level

kaisecheng:
This commit deprecated `http_compression` in favor of `compression_level`.
http_compression `false` mapped to compression_level `0`, and `true` mapped to `1`.
The default setting has changed from disabling compression to enabling compression at level 1 (best speed)

Fix: https://github.com/elastic/ingest-dev/issues/2217

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,5 @@
 ## 11.17.0
-  - Added support to http compression level. Deprecated `http_compression` in favour of `compression_level` [#1148](https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/1148)
+  - Added support to http compression level. Deprecated `http_compression` in favour of `compression_level` and enabled compression level 1 by default. [#1148](https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/1148)
 
 ## 11.16.0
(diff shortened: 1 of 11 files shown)


kaisecheng:
The unit tests fail with 8.10.0-SNAPSHOT [BC5](https://artifacts-staging.elastic.co/logstash/8.10.0-7f5c23a8/summary-8.10.0.html) in main branch. 
rspec passed all tests but the docker doesn't exit.

edmocosta (commented):
Nice job @kaisecheng! 

I'm wondering if instead of changing the existing `boolean` configuration into a multi-type value, we could add a new option to set the compression level (e.g. `http_compression_level`).  IMO, I find it a bit confusing having settings accepting boolean and numbers, and specifically for this setting, the `_level` suffix would give it more meaning. WDYT?

edmocosta (approved):
LGTM! 🎉

I've tested it using the latest LS and ES 8.6, and it seems to be working fine.

Just left a few docs suggestions :)

strawgate:
Can we backport the compression changes to latest 7.x?

kaisecheng:
@strawgate we won't pin the minor update in 7.17, but you can install it by `bin/logstash-plugin update logstash-output-elasticsearch`

PR #1151, opened 2023-09-22
shut down pipeline when finish register fails

kaisecheng:
Prior to this change, the errors in the post-register phase (finish_register) behaved differently: template manager logged failure; ILM setup loop printed error messages. Continuing the pipeline would not be meaningful.

With this commit, ConfigurationError is introduced for critical bootstrap failures, such as when the template path is not found. This change stops the pipeline when it gets ConfigurationError or 4xx status code calling template or ILM related API. For HTTP 429 too many request, finish_register retries the bootstrap


kaisecheng:
It would be better to add integration test in core rather than in plugin to test pipeline shutdown.

Test against Logstash 6 fails. Believe we can stop testing 6.x
```
public_suffix-5.0.3 requires ruby version >= 2.6, which is incompatible with the current version, ruby 2.5.3p0 (jruby 9.2.7.0)
```

lib/logstash/outputs/elasticsearch.rb
```diff
@@ -322,10 +322,21 @@ def register
...
     @after_successful_connection_thread = after_successful_connection do
       begin
         finish_register
         true # thread.value
+      rescue LogStash::ConfigurationError, LogStash::Outputs::ElasticSearch::HttpClient::Pool::BadResponseCodeError => e
+        if respond_to?(:execution_context) && execution_context.respond_to?(:pipeline_id) &&
+          execution_context.respond_to?(:agent) && execution_context.agent.respond_to?(:stop_pipeline)
+
+          @logger.error("Failed to bootstrap. Pipeline \"#{execution_context.pipeline_id}\" is going to shut down",
+                        { message: e.message, exception: e.class, backtrace: e.backtrace })
```
jsvd, line 335:
```suggestion
          details = { message: e.message, exception: e.class }
          details[:backtrace] = backtrace: e.backtrace if @logger.debug?
          @logger.error("Failed to bootstrap. Pipeline \"#{execution_context.pipeline_id}\" is going to shut down", details)
```

lib/logstash/outputs/elasticsearch.rb
```diff
@@ -322,10 +322,21 @@ def register
...
     @document_level_metrics = metric.namespace(:documents)
 
+    @stop_after_finish_register = Concurrent::AtomicBoolean.new(false)
     @after_successful_connection_thread = after_successful_connection do
       begin
         finish_register
         true # thread.value
+      rescue LogStash::ConfigurationError, LogStash::Outputs::ElasticSearch::HttpClient::Pool::BadResponseCodeError => e
+        if respond_to?(:execution_context) && execution_context.respond_to?(:pipeline_id) &&
+          execution_context.respond_to?(:agent) && execution_context.agent.respond_to?(:stop_pipeline)
```
jsvd, line 332: do we need all these checks? is it because during tests we may not have all these available?

kaisecheng: I see similar checking in [dlq](https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/v11.18.0/lib/logstash/plugin_mixins/elasticsearch/common.rb#L400) and the `stop_pipeline` function is added in recent two years. `pipeline_id` and `agent` are not necessary

kaisecheng: Tracing back the history of `execution_context.rb` agent and pipeline_id was not there at the [beginning](https://github.com/elastic/logstash/commit/ab1cfe8cf7e20114df58bcc6c996abcb2b0650d7). 
Considering the execution_context was there since 2017, do you think we can safely remove and only check `execution_context.agent.respond_to?(:stop_pipeline)`? I can see other plugins do not check `respond_to?(:execution_context)`

kaisecheng: @jsvd ☝️

jsvd (approved):
LGTM

Issue #1153, opened 2023-09-28
Reduce ES response size through use of filter_path


jsvd:
see https://github.com/elastic/beats/pull/36275

during the _bulk request if `"errors,items.*.error,items.*.status"` is applied to filter_path, it should reduce > 90% of the response size from ES, reducing bandwitdh usage and also making response processing faster (less json to deserialize).

flexitrev:
Cool, can we track performance implications on Logstash instances as this gets rolled out?

jsvd:
For those wanting to test it out, current versions of Logstash allow configuring the bulk path, where the `filter_path` can be set:

```
output {
  elasticsearch {
    hosts => [ ... ]
    bulk_path => '_bulk?filter_path=errors,items.*.error,items.*.status'
  }
}
```

strawgate:
I did a quick initial benchmark with 2GB of data in a simple file input/elasticsearch output pipeline and I see a roughly 5% drop in CPU usage and a 75% drop in response traffic from Elasticsearch.

PR #1154, opened 2023-10-20
Add filter_path to bulk messages

robbavey:
This commit sets the `filter_path` query parameter when sending messages
to Elasticsearch using the bulk API. This should significantly reduce
the size of the query response from Elasticsearch, which should help
reduce bandwidth usage, and improve response processing speed due to
the lesser amount of JSON to deserialize

Resolves: #1153

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 11.19.0
+  - Added `filter_path` to bulk requests to reduce the size of responses from elasticsearch
+  
 ## 11.18.0
   - Added request header `Elastic-Api-Version` for serverless [#1147](https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/1147)
 
--- a/lib/logstash/outputs/elasticsearch/http_client.rb
+++ b/lib/logstash/outputs/elasticsearch/http_client.rb
@@ -182,6 +182,7 @@ module LogStash; module Outputs; class ElasticSearch;
       params[:headers] = {"Content-Encoding" => "gzip"} if compression_level?
 
       response = @pool.post(@bulk_path, params, body_stream.string)
+
       @bulk_response_metrics.increment(response.code.to_s)
 
       case response.code
--- a/lib/logstash/outputs/elasticsearch/http_client.rb
+++ b/lib/logstash/outputs/elasticsearch/http_client.rb
@@ -176,10 +176,7 @@ module LogStash; module Outputs; class ElasticSearch;
     end
 
     def bulk_send(body_stream, batch_actions)
-      params = {
-          :query => {"filter_path" => "errors,items.*.error,items.*.status"}
-      }
-      params[:headers] = {"Content-Encoding" => "gzip"} if compression_level?
+      params = compression_level? ? {:headers => {"Content-Encoding" => "gzip"}} : {}
 
       response = @pool.post(@bulk_path, params, body_stream.string)
 
--- a/lib/logstash/outputs/elasticsearch/http_client.rb
+++ b/lib/logstash/outputs/elasticsearch/http_client.rb
@@ -176,9 +176,12 @@ module LogStash; module Outputs; class ElasticSearch;
     end
 
     def bulk_send(body_stream, batch_actions)
-      params = compression_level? ? {:headers => {"Content-Encoding" => "gzip"}} : {}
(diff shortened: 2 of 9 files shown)


roaksoax (approved):
lgtm~!

lib/logstash/outputs/elasticsearch/http_client.rb
```diff
@@ -176,7 +176,11 @@ def join_bulk_responses(bulk_responses)
     end
 
     def bulk_send(body_stream, batch_actions)
-      params = compression_level? ? {:headers => {"Content-Encoding" => "gzip"}} : {}
+      params = {
+          :query => {"filter_path" => "errors,items.*.error,items.*.status"}
```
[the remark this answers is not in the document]
robbavey: I wondered about that, and was trying to think of a valid reason why they would change the filter path, other than to filter the response to reduce the payload - this feels like it is an implementation details, which I think we only exposed for the monitoring use case, and the payload reduction we already covered.

Possibly debugging?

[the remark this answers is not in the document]
jsvd: It looks like manticore (and httpclient) will allow both to coexist:

`Manticore.get("http://localhost:3333/?q=test",  query:  {q: "kittens"}).body`

causes:

```
❯ nc -l 3333           
GET /?q=test&q=kittens HTTP/1.1
Connection: Keep-Alive
Content-Length: 0
Host: localhost:3333
User-Agent: Manticore 0.9.1
Accept-Encoding: gzip,deflate
```

So it depends how ES chooses to treat these cases since the RFC doesn't seem to prohibit this case, nor explain what the server should do.
Either way we should confirm how ES handles this situation and add a note to the docs about how query parameters  in `bulk_path` interact with [`parameters`](https://www.elastic.co/guide/en/logstash/current/plugins-outputs-elasticsearch.html#plugins-outputs-elasticsearch-parameters), w/ a special mention about `filter_path`.

CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## 11.19.0
+  - Added `filter_path` to bulk requests to reduce the size of responses from elasticsearch
```
mashhurs, line 2:
```suggestion
  - Added `filter_path` to bulk requests to reduce the size of responses from elasticsearch [#1154](https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/1154)
```

lib/logstash/outputs/elasticsearch/http_client.rb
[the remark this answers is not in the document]
robbavey: I guess we have options

* Keep it as it is in the PR
* Detect `filter_path` in the existing URL and use that instead of the new setting, and log appropriately.
* Detect `filter_path` in the existing URL and drop it, enforcing the new setting, and log appropriately.

I'm easy with any of them, as long as we document it clearly

lib/logstash/outputs/elasticsearch/http_client_builder.rb
```diff
@@ -197,5 +197,14 @@ def self.setup_api_key(logger, params)
     def self.dedup_slashes(url)
       url.gsub(/\/+/, "/")
     end
+
+    def self.resolve_filter_path(url)
+      return url if url.nil? || url.match?(/(?:[&|?])filter_path=/)
```
jsvd, line 202: dedup_slashes can never return nil since it performs gsub on a string without checking it it's nil:

```suggestion
      return url if url.match?(/(?:[&|?])filter_path=/)
```

lib/logstash/outputs/elasticsearch/http_client_builder.rb
```diff
@@ -197,5 +197,14 @@ def self.setup_api_key(logger, params)
     def self.dedup_slashes(url)
       url.gsub(/\/+/, "/")
     end
+
+    def self.resolve_filter_path(url)
+      return url if url.match?(/(?:[&|?])filter_path=/)
+      ("#{url}#{query_param_separator(url)}filter_path=errors,items.*.error,items.*.status")
```
jsvd, line 203: Can you add a note about this injection of filter_path and it's skipping in the docs?

robbavey: done

PR #1159, opened 2023-11-09
Replace document_already_exist_exception with version_conflict_engine_exception in the silence_errors_in_log setting example

herrBez:
The example reported in documentation for the setting [`silence_errors_in_log`](https://www.elastic.co/guide/en/logstash/current/plugins-outputs-elasticsearch.html#plugins-outputs-elasticsearch-silence_errors_in_log) does not appear to work with modern version of Elasticsearch. The correct value should be `version_conflict_engine_exception`, which as per  https://github.com/elastic/elasticsearch/blob/main/server/src/main/java/org/elasticsearch/ElasticsearchException.java#L1271C1-L1272C replaced the exception `document_already_exist_exception`.

This PR replace in the example for silence_errors_in_log the deprecated/not more existent `document_already_exist_exception` with the value `version_conflict_engine_exception`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 11.20.1
+- Replace `document_already_exist_exception` with `version_conflict_engine_exception` in the `silence_errors_in_log` setting example [#1159](https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/1159)
+
 ## 11.20.0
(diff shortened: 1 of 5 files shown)


kaisecheng:
@logstashmachine bump patch

kaisecheng:
The changes look good. Please bump the version to `11.20.1` in `CHANGELOG.md` and `logstash-output-elasticsearch.gemspec`. Could you sign the contributor agreement?

kaisecheng (approved):
CI is green. LGTM! Thanks for contributing this project!

PR #1161, opened 2023-11-20
When datastream is enabled then no version related attributes fills into document actions

### LINKED ISSUE #1160: New management of `version` and `version_type` corrupt datastreams processing
**Build scan**:
https://buildkite.com/elastic/logstash-serverless-integration-testing/builds/195

**Repro line**:
To reproduce this follow the steps described in "How to test this PR locally" of PR #1155

**Reproduces locally?**:
yes

**Applicable branches**:
main

**Failure history**:

**Failure excerpt**:

```
[2023-11-20T14:56:18,282][ERROR][logstash.outputs.elasticsearch][main][baa2932b460534470bcad648824fb6b847f062098ce5a136db7a48475e38f60b] Encountered a retryable error (will retry with exponential backoff) {:code=>400, :url=>"https://<an instance>.elastic-cloud.com:443/_bulk?filter_path=errors,items.*.error,items.*.status", :content_length=>5480, :body=>"{\"error\":{\"root_cause\":[{\"type\":\"action_request_validation_exception\",\"reason\":\"Validation Failed: 1: create operations do not support explicit versions. use index instead;\"}],\"type\":\"action_request_validation_exception\",\"reason\":\"Validation Failed: 1: create operations do not support explicit versions. use index instead;\"},\"status\":400}"}
```

andsel:
## Release notes

Avoid to populate version and version type attributes when processing integration metadata and datastream is enabled

## What does this PR do?

During PR #1155 the resolution of `version` and `version_type` ES parameters was moved from the index only event action tuple creation to the common method. This changed was due to do the intent to collect all integration-aware metadata fields in one place, but the common method is used also by the datastream part this result in populating `event` and `event_type` request parameters not only for normal index operations. 
During an index operation on a datastream, if one of those parameters is valued, generated an error on ES indexing, resulting in request fail.

This PR move the processing and creation of event and event_type parameters, back in its original position, splitting the capture of integration metadata in 2 parts.

It's a bugfix that let the plugin to work properly on datastreams.

## Checklist

- ~~[ ] I have commented my code, particularly in hard-to-understand areas~~

## Author's Checklist

- [ ] check with a real ES and integrations

Follow the same plan as described in PR #1155 

## Related issues

- Closes #1160

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 11.22.1
+  - Fix, avoid to populate `version` and `version_type` attributes when processing integration metadata and datastream is enabled. [#1161](https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/1161)
+
 ## 11.22.0
   - Added support for propagating event processing metadata when this output is downstream of an Elastic Integration Filter and configured _without_ explicit `version`, `version_type`, or `routing` directives [#1158](https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/1158)
 
--- a/lib/logstash/outputs/elasticsearch.rb
+++ b/lib/logstash/outputs/elasticsearch.rb
@@ -499,6 +499,16 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
       params[retry_on_conflict_action_name] = @retry_on_conflict
     end
 
+    event_control = event.get("[@metadata][_ingest_document]")
+    event_version, event_version_type = event_control&.values_at("version", "version_type") rescue nil
+
+    resolved_version = resolve_version(event, event_version)
+    resolved_version_type = resolve_version_type(event, event_version_type)
+
+    # avoid to add nil valued key-value pairs
+    params[:version] = resolved_version unless resolved_version.nil?
+    params[:version_type] = resolved_version_type unless resolved_version_type.nil?
+
     EventActionTuple.new(action, params, event)
   end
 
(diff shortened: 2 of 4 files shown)


yaauie (approved):
This fixes the issue of incompatible fields being sent when data streams are enabled. In that sense, I am okay merging as-is.

I've left one note about trimming some commented-out code in the specs.

***

But this fix also reveals a bit of a gap in testability, since the specs need to rely on outside knowledge to test either `event_action_tuple` or `data_stream_event_action_tuple`, depending on whether we expect data streams to be enabled.

These specs currently do _not_ set up data streams, but test the result of a method that _would_ be called if data streams were enabled. The problem is that this diverges from testing plugin behaviour _as it would exist in real life_.

I would like to at least file for follow-up work to clean this up.

This could be resolved by introducing a new method like `action_tuple(event)` that uses the `@event_mapper` (which presently wraps either `event_action_tuple` or `data_stream_event_action_tuple` depending on other configuration):

~~~ ruby
  def action_tuple(event)
    @event_mapper.call(event)
  end
~~~

And then changing the current only call-site of `@event_mapper.call` to go through our new unified boundary:

~~~ diff
diff --git a/lib/logstash/outputs/elasticsearch.rb b/lib/logstash/outputs/elasticsearch.rb
index 598265d..55e9a87 100644
--- a/lib/logstash/outputs/elasticsearch.rb
+++ b/lib/logstash/outputs/elasticsearch.rb
@@ -424,7 +424,7 @@ class LogStash::Outputs::ElasticSearch < LogStash::Outputs::Base
     event_mapping_errors = [] # list of FailedEventMapping
     events.each do |event|
       begin
-        successful_events << @event_mapper.call(event)
+        successful_events << action_tuple(event)
       rescue EventMappingError => ie
         event_mapping_errors << FailedEventMapping.new(event, ie.message)
       end
~~~

We could then test `action_tuple` directly for a given plugin config without needing to _know_ whether we should be testing `event_action_tuple` or `data_stream_event_action_tuple`.

Another possible way would be to pull the datastream-specific and legacy-specific behaviour out into a self-contained Modules so that `setup_mapper_and_target` wouldn't need to store procs-that-call-specific-methods.

PR #1165, opened 2024-02-01
[DOC] Remove "in your" duplication in Serverless section

jsvd:


karenzone (approved):
LGTM! Thanks for the fix, @jsvd

PR #1171, opened 2024-03-19
(backport: 11.4 for LS7.x) Fixes an issue where events containing non-unicode strings.

mashhurs:
See [origin PR](https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/1169) for the details, applying to 11.4.x versions to be available in LS 7.17.x core

### DIFF
--- a/.ci/docker-setup.sh
+++ b/.ci/docker-setup.sh
@@ -48,10 +48,10 @@ if [ "$ELASTIC_STACK_VERSION" ]; then
 
     if [ "$INTEGRATION" == "true" ]; then
         docker-compose down
-        docker-compose build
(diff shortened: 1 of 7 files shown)


yaauie:
I believe we also need to backport https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/1139 to fix the SSL-related permaloops we are seeing in CI

yaauie:
Ah, it looks like https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/1137 is relevant backport-to-fox-tests candidate

yaauie (approved):

PR #1074, opened 2022-05-23
CA Trusted Fingerprint Support

yaauie:
Includes some prerequisite cleanup:
 - renews existing certificates and adds a helper to make future renewals simple
 - adds failure test for connecting to a secured host without a PKIX path to a trusted certificate

Includes `ca_trusted_fingerprint` support from [`logstash-plugins/logstash-mixin-ca_trusted_fingerprint_support`](https://github.com/logstash-plugins/logstash-mixin-ca_trusted_fingerprint_support), routing the generated `TrustStrategy` through the original `params` that are passed to the client builder.

PR #1173, opened 2024-03-19
renew test certs

yaauie:
backports script to renew cert fixtures from #1074 to 11.4.x series, and then executes the script so that non-expired certs are available


mashhurs (approved):
LGTM!

yaauie:
The renew script as executed using my local `openssl` appears to not have copied the CA extension, so the self-signed root is not a valid CA. I'm working to mitigate, and confused as to how executing the script worked previously.

Issue #1167, opened 2024-03-07
Doc: Update data streams info to explain implications of ECS settings 


karenzone:
Reported: "The data streams page doesn't talk about ECS and the fact that you'll hit an error if you have `pipeline.ecs_compatibility` set to `true`."

* Source: https://github.com/logstash-plugins/logstash-output-elasticsearch/edit/main/docs/index.asciidoc
* Logstash Reference page: https://www.elastic.co/guide/en/logstash/current/plugins-outputs-elasticsearch.html#plugins-outputs-elasticsearch-data-streams

Related: #1046

PR #1174, opened 2024-03-20
Doc: Add notice that ecs compatibility is required for data streams

karenzone:
Fixes: #1167


karenzone:
@yaauie, please see if I messaged this to your liking

karenzone:
Update: We're going deeper to be sure we've identified any additional complexities.

PR #1175, opened 2024-03-21
Resync CHANGELOG

yaauie:
- The changelog on `main` had a docs-only entry for `11.4.2` that was not packaged and released until `11.5.0`; update the changelog to reflect where it was actually packaged
 - We recently _backported_ a fix that was first available in `v11.22.3` to a new `11.4-maintenance` branch (cut from the `v11.4.1` tag), and released that fix with `v11.4.2`; we forward-port its changelog entry to `main`, along with a caveat so that users don't assume the fix's availability in intermediate releases.


mashhurs (approved):
LGTM! ❤️

PR #1189, opened 2024-09-13
Add X-Elastic-Product-Origin

lucabelluccini:
Add the `X-Elastic-Product-Origin` header set to `logstash` (similar to https://github.com/elastic/beats/blob/3f44bd1f9bf0b3e17c8254119c464cbb7e872c71/libbeat/common/productorigin/productorigin.go#L24).


donoghuc:
Thanks @lucabelluccini I missed this PR when working on https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/1195 . This should be completed now and shippin in` => 11.22.10`

PR #1195, opened 2024-12-06
Set x-elastic-product-origin header for ES requests

donoghuc:
This commit updates the `Elasticsearch::Client` used to make requests to ES to
send along a header identifying the request as originating from an internal
component.

closes https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/1191

PR #1194, opened 2024-11-28
remove code related to ES 6.x and earlier

jsvd:
- fixes secure integration tests by using credentials to check ES being up
- remove ECS content for ES 6.x
- remove document_type handling for 6.x
- remove two different ways to define certain parameters like `routing` vs `_routing`
- ~~remove warning for ES 6.x~~
- remove ES < 7 specific version handling
- remove ES < 7 specific ILM feature checking
- remove ES < 7 specific template handling
- remove ES 6.x template
- large cleanup related to ES < 7 specific tests

closes https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/1201

PR #1196, opened 2024-12-10
Update CHANGELOG.md to fix PR number and link

karenzone:
Changelog entry merged in #1195 incorrectly links to #1194 which is still WIP. 
If we go with this approach to make the correction, we'll have to fix it again in release notes for all releases that are locked to this plugin version.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,5 @@
 ## 11.22.10
- - Add `x-elastic-product-origin` header to Elasticsearch requests [#1194](https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/1194)
+ - Add `x-elastic-product-origin` header to Elasticsearch requests [#1195](https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/1195)
 
 ## 11.22.9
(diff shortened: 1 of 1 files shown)


donoghuc:
Oops! I'm fine with a re-release, not sure if there is another option though.

karenzone:
@donoghuc, I suggest that we approve/merge this PR so that it'll be correct for the next version push. 
Yanking gems can be ugly, and fixing in output isn't the worst. 
If that works for you, it works for me.

donoghuc:
For sure. I realize saying "re-release" is confusing. I didnt mean to try to *replace* the `11.22.10` release. I meant i would be ok with doing a `11.22.11` to trigger the update for the release notes. I guess my question is: do we want to do a release just for the RN update? Or should we just merge this and let it go out in the next release whenever it may go out? 

Also, i just re-kicked the travis run (looks like maybe a race condition in one of our tests), i'll merge this once that goes through green.

karenzone:
> I guess my question is: do we want to do a release just for the RN update? Or should we just merge this and let it go out in the next release whenever it may go out?

No need to bump and re-publish because it wouldn't really change anything in docs anyway.  After this is merged (thank you for offering to handle this, by the way), the changelog will be correct for the next time we update this plugin.

donoghuc (approved):
Failures just due to no 8.17.1-SNAPSHOT image available yet. this is G2G

PR #1199, opened 2025-01-07
Properly handle 413 Payload Too Large errors

donoghuc:
Previously when Elasticsearch responds with a 413 (Payload Too Large) status,
the manticore adapter raises an error before the response can be processed
by the bulk_send error handling. This commit refactors the way
`BadErrorResponse` codes are handled. Previously we had logic in the manticore
adaptor which special cased raising errors on some codes. This commit refactors
such that the adaptor raises on any error status and the caller is now
responsible for special case handling the code.

PR #1203, opened 2025-01-15
make requests return exception only on 2xx codes

jsvd:
attempting a different strategy from https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/1199 to reduce duplication.

failing tests still present at this point.


donoghuc:
Thanks so much. I rebased https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/1199 on this and fixed up a couple things. I'm going to go ahead and close this so we can keep all the context in the PR conversation in 1199. :)

Issue #1064, opened 2022-01-19
Support base64 encoded Api Key


jsvd:
Currently logstash supports the api key in the form of "id:key":

```
input {
  elasticsearch {
    "api_key" => "TiNAGG4BaaMdaH1tRfuU:KnR6yE41RrSowb0kQ0HWoA" 
  }
}
```

However, the Elasticsearch output when creating an API Key is:

```
{
  "id" : "9tPacX4BGhtIQNIMawN9",
  "name" : "my-api-key",
  "api_key" : "hATekzVwRemmgCSxJXcWsw",
  "encoded" : "OXRQYWNYNEJHaHRJUU5JTWF3Tjk6aEFUZWt6VndSZW1tZ0NTeEpYY1dzdw=="
}
```

Which means a user must compose the "id:key" themselves. Using the encoded form would mean less copy pasting, less errors, and simpler experience.
For backwards compatibility sake we can still support the "id:key" format by detecting if the input has a ":". In this case we can skip the base64 encoding and use the input as is.

PR #1207, opened 2025-03-19
support base64 encoded api key

jsvd:
fixes https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/1064


yaauie:
build failure is unrelated; something in our dependency graph no longer causes this plugin's open-ended development dependency on `elasticsearch` to resolve to `7.x`, so we get 8.x and fail.

mashhurs:
> build failure is unrelated; something in our dependency graph no longer causes this plugin's open-ended development dependency on `elasticsearch` to resolve to `7.x`, so we get 8.x and fail.
> 
> EDIT: it's [logstash-core 8.17+ that now relies on `elasticsearch ~>8`](https://github.com/elastic/logstash/pull/17161/files) 😩, so the build is just plain broken.
> 
> We'll need to either:
> 
> * (a) update this code to use the v8 es client (which is _not_ backward-compatible with older ES's); OR
> * (b) update all tests to http at elasticsearch manually.

Sorry for that! https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/1208 fixes the CI failures.
I have opened [an issue](https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/1209) to remove elasticsearch-ruby v7 sources once LS-core fully moves to v8 version.

PR #1208, opened 2025-03-20
Tolerate the elasticsearch-ruby v8 client in integration tests.

mashhurs:
Recently LS core upgraded elasticsearch-ruby client to v8. This plugin uses it in integration tests and recent failures show due to es-ruby client upgrade: https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/1207/checks?check_run_id=39057545885

This change tolerates  elasticsearch-ruby both v7 and v8 clients.


spec/es_spec_helper.rb
```diff
@@ -1,15 +1,20 @@
...
 require 'cabin'
 
+# remove this condition and support package once plugin starts consuming elasticsearch-ruby v8 client
+# in elasticsearch-ruby v7, ILM APIs were in a separate xpack gem, now directly available
+unless elastic_ruby_v8_client_available?
+  require_relative "support/elasticsearch/api/actions/delete_ilm_policy"
+  require_relative "support/elasticsearch/api/actions/get_alias"
+  require_relative "support/elasticsearch/api/actions/put_alias"
+  require_relative "support/elasticsearch/api/actions/get_ilm_policy"
+  require_relative "support/elasticsearch/api/actions/put_ilm_policy"
```
jsvd, line 15: With this refactoring we no longer need the files related to aliases, which can be deleted from the branch and the require removed too:

```suggestion
  require_relative "support/elasticsearch/api/actions/delete_ilm_policy"
  require_relative "support/elasticsearch/api/actions/get_ilm_policy"
  require_relative "support/elasticsearch/api/actions/put_ilm_policy"
```

spec/integration/outputs/ilm_spec.rb
```diff
@@ -532,3 +533,8 @@
   end
 
 end
+
+def get_expected_error_class
+  return Elastic::Transport::Transport::Errors::NotFound if elastic_ruby_v8_client_available?
+  Elasticsearch::Transport::Transport::Errors::NotFound
+end
```
jsvd, line 540: I suggest making this more explicit:

```suggestion
let(:not_found_exception_class) do
  if elastic_ruby_v8_client_available?
    Elastic::Transport::Transport::Errors::NotFound
  else
    Elasticsearch::Transport::Transport::Errors::NotFound
  end
end
```

spec/es_spec_helper.rb
mashhurs: Ah nice catch, thanks. Removed them with their usages.

spec/integration/outputs/ilm_spec.rb
jsvd: ok let's keep it.

jsvd (approved):
LGTM assuming tests are all green

PR #1211, opened 2025-04-11
Demote WARN to INFO if behaviour is expected for connect.

ash-darin:
This will prevent logstash from throwing a warning if connecting is an expected behaviour.


jsvd (approved):
LGTM

PR #1217, opened 2025-07-21
add headers reporting uncompressed size and doc count

jsvd:


robbavey (approved):
Code looks good to me - I added a comment on Travis test scopes, which can be resolved here, or in a separate PR

PR #1220, opened 2025-08-04
Update .travis.yml to account for 9.x / main

jsvd:
Tests were failing in this plugin, this PR fixes that by:

- Allowing for 9.next to not exist but still mark the build as green
- Fixing a test related to metrics of a partially failed bulk by writing to a random index instead of the logs- stream (which has failure store enabled by default)
- Adding main to secure integration testing.


spec/integration/outputs/metrics_spec.rb
```diff
@@ -5,7 +5,8 @@
     require "logstash/outputs/elasticsearch"
     settings = {
       "manage_template" => false,
-      "hosts" => "#{get_host_port()}"
+      "hosts" => "#{get_host_port()}",
+      "index" => "custom_index_#{rand(10000)}"
```
donoghuc, line 9: Clever! Does this deserve a comment in the code?

jsvd, line 9:
```suggestion
      # write data to a random non templated index to ensure the bulk partially fails
      # don't use streams like "logs-*" as those have failure stores enabled, causing the bulk to succeed instead
      "index" => "custom_index_#{rand(10000)}"
```

PR #1221, opened 2025-08-06
Track logstash versions in .ci

donoghuc:
Previously logstash versions were tracked in the logstash repo and curled during
CI steps. This commit moves the source of truth to the `.ci` repo where the
consumer of that information lives. This also changes the format from json to
yaml to allow commenting in the file.


donoghuc:
https://github.com/logstash-plugins/.ci/pull/79 should fix CI

mashhurs (approved):
LGTM!

Issue #1222, opened 2025-09-03
Support opaque `api_key`


robbavey:
The Elasticsearch output accepts an `api_key` parameter, which is expect to be in the `name:key` format. This value is base64-encoded into an opaque value before being used in an `Authorization: ApiKey #{opaque}` HTTP header.

To reduce friction, we should accept the opaque value for `api_key` anywhere that we accept the `name:key` format `api_key`. We can do so by assuming that the value is opaque unless it includes a colon (:) character.

PR #1223, opened 2025-09-15
Support both, encoded and non encoded api-key formats on plugin configuration

alexcams:
Avoid double-encoding opaque Elasticsearch API keys.

Previously, API keys that were already in encoded format were incorrectly re-encoded, causing connection failures. This update ensures that only raw keys are encoded, supporting both opaque and `id:api-key` formats without issue.

Closes [#1222](https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/1222)

### DIFF
--- a/lib/logstash/outputs/elasticsearch/http_client_builder.rb
+++ b/lib/logstash/outputs/elasticsearch/http_client_builder.rb
@@ -188,9 +188,14 @@ module LogStash; module Outputs; class ElasticSearch;
     def self.setup_api_key(logger, params)
       api_key = params["api_key"]
 
-      return {} unless (api_key && api_key.value)
+      return {} unless (api_key&.value)
 
-      { "Authorization" => "ApiKey " + Base64.strict_encode64(api_key.value) }
+      api_key_value = api_key.value
+      if api_key_value =~ /\A[^:]+:[^:]+\z/
+        api_key_value = Base64.strict_encode64(api_key_value)
+      end
+
+      { "Authorization" => "ApiKey " + api_key_value }
     end
 
     private
--- a/lib/logstash/outputs/elasticsearch/http_client_builder.rb
+++ b/lib/logstash/outputs/elasticsearch/http_client_builder.rb
@@ -190,12 +190,9 @@ module LogStash; module Outputs; class ElasticSearch;
 
       return {} unless (api_key&.value)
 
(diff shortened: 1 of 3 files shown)


lib/logstash/outputs/elasticsearch/http_client_builder.rb
```diff
@@ -188,9 +188,14 @@ def self.setup_basic_auth(logger, params)
     def self.setup_api_key(logger, params)
       api_key = params["api_key"]
 
-      return {} unless (api_key && api_key.value)
+      return {} unless (api_key&.value)
 
-      { "Authorization" => "ApiKey " + Base64.strict_encode64(api_key.value) }
+      api_key_value = api_key.value
+      if api_key_value =~ /\A[^:]+:[^:]+\z/
```
donoghuc, line 194: I think we would have better luck just detecting if something is base64 already (instead of a complex regex trying to find `key:value` (I can think of an edge case here for where `value` has a `:` char which breaks this). 

Instead I think we should simply try to decode base 64, if that works we dont try to encode it, otherwise we assume there was a `:` char and therefore it needs to be encoded. 

Maybe something like:
```ruby
def is_base64?(string)
  begin
    Base64.strict_decode64(string)
    true
  rescue ArgumentError
    false
  end
end
```

alexcams: Yeah, I agree. I've changed it by a function that decodes-encodes the input string and compares it with the original string. This way we avoid to take as correctly encoded, any string that will not raise an error while decoding it.

lib/logstash/outputs/elasticsearch/http_client_builder.rb
```diff
@@ -208,5 +210,13 @@ def self.resolve_filter_path(url)
     def self.query_param_separator(url)
       url.match?(/\?[^\s#]+/) ? '&' : '?'
     end
+
+    def self.is_base64?(string)
```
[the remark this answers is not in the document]
alexcams: Thanks for clarification, I've just pushed this suggested refactoring 😄

donoghuc (approved):
Thanks for doing the refactor to make those internal methods actually private! This looks good. I see you added a section in the changelog (and proposed a patch release version there). We can follow up on all the steps to push a release separately. Nice work!

PR #1224, opened 2025-09-15
Update data streams documentation to note that sprintf syntax is unsupported

robbavey:
Using sprintf syntax in the `data_streams` options is not supported. This commit adds a supplemental note, explaining this, and providing an example on how to dynamically route using `mutate` filter and the auto routing facility.

Relates: https://github.com/elastic/logstash/issues/17689


karenzone (approved):
LGTM

PR #1227, opened 2025-09-29
Use docker-setup.sh from upstream .ci repo

donoghuc:
The custom docker-setup.sh script here is not materially different from the shared script in the `.ci` repository. The only difference were pulling and starting the elasticsearch image. This should not affect CI materially and the consistency gain we get using the same `docker-setup.sh` file everywhere is preferrable to having a slightly different script here.

See https://github.com/elastic/ingest-dev/issues/5960 for more details


andsel (approved):
LGTM

PR #1257, opened 2026-04-14
docs: update CI badges

v1v:
image


donoghuc (approved):
I hadnt noticed this use case. I'll have to update plans for rolling out migration to GHA ;)

PR #1252, opened 2026-03-23
Doc: Use index with/out datemath OR alias/datastream

karenzone:
Resolves: https://github.com/elastic/docs-content/issues/2761

Test is here:  
[DocTest - elasticsearch-output - DO NOT MERGE](https://github.com/elastic/logstash-docs/pull/2038#top)#2038

PR #1269, opened 2026-05-29
Index target no date math for data stream nor alias

stefnestor:
👋 Regarding https://github.com/elastic/docs-content/issues/2761 & https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/1252, 

Attempts to reduce doc pain point surface area to explicitly note to users to stop using date math with aliases or data streams. 

Aliases will not ILM rollover bootstrap & data streams will bootstrap but incur multiple data streams which then fail to progress through ILM rollover as expected.

### DIFF
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -843,9 +843,9 @@ The indexing target to write events to.
 Can point to an {ref}/index-mgmt.html[index], {ref}/aliases.html[alias], or {ref}/data-streams.html[data stream]. 
 This can be dynamic using the `%{foo}` syntax.
 
-Dynamic syntax might make sense with data streams and aliases, but results should be
-validated against cluster index templates and Index lifecycle management (ILM) setup.
-Placeholder date math should not normally be used with aliases nor data streams.
+NOTE: When using dynamic syntax, verify that the resolved index names align with your cluster index templates and Index Lifecycle Management (ILM) configuration.
+
+WARNING: Date math (e.g. `%{+yyyy.MM.dd}`) is not recommended for use with aliases or data streams.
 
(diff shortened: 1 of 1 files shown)


docs/index.asciidoc
```diff
@@ -842,10 +842,15 @@ NOTE: `ilm_rollover_alias` does NOT support dynamic variable substitution as
 The indexing target to write events to. 
 Can point to an {ref}/index-mgmt.html[index], {ref}/aliases.html[alias], or {ref}/data-streams.html[data stream]. 
 This can be dynamic using the `%{foo}` syntax.
+
+Dynamic syntax might make sense with data streams and aliases, but results should be
+validated against cluster index templates and Index lifecycle management (ILM) setup.
+Placeholder date math should not normally be used with aliases nor data streams.
+
```
kaisecheng, line 849:
```suggestion

NOTE: When using dynamic syntax, verify that the resolved index names align with your cluster index templates and Index Lifecycle Management (ILM) configuration.

WARNING: Date math (e.g. `%{+yyyy.MM.dd}`) is not recommended for use with aliases or data streams.

```

kaisecheng:
/pr_prepare release:patch
- [Doc] Add note for index option

docs/index.asciidoc
kaisecheng: @stefnestor thank you for contributing the doc. I transform your change to note and warning. Can you check if it aligns to your original message?

stefnestor: I like it, cheers!

alexcams:
/pr_prepare release:patch
- [Doc] Add note for index option

PR #1274, opened 2026-06-23
Support Elastic Cloud API keys in the `api_key` option

### LINKED ISSUE #1275: Support Elastic Cloud Org level API keys
es-output plugin accepts 
- `<id>:<key>` pair API key (Logstash, beats format)
- base64 (obfuscated key:val) encoded string
Any other formats will be base64 encoded and used in the HTTP requests.
Elastic cloud org level APIs carry `essu_` prefix + `base64`  string which is neither `<id>:<key>` pair or `base64` format, so plugin base64 encodes this key and ES rejects.

Benefit of using Elastic cloud API is person independent but Hosted/Project API keys are tight to the person who creates the API keys.

Relates to https://github.com/elastic/logstash/issues/17360

### Expectation
When providing Elastic cloud org level APIs to `api_key` param, plugin needs to utilize as it is in its requests to connecting Elastic cloud.

arenard:
## Summary

Allows an Elastic Cloud API key (the `essu_`-prefixed keys, including the unified Serverless keys) to be used in the `api_key` option. These keys are not in the `id:api_key` form and cannot currently be used with this output.

## Background

The `api_key` value is turned into an `Authorization: ApiKey <value>` header. The existing logic base64-encodes any value that is not already valid standard base64. An `essu_` key is neither an `id:api_key` pair nor valid standard base64 (the `essu_` prefix contains `_`, which is outside the standard base64 alphabet), so it ends up being base64-encoded a second time and Elasticsearch rejects the resulting header.

Elasticsearch accepts these keys directly as `Authorization: ApiKey <key>` (Cloud API key authentication, available since Elasticsearch 9.1), so the gap is only on the Logstash side.

## Change

`setup_api_key` now recognises the `essu_` prefix and forwards such keys unchanged. Raw `id:api_key` pairs and their base64-encoded form keep their previous behaviour, so existing configurations are unaffected.

## Tests

Added a unit test that asserts an `essu_` key is sent verbatim, with no re-encoding. The existing raw and already-encoded api-key tests still pass.

## Docs

Updated `docs/index.asciidoc` and the inline `api_key` option documentation to list the three accepted forms.

## Related

The same fix is applied to the other two Elasticsearch plugins so the behaviour is consistent:
- logstash-plugins/logstash-input-elasticsearch#274
- logstash-plugins/logstash-filter-elasticsearch#215


alexcams (approved):
LGTM

PR #1276, opened 2026-07-03
Fix serverless compatibility: nil params in pool requests and unsupported template settings

alexcams:
The Elasticsearch output plugin fails to start when targeting a Serverles cluster. There are two bugs in the serverless code path that prevent template installation.

Both issues only surface when using a **custom index name** (e.g., `index => "test-index"`), which disables data streams. With data streams enabled (the default), `manage_template` is auto-set to `false`, ILM is skipped, and none of the affected code paths are reached — which explains why these bugs were not caught during previous serverless testing.

### 1. `NoMethodError: undefined method '[]' for nil`

[`perform_request_to_url`](https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/main/lib/logstash/outputs/elasticsearch/http_client/pool.rb#L353) injects the `Elastic-Api-Version` header on serverless via `params[:headers]`, but three callers in [`http_client.rb`](https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/main/lib/logstash/outputs/elasticsearch/http_client.rb) pass `nil` explicitly as params — overriding the `params={}` default:

- [L436](https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/main/lib/logstash/outputs/elasticsearch/http_client.rb#L436): `@pool.put(path, nil, ...)` in `template_put`
- [L450](https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/main/lib/logstash/outputs/elasticsearch/http_client.rb#L450): `@pool.put(CGI::escape(alias_name), nil, ...)` in `rollover_alias_put`
- [L476](https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/main/lib/logstash/outputs/elasticsearch/http_client.rb#L476): `@pool.put(path, nil, ...)` in `ilm_policy_put`

**Fix:** Replace `nil` with `{}` at these call sites.

### 2. `illegal_argument_exception: Settings [index.number_of_shards] are not available when running in serverless mode`

The default templates include [`"number_of_shards": 1`](https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/main/lib/logstash/outputs/elasticsearch/templates/ecs-v8/elasticsearch-9x.json#L7) (all 4 template files), which is not available in serverless. [`TemplateManager.install_template`](https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/main/lib/logstash/outputs/elasticsearch/template_manager.rb#L32) does not strip these settings before sending.

Confirmed against the [Serverless index settings documentation](https://www.elastic.co/docs/reference/elasticsearch/index-settings/serverless): `number_of_shards` and `number_of_replicas` are the only settings used in the default templates that are not in the serverless-supported list. All other template settings (`index.refresh_interval`, mappings, etc.) are supported.

**Workaround:** Setting `manage_template => false` avoids the error, but the user loses automatic index template management. The plugin should handle this transparently since it already detects serverless via `build_flavor` and adapts behavior accordingly for ILM and template API selection.

**Proposed enhancement:** Strip `number_of_shards` and `number_of_replicas` from the template when `serverless?` is true, consistent with how the plugin already handles other serverless incompatibilities.

### Steps to reproduce

1. Configure `logstash-output-elasticsearch` targeting a Serverless Elasticsearch cluster with a standard index (not data streams)
2. Run Logstash — template installation crashes


kaisecheng (commented):
Good catch! How about adding `params ||= {}` at the begining of [perform_request_to_url](https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/main/lib/logstash/outputs/elasticsearch/http_client/pool.rb#L353) , so no further nil pass can trigger the same problem?

alexcams:
Agree @kaisecheng! Better to be protected. Added it in last commit :https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/1276/changes/f38c446addb990e35f5a3c2b5be28c1e5eb3e7a9

kaisecheng (approved):
LGTM. Thank you!

kaisecheng:
remember to add changelog

kaisecheng:
Also, need to backport to 11.x

PR #1278, opened 2026-07-17
Support Base64 and Elastic Cloud org level api keys.

alexcams:
## Align API key handling with main branch

Backports the `api_key` format handling from the `main` branch to `4.x`, which ships with Logstash 8.19.x.

### Changes

The `api_key` option now accepts three formats:

- **`id:api_key`** — raw pair, base64-encoded automatically by the plugin (existing behavior)
- **Base64-encoded value** — passed through as-is, avoiding double-encoding (e.g. the `encoded` field returned by the Create API Key API, available since ES 7.16.0)
- **Elastic Cloud API key (`essu_` prefix)** — passed through verbatim, enabling authentication against Serverless projects using organization-level Cloud API keys

Unrecognized formats are rejected at startup with a clear `ConfigurationError`, so misconfigured keys surface immediately rather than as opaque authentication failures at runtime.


alexcams:
/pr_prepare release:patch
- Support base64-encoded and Elastic Cloud (`essu_`-prefixed) API keys in the `api_key` option; reject unrecognized formats at startup.

andsel (approved):
LGTM

PR #1282, opened 2026-07-28
Auto-generate SSL test certificates at CI runtime

### LINKED ISSUE #1281: Auto renew or generate SSL certs at the integration tests rutime
Currently the plugin specs carry already generated SSL certs - https://github.com/logstash-plugins/logstash-output-elasticsearch/tree/main/spec/fixtures/test_certs for integration tests. When SSL certs are expired, CIs fail, example - https://github.com/logstash-plugins/logstash-output-elasticsearch/actions/runs/30080195555/job/89439763503

It is not an huge issue for normal cases that we can renew them manually, (for example, the recent one - https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/1279) but if feature is landing and SSL certs are expired (especially hard time like during the FF), no way to prove quickly except local runs.

For the long term, it would be nice if integration tests generate certs at the runtime (similar to `elastic_integration` plugin - https://github.com/elastic/logstash-filter-elastic_integration/tree/main/utils/ssl-certs-generator).

kaisecheng:
Closes #1281

## Summary

Replace pre-committed SSL test certificates with runtime generation via an idempotent shell script. This eliminates recurring CI failures caused by expired certificates and removes the need for manual certificate renewal.

## Changes

- Add `spec/fixtures/test_certs/generate.sh`: generates all SSL test certs (CA, server, invalid SAN, self-signed) from scratch using openssl. Idempotent (skips if all certs exist and are not expired; `--force` to override).
- Add `spec/fixtures/test_certs/openssl.cnf`: centralizes certificate extensions (SAN config, key usage, CA constraints).
- Add `spec/fixtures/test_certs/.gitignore`: prevents generated cert artifacts from being committed.
- Modify `.ci/docker-run.sh`: call `generate.sh` before `docker compose up` so certs exist on disk before being volume-mounted into the Logstash/Elasticsearch containers (see `.ci/docker-compose.override.yml`).
- Modify `.ci/Dockerfile.elasticsearch`: drop the build-time `COPY` of cert files (no longer needed now that certs are supplied via volume mount at runtime instead of baked into the image).
- Remove `renew.sh` and all 14 previously committed cert files.


donoghuc (approved):

donoghuc:
I like it. Dont forget to backport to 11.x :)

PR #1283, opened 2026-07-31
Auto-generate SSL test certificates at CI runtime (backport #1282 to 11.x)

kaisecheng:
Backport of #1282 to `11.x`.

## Summary

Replace pre-committed SSL test certificates with runtime generation via an idempotent shell script. This eliminates recurring CI failures caused by expired certificates and removes the need for manual certificate renewal.

## Changes

- Add `spec/fixtures/test_certs/generate.sh`: generates all SSL test certs (CA, server, invalid SAN, self-signed) from scratch using openssl. Idempotent (skips if all certs exist and are not expired; `--force` to override).
- Add `spec/fixtures/test_certs/openssl.cnf`: centralizes certificate extensions (SAN config, key usage, CA constraints).
- Add `spec/fixtures/test_certs/.gitignore`: prevents generated cert artifacts from being committed.
- Modify `.ci/docker-run.sh`: call `generate.sh` before `docker compose up` (gated on `SECURE_INTEGRATION == true`, so plain unit-test runs skip it) so certs exist on disk before being volume-mounted into the Logstash/Elasticsearch containers (see `.ci/docker-compose.override.yml`).
- Modify `.ci/Dockerfile.elasticsearch`: drop the build-time `COPY` of cert files — no longer needed now that certs are supplied via volume mount at runtime instead of baked into the image.
- Remove `renew.sh` and all previously committed cert files.


kaisecheng:
Red CI is unrelated to the change. This PR is not opened from personal forks but a logstash-plugins's branch, which trigger the first ever build `logstash-output-elasticsearch-pull-request`

```
buildkite-agent: fatal: failed to read file: open .buildkite/pull-request-pipeline.yml: no such file or directory
```

https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/1265 removed `.buildkite/pull-request-pipeline.yml` from this repo back in May. The buildkite [pipeline](https://buildkite.com/elastic/logstash-output-elasticsearch-pull-request) itself was re-created afterward (2026-06-23) without the file ever being restored, so it's been broken since day one. Needs someone with Buildkite admin access to archive it.

donoghuc (approved):
ah, yeah i thought we had cleaned all the buildkite stuff out. Missed this one. Thanks.

